PR #150, opened 2017-07-19
#148 Fix infinite loop when word is too large for column

### LINKED ISSUE #148: Long string in default value causes infinite loop within usage printer
In using picocli version 0.9.7 as part of a Groovy script I am writing, I have uncovered an infinite loop in the usage() printer. Specifically, I have a command line so:

    @Command(showDefaultValues = true)
    class MainCommand {
        @Option(names = "--config-dir", description = "The path to the directory holding the configs")
        File configDir = Paths.get(getClass().getProtectionDomain().getCodeSource().getLocation().toURI()).getParent().getParent().resolve("configs").toFile()
    }

This results in a long path like so: `C:\Users\USERNAME\Development\MY_SOFTWARE\my-software-config\configs`

This value, after running through break iterator, results in the value of variable text within method `CommandLine.Help.TextTable.copy(BreakIterator line, Text text, Text destination, int offset)` being `\Users\USERNAME\Development\MY_SOFTWARE\my-software-config\configs`, which is unable to be wrapped when used by `CommandLine.Help.TextTable.putValue(int row, int col, Text value)` line 3013. This inability to wrap causes the wrap iterator inside that function to go into an infinite loop.

Now, my workaround is to stop outputting the default value, but I imagine this could bite others as well, but that makes my script less clear.

smartboyathome:
Splits the word naively at the column boundary. Don't know if this is the best way but seems to pass all the tests, including the one I made to repro this issue.

Fixes #148

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -3040,6 +3040,10 @@ public class CommandLine {
                         break;
                     }
                 }
+                if (done == 0 && length(text) > columnValue.maxLength) {
+                    // The value is a single word that is too big to be written to the column. Write as much as we can.
+                    done = copy(text, columnValue, offset);
+                }
                 return done;
             }
             private static int copy(Text value, Text destination, int offset) {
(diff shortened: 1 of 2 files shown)


remkop:
Thanks a lot for the PR! I will review tonight when I get home from work.

remkop:
Thank you for your pull request!

PR #177, opened 2017-08-26
Forgotten class InetAddress in the description of enum.

FilipJirsak:
There was forgotten class `InetAddress` in the description of enums in autocomplete documentation.


remkop:
Oops! Good catch! Thank you very much!

Issue #223, opened 2017-10-30
Clean running examples!


aadrian:
Hi,

Would be nice to have some clean and running examples.

The existing ones seem "overloaded" and and mixed with unit tests :( .

Thank you.

remkop:
Good point. The manual contains mostly snippets to illustrate a point very briefly. 
The [`Checksum` example](http://picocli.info/#CheckSum-application) at the beginning of the manual is the only full running example. 

If you have ideas for additional full examples that’d be great. They need to be fairly short though, about  20-25 lines of code.

aadrian:
> Where should they go to?

One possibility would be to have an `/examples` subdirectory/package, but that would require changing the `build.gradle` too.
Many other projects just use the multi-module features of Gradle but that also requires allot of changes.

Maybe a much simpler way would be to just make another GitHub repo: e.g. `picocli-examples` that would have:
```groovy
dependencies {
    compile 'info.picocli:picocli:2.0.1'
    compile 'org.fusesource.jansi:jansi:1.16'
}
```
in `build.gralde`

> Ideally, they could be self-contained,

Yes, but each example in it's own sub-directory? This way best practices could be better demonstrated (e.g. even if one example requires more than one file)

remkop:
I'd prefer to keep the examples in the same repo. 

I'm open to other ideas but I was thinking to add an `picocli.examples` package under `src/test/` (with subpackages if needed for elaborate examples). One nice thing is that the tests can have dependencies on other libraries, that may be useful for some examples.

Something like:
```
src/test/groovy/picocli/examples
src/test/java/picocli/examples
```

RobertZenz:
I'd also opt for the same repository. On a similar note, I wouldn't put the examples into the test directory, because, at least for me, I'd only expect "automated" tests in there not some form of documentation.

Maybe `src/examples/java/Blah.java` is an alternative? Or putting the examples directory directly under the root, to make it easily findable/visible?

remkop:
Yes that would work. Would be cleaner but just need to work out how the gradle build needs to change. Should be doable...

```
// assuming a picocli.examples package:
src/examples/java/picocli/examples/Blah.java
src/examples/groovy/picocli/examples/Blah.groovy
```

remkop:
Or we can have a separate module (in the same repo):

```
// main module (part of the distribution)
src/main/java
src/test/groovy
src/test/java

// separate module (not part of the distribution)
examples/src/main/java
examples/src/main/groovy
```

[...]

PR #225, opened 2017-11-01
Java examples for #223.

RobertZenz:
This is supposed to be a preview of how the usage examples I'm going to write will look like, so no merge is requested at this point. Just let me know if the style is okay like this or if you'd like to have them differently.

I'll add more later, of course and update this pull request until I'm done.

### DIFF
--- a/examples/src/main/java/picocli/examples/PopulateFlagsMain.java
+++ b/examples/src/main/java/picocli/examples/PopulateFlagsMain.java
@@ -1,0 +1,128 @@
+/*
+   Copyright 2017 Robert 'Bobby' Zenz
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ */
+
+package picocli.examples;
+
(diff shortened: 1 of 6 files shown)


remkop:
Thanks for working on this! Generally looks great!

I think it would be a good idea for all examples to show error handling for invalid input. (Like, when a non-existing an option is specified.) What do you think?

Also, could the package for the examples be `picocli.examples`? If the example class is viewed in isolation it's nice if the package name still shows the intention.

aadrian:
> I think it would be a good idea for all examples to show error handling for invalid input.

+1 .
Would be really useful for the examples to show best practices as much as possible.

remkop:
The directory structure `examples/src/main/java/picocli/examples` is now pushed to master.

I put some initial Groovy examples in `examples/src/main/groovy/picocli/examples`. These are meant to be used in a DZone article I'm working on. I wanted to keep these scripts very short, hence the lack of javadoc or groovydoc...

RobertZenz:
I've updated the examples, added error handling and moved them to the examples package. I just realized that another example for when help is requested is missing, I will add that later.

remkop:
This is great! 
Can you rename Unnamed Parameters to Positional Parameters to be consistent with the user manual?
Add yourself as an `@author` in the javadoc, and it’s ready to merge I think.

remkop:
Would you mind prefixing the class names these examples with `Populate`? 

I’m thinking later we’ll probably also have examples for the `CommandLine::parse` method, and for the `run`, `call` and `parseWithHandlers` convenience methods. 

Class names may help users quickly find examples relevant for their needs.

RobertZenz:
I've renamed the classes, added myself as author and also renamed the example for the positional parameters. Also there is now a simple example for printing the usage/help (though, a little bit redundant because I already do that in all examples, but still) and mostly spelling fixes are in it.

PR #264, opened 2018-01-14
Added ability to mask default values

ymenager:
This adds a defaultValueMask() value to @Option. 

When specified the string will be used instead of the default value (for example when the default value is sensitive and shouldn't be displayed like passwords and such.

Unit Test included


ymenager:
actually last night I've noticed this needs to be #265 need to be merged in the same branch since they have interactions.

I'll do the changes tonight, so for now closing this PR and I'll put a note in 265 to hold merging until I've done the changes

PR #265, opened 2018-01-14
Added notRequiredWithDefault to Command annotation which will make…

ymenager:
…annotations with default values to never be required


ymenager:
I've added an extra feature to this branch, which is the ability to refresh the default values. 

Couldn't do it in a different branch because of required interaction with "notRequiredWithDefault"

ymenager:
as mentioned in #264 , please hold off merging this at this moment. I'll merge the default masking changes into this branch and update this PR to let you know it's good to go ahead.

ymenager:
Actually I'll create another branch that incorporate all the changes related to default values, and I'll create a new PR, so closing this PR as well as the other one about fixing setAccessible()

PR #266, opened 2018-01-15
default values in private fields not being read

ymenager:
initFromAnnotatedFields not using setAccessible and thus can't get default values from private fields


ymenager:
will create a new PR later tonight incorporating all default value improvements

Issue #261, opened 2018-01-14
Allow Options with defaults to be always non-required


ymenager:
It would be useful to be able to programatically make a required field to "not required" before doing the parsing.

My use case is that I have defaults which are in a configuration file, so I'd like to be able to switch a field to non-required when a default value was saved in the config file

ymenager:
Oh, actually I can see that in 3.x it's possible to do that by manipulating the CommandSpec... yay ! closing non-issue :)

ymenager:
Actually i spoke too soon :-(

commandSpec.requiredArgs() is protected against modification, and there's no method that allows to remove something from that list. Would it be possible to make that modifiable ?

ymenager:
Alternatively it would be nice for required to support default values, and not fail if a default value exists (also usage would need to support that to not include it's marker when default value exists).

This solution architecturally make actually make the more sense (if there's a default value then required should always be fulfilled)

ymenager:
Ok, actually I've implemented that suggestion about allowing a field with a default value to never be required. I've made it optional through a parameter in @Command so that picocli to avoid changing current behavior.

See pull request #265

remkop:
@ymenager Very sorry for the late response. I must have accidentally switched off email notifications because I used to get emails for new issues but I did not for yours...

What you are saying makes sense. Options or positional parameters that have a value (either a user specified value or a default value) should not be the cause of a "missing required parameter" error. 

I still need to take a closer look at your pull request but my first thought is that it is okay to change the current behaviour to the previous paragraph. If so, the `@Command(notRequiredWithDefault)` attribute would not be needed, which would simplify things.

ymenager:
Sure I'm happy with it being the default behavior, I'm going to change the code in the PR branch accordingly

remkop:
ok cool. Is it possible to separate this pull request from #262 ?

ymenager:
Sure

ymenager:
Just as note, one consequence of this change is that all primitives will never be "required" since they always have a default value

remkop:
Thanks for pointing that out. That is not good...

It's getting a bit late for me so I need to sleep on it, but perhaps `@Option` and `@Parameters` need to have an additional attribute that provides the default value, so that picocli can verify that a default value has actually been specified.

This could be a simple `defaultValue = "..."` String attribute, and/or a `defaultValueProvider = Class<? extends IDefaultValueProvider>` attribute, where we introduce an additional interface, something like the below:

```java
[...]

PR #267, opened 2018-01-16
Various improvements to default values

ymenager:
Various improvements to default values (standard behavior remains the same, all features are options and must be enabled explicitely):

- bugfix: Default values couldn't be read from non-public fields (was missing setAccessible())
- mask default value: This allows to specify a mask string in @Options that will be displayed instead of any existing default value (will not be displayed if no default value exists)
- notRequiredWithDefault in @Command allows a default value in a required field to become non-required
- refreshDefaultValues() added to CommandLine

Unit tests for all features included


remkop:
Implemented, see comment on #261.

PR #270, opened 2018-01-19
Not required with default stdbehavior

ymenager:
As discussed in #261 , this makes not required with default value the standard behavior


remkop:
Implemented, see comment on #261

PR #271, opened 2018-01-22
Add closing token for wildcard in documentation

mkobit:


remkop:
Thanks for pointing this out! I will merge this when I get home from work.

remkop:
Merged. Thanks again!

Issue #159, opened 2017-08-04
Make usage help width configurable


remkop:
As suggested by @waacc-gh in #142:

> You could think about adding an additional attribute to the @command annotation and/or an additional parameter to the CommandLine.usage() methods to change the maximum allowed characters in a line. The improved usage() methods might be helpful to align the output to different windows/screen widths.

remkop:
There are two aspects to this: what should be the API, and can picocli autodetect the console screen width. 

API-wise, I imagine something like this:
```
// programmatically
CommandLine.setUsageWidth(int) : void
CommandLine.getUsageWidth() : int

// declaratively
Command.usageWidth(int);
```

The above would make it possible for command line applications to set a different usage help width than the default 80, but it still puts the burden on these applications to figure out what width to use (and what width is available).

----
In addition, picocli could make an effort to autodetect the console width and adapt the `usageWidth` to the console width. Some options for accomplishing this are documented here:
https://stackoverflow.com/questions/1286461/can-i-find-the-console-width-with-java

Autodetection should be explicitly enabled, likely with a new Command attribute:

```
Command.autoUsageWidth(true|false); // default false
```

pditommaso:
Auto-detection would be a plus, but the giving the ability the usage usage text width would be enough. I would go for a global setting `CommandLine.setUsageWidth(n)` that's applied to all commands (or sub-commands).

remkop:
Yes, sounds good. Do you think you can provide a pull request for this?

pditommaso:
I will try to give a look at the code base and see how much I'm confident with it.

remkop:
Cool. 

FYI, there’s currently a private field `usageHelpWidth` in the `Help` class. I’ve tried to take care to ensure that all alignment is driven off that field (but I could have missed something so needs tests).

remkop:
PR #289 merged into master and the 2.x branch.

System property `picocli.usage.width` can be used to specify a custom width.
The minimum width that can be configured is 55 characters.

PR #281, opened 2018-02-05
Make help width configurable

edgano:
Issue #159
void setUsageWidth(int value)
int getUsageWidth()

To make configurable the help width

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -303,6 +303,22 @@ public class CommandLine {
     }
 
     /**
+     * Set the usage help width 
+     * @param width the max usage text with 
+     */
+    public static void setUsageWidth(int width){
+        Help.usageHelpWidth=width;
+    }
+
+    /** 
+     * @return the current usage text width value
+     */
+    public static int getUsageWidth() {
+        return Help.usageHelpWidth;
+    }
+
+
+    /**
      * <p>
      * Convenience method that initializes the specified annotated object from the specified command line arguments.
      * </p><p>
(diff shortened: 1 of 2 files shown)


edgano:
I was trying to change the declaration of `Help.usageHelpWidth` to an instance field but the will need to change several other static methods like `join` and the `TextTable` class. Is that what you really want to do? also because there any value in doing that, since you cannot have more than usage width at time? do you agree on that?

remkop:
Hi, thanks for taking another look. (By the way, I thought I had put some feedback in a comment here earlier but I can't find it any more. Not sure what happened.)

How about this:
Let's make the default usage width configurable with a system property (`picocli.usage.width`). That gives users the ability to change the usage width while leaving our options open on the API side of things. I hesitate to commit to a programmatic API with static methods, and using a system property instead allows us to side-step that altogether and keep code changes to a minimum.

I would like to enforce a minimum usage width of 55. If a smaller value (or a non-numeric value) is specified, we log a warning and use the minimum instead. (The reason for this minimum is that option names take up 29 characters of each line, and the option descriptions need some minimum width to keep the result looking reasonably good.)

Do you think you can add tests that verify the resulting usage message? The `CommandLineHelpTest` class has many examples. It would be great if you could also add tests for edge cases (like checking that the line wrapping still works correctly with non-default width, and checking that the validation works for invalid system property values).

pditommaso:
> we log a warning and use the minimum instead

I don't any logging in picocli, what logging approach do you or do you want to use ?

remkop:
There’s a small Tracer class in picocli. At decision points the parser outputs messages at DEBUG, INFO or WARN level to help users troubleshoot. See http://picocli.info/#_tracing

If the user specifies an invalid value that will be ignored, there should be a warning.

Issue #242, opened 2017-11-21
Support for interactive shell using picocli


anshunjain:
Just started using picocli for a simple command line app, and find it amazingly easy to use and cool. I have also been experimenting with having interactive shell in my program which once launched will also allow execution of several subcommands. I am using cliche for this. Though it seems like picocli should be already able to handle this. Am I missing some obvious way of putting CommandLine in a loop to continuously parse for subcommands.

Happy to contribute if there is an agreeable change which could make it work that way.

anshunjain:
https://code.google.com/archive/p/cliche/
this is the cliche i was referring to

remkop:
Thank you for the positive feedback! Great to hear!

I think the only thing missing to accomplish a full interactive shell is a tokenizer that understands quotes. Basically, in your for loop you want to read in a single string from the console, and then parse it into separate arguments (like the shell does for you usually).

So, input like
```
<command> a b c "1 2 3" '4 5 6'
```
Is parsed into 5 tokens:
```
a
b
c
1 2 3
4 5 6
```
This can be a separate class from picocli.CommandLine.

Do you think you'll be able to submit a pull request for this?

PorygonZRocks:
I think that could be accomplished by looping over the output and combining stuff. I might give this a try at some point.

remkop:
Looking forward to any contributions!

FYI, I found the `java.io.StreamTokenizer` class extremely helpful when implementing @-file expansion. The following configuration is probably generally useful for parsing command line arguments:
```
StreamTokenizer tok = new StreamTokenizer(reader);
tok.resetSyntax();
tok.wordChars(' ', 255);
tok.whitespaceChars(0, ' ');
tok.commentChar('#');
tok.quoteChar('"');
tok.quoteChar('\'');
while (tok.nextToken() != StreamTokenizer.TT_EOF) {
    handleToken(tok.sval,  ...);
}
```

PorygonZRocks:
I have some code tested for quotes and escaping spaces, I just need to work on putting it into picocli. Before I do so, any code style rules I need to be aware of? And I could also probably make some sort of runInteractive method, should I do that as a separate method, a flag for normal run, an annotation, or what?

remkop:
Generally, try to stick to the style of the existing code. I may reformat so don't worry too much. 
Please include unit tests though. :-)

Also, can you show some examples of how this facility can/should be used?

PorygonZRocks:
[...]

PR #293, opened 2018-03-01
(WIP)  Add support for quoting parameters

PorygonZRocks:
This adds a class (`CommandTokenzer`) that takes text, and splits it based on customizable whitespace characters, and has support for quoting/escaping spaces with customizable quote and escape characters. The methods are intended to be similar to that of `StreamTokenizer` (hence the name). See [here](https://github.com/remkop/picocli/issues/242) for some more details. 

Note that currently the class is not used, as it is not finalized yet and (as I mentioned in the issue) I have to edit `CommandLine.java` with a normal text editor, not my IDE, so it is easier to keep it separate until completion.


src/main/java/picocli/CommandTokenizer.java
```diff
@@ -0,0 +1,371 @@
+package picocli;
```
remkop: Let's put this class in a new package `picocli.interactive`. I'm thinking that this package may grow in the future to contain additional functionality for interactive command line applications.

src/main/java/picocli/CommandTokenizer.java
```diff
@@ -0,0 +1,371 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Scanner;
+import picocli.CommandLine;
+import picocli.CommandLine.Option;
+import picocli.CommandLine.Parameters;
+
+public class CommandTokenizer {
```
remkop: Would be great to have some javadoc that describes what this class does, ideally with an example of how it can be used.

src/main/java/picocli/CommandTokenizer.java
```diff
@@ -0,0 +1,371 @@
...
+		// Characters that begin/end quotes
+		private List<Character> quoteChars = new ArrayList<Character>(Arrays.asList('"', '\''));
+		// Characters to be treated as whitespace
+		private List<Character> whitespaceChars = new ArrayList<Character>(Arrays.asList(' '));
+		// Characters to use to represent a new line
+		private List<String> eolPatterns = new ArrayList<String>(Arrays.asList("\r\n", "\r", "\n"));
+		// Remove blanks at start and end of a string
+		private boolean trimBlanks = true;
+
+		public CommandTokenizer(String cmd) {
```
remkop: Both constructors start parsing the specified input immediately before users have had a chance to configure the tokenizer. It may be better to postpone the call to `parse` until the user calls the `tokens` or `nextToken` method.

src/main/java/picocli/CommandTokenizer.java
```diff
@@ -0,0 +1,371 @@
...
+			this.tokens = parse(cmd);
+		}
+
+		public CommandTokenizer(Scanner input) {
+			this.tokens = parse(input);
+		}
+
+		private String[] parse(String cmd) {
```
remkop: Can we add unit tests for this method and the other public `parse` method? We can make this method package-private to allow unit tests to call it.

src/main/java/picocli/CommandTokenizer.java
PorygonZRocks: Oops, you're right. Didn't think about that, will fix.

src/main/java/picocli/CommandTokenizer.java
PorygonZRocks: Except this is useful for normal as well as interactive programs. I'm planning (if you're OK with it) to add the finalized version in as an inner class of `CommandLine`, similar to everything else, and to help with the whole "one file" thing. But if I add anything else for interactive stuff, I'll do that in a separate package (and pull request).

src/main/java/picocli/CommandTokenizer.java
PorygonZRocks: Good idea. I'm waiting until it's done though, in case we decide anything needs to be changed.

src/main/java/picocli/CommandTokenizer.java
PorygonZRocks: Sure. I've never done unit testing before, but for something like this, I think it should be simple. 
I might actually make them `public`, so that people can reuse the object. In that case, I would also add a no-arg constructor.

remkop: If you’ve never created unit tests before it will be a revelation. :-) Unit tests are a great way to shake out bugs and create confidence that future changes didn’t break anything. They are crucial to keeping software “soft”. 

Please see the picocli tests for examples.

src/main/java/picocli/CommandTokenizer.java
remkop: In normal usage the operating system shell will take care of the tokenizing and quoting so this is really for interactive use as far as I can see. There are all kinds of interesting things that can be done for picocli-based interactive shells, including things like autocompletion. I really think this should live in a new `picocli.interactive` package.

PorygonZRocks: Ok, will do. I wasn't sure if the OS handled all that or if picocli did (all my testing has been interactive, so I apparently had forgotten how non interactive apps worked 🙄).

PorygonZRocks:
Huh, somehow the javadocs broke, and that is failing Travis. Not sure how to fix it.

PorygonZRocks:
I think I included all the parameters and returns.

remkop:
The problem seems to be the `@see` javadoc tags. You probably should not surround the reference with curly braces. See the documentation for the [@see javadoc tag](url).

PorygonZRocks:
Currently, I only have unit tests for the `String` parse method, but I will add tests for the `Scanner` method.

PorygonZRocks:
...and the javadoc broke again 😤

remkop:
You'll get there! 😄

PorygonZRocks:
Any idea what broke the javadoc this time? There are a ton of errors in the log, and I've never used javadoc before.

remkop:
(Away from PC). Most of those are warnings, not errors. If you search for “error” you’ll quickly find it.

remkop:
I found a few:

* error: self-closing element not allowed `<br />`
* error: tag not allowed here: `<tbody>`
* error: reference not found `{@link #tokens()}`

PorygonZRocks:
I'm not sure how to fix the issue. During my manual testing, it should succeed. I'm not sure what JUnit is doing that I'm not (or vice-versa) that causes the issue. I'm pretty sure what is happening is it is splitting on the second space in the comments test, but when I manually run parse on it, it doesn't split there.

remkop:
The test failure says:
```
picocli.interactive.CommandTokenizerTest > testComments FAILED
    java.lang.AssertionError: array lengths differed, expected.length=1 actual.length=2
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.internal.ComparisonCriteria.assertArraysAreSameLength(ComparisonCriteria.java:76)
        at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:37)
        at org.junit.Assert.internalArrayEquals(Assert.java:532)
        at org.junit.Assert.assertArrayEquals(Assert.java:283)
        at org.junit.Assert.assertArrayEquals(Assert.java:298)
        at picocli.interactive.CommandTokenizerTest.testComments(CommandTokenizerTest.java:51)
```
Unsure why. You can try printing the result of the parse to stdout so you can see what the actual results are in the Travis CI log.

PorygonZRocks:
Yep, it's splitting on the second space. Like I said, I have zero idea why that would happen, whenever I try the test input, it works, but in JUnit it isn't.

remkop:
Could it have something to do with the difference in environment? Different behaviour when running on Linux?

PorygonZRocks:
Possibly? I'm not using JUnit, because AIDE does not support it. I'm using a mini interactive shell that returns the tokenizer output.

PorygonZRocks:
Maybe a difference between the JVM and the Dalvik VM (or ART, whatever), because AIDE works by dexing stuff?

PorygonZRocks:
The issue was that the code to trim blanks was in the Scanner method, not the String one. It's weird that switching to the String one in my manual testing didn't point that out, I must've still been using the Scanner one somehow.

remkop:
Good catch!

saaadel:
From:
https://github.com/remkop/picocli/issues/324#issuecomment-379919245

Feedback:
- Throws ArrayIndexOutOfBoundsException for empty command line or for command line with spaces only. I think it will be better to return empty array, ArgumentTokenizer works and returns zero length array.
- if we will have multiple spaces between params - we will get this as empty params in the array. Example: "abc xyz" (three spaces between params) -> Returns: [abc, , , xyz]. FYI ArgumentTokenizer skips those repeatable spaces

Also question:
Why you have two methods with separate implementation?
`new CommandTokenizer().parse(str)`
`new CommandTokenizer().parse(scanner)`

I mean, why you not replaced parse(str) implementation with short code:
```
parse(str) {
  return parse(new Scanner(cmd))
}
``` 
?

if in following code: 
`new CommandTokenizer(cmd).getTokens()`
you make something same (meta code):
```
input = new Scanner(cmd);
parse(input);
```

PorygonZRocks:
Will fix those issues, thanks!

Currently, the Scanner method actually calls the String method, it just has extra code to allow newlines to be escaped in an interactive shell. But I'll look into using less code for that, so that it is clearer that it is mostly the same.

saaadel:
do you plan to add symbol escaping in windows notation? I mean, caret symbol and something. For example: ^" - escaped quotes (instead slash notation for *nix)
Maybe in another method? or with optional flag in the current methods?

remkop:
Shouldn't this be possible in `CommandTokenizer` already by adding the caret `^` character to the `escapePatterns`, or am I missing something?

saaadel:
```
if (ch < 0) {
	this.quoteChars.clear();
}
```
IMHO it's dirty way, add another methods to clear lists.

saaadel:
@remkop 
> Shouldn't this be possible in CommandTokenizer already by adding the caret ^ character to the escapePatterns, or am I missing something?

Not so simple, he-he 😄 

See there: https://ss64.com/nt/syntax-esc.html
- simple chars:
`^^` = `^`
`^"` = `"`
`^&` = `&`
`^|` = `|`
etc.

- but percents
`%%` = `%`

- Escaping the pipeline: symbol escaped twice!  (for second command in the pipe, or more for next commands in the pipe). For example: `abc | echo ^^^&`
> When a pipe is used, the expressions are parsed twice. First when the expression before the pipe is executed and a second time when the expression after the pipe is executed. So to escape any characters in the second expression double escaping is needed:

`^^^&` = `&` (escaped twice if second command in the pipe)

PorygonZRocks:
I'm not dealing with pipes yet, that would probably be really  tough, because I think I'd have to change `System.out`, parse it to a `String`, and probably more. But just using a caret as an `eecapePattern` is already possible if you add it, as remkop said, and the rest seems to mostly be pipe related. The percent and exclamation points would require me to build them into the code, which would require yet another flag, and would not really be able to be customized, I don't think. I *might* be able to make a `Map` with "special escapes" or something, but not sure.

PorygonZRocks:
If you look, there are already methods that take a `boolean`, which clear and set to defaults (depending on if it is `false` or `true`, respectively). I had actually planned to remove the whole "wipe if under 0" thing, if you look I actually did remove it for `commentChar`, so I must've forgotten. Either way, the methods take a `char`, which is unsigned, so it can't ever even be under 0, that was when I was using `int`s (because `StreamTokenizer` does).

saaadel:
About flags, I think it will be better to move all of them to the factory, that returns actual parser via interface. So, we can extend the factory with custom flags, and to add new parser implementations for new shells

saaadel:
about `char`: I know that it was an old code in current context. I mentioned just because it is the dirty way in any solution

PorygonZRocks:
What do you mean "move them to factory"?

saaadel:
factory method pattern

remkop:
The factory pattern is great if you want to control which class is created, potentially to give different functionality in different classes. 

I’m not sure that’s what we need here. The current design is to control the behavior by configuring the parser before we use it, not by having many parser classes that work differently. 

I think the current design is a good fit for what we’re trying to do and there’s no need for a factory in this case.

remkop:
I’m a bit confused. I thought we were talking about the tokenizer parser. Are you talking about the `picocli.CommandLine.Interpreter` instead? That class is not published API.

remkop:
When I read POSIX parser I assumed you meant a parser for the [POSIX.1-2017 Utility Argument Syntax](url).

PorygonZRocks:
I should probably actually read that, I have just been doing this based off of my current shell knowledge and never even thought of the POSIX standard. 
Like I said, I'll look into some sort of special escapes, which could allow for some changes, and the variable number of escape characters would probably be implemented as an option somewhere else, if I end up creating pipe support  (a full interactive shell will probably take a while, to get stuff like pipes done). 
When everything is done for the tokenizer, etc, I'll probably create presets for Windows and POSIX syntaxes.

saaadel:
Yep, I meant about token recognition for POSIX shells.

remkop:
@PorygonZRocks  your approach sounds reasonable to me. Adding Presets to imitate the behaviour of specific OS shells would be a nice convenience layer on top of the tokenizer class.

PorygonZRocks:
That sounds a lot better, for some reason I had thought JLine was made with native libraries.

PorygonZRocks:
I may still look at creating a simpler way to use the two to create an interactive shell easily, but looks like this tokenizer isn't needed.

PR #413, opened 2018-07-14
resolve discrepancy between completionCandidates() and ${COMPLETION-CANDIDATES}

rpanak:


remkop:
Merged. Thanks for the PR!

Issue #22, opened 2017-02-07
i18n


remkop:
Ideas:
* @Option(..., i18nKey="myKey") -look up myKey in ResourceBundle
* @Option(..., description="A b c d") -look up A_b_c_d in ResourceBundle
* look up fully qualified field name in ResourceBundle

remkop:
The name of the resource bundle could be fixed or specified in `@Command(resourceBundle = "name")`.
Resource bundle keys could be based on the option name (either long or short), for positional parameters, the key could be the index.

This may be interesting for reference: https://github.com/tomitribe/crest#help-text

remkop:
Note on conflicts:

* Resource bundle set in a superclass are used if subclass doesn’t set its own resource bundle 
* If subclass does set its own bundle that one is used (an empty bundle essentially disables the superclass bundle)
* Bundle set by mixins are used unless the “host” command sets its own resource bundle (then the mixin bundle is ignored, even for the options and positional parameters contributed by the mixin)
* Bundle set by parent command are used for all subcommands, unless the subcommand sets its own resource bundle

PR #416, opened 2018-07-16
add support for @Command annotation on methods (in addition to classes)

illes:
allowing `@Command` annotation on methods cuts the boilerplate in situations where a class would have to be created just to allow integration with `picocli`.

Consider a simple _tail_ example:

```java
@Command(name="tail")
public class Tail implements Runnable {
    	
    @Option(names="-n", defaultValue="10")
    int n;
    	
    @Parameters()
    File[] files;

    public void run() {
        // ... do your thing
    }
}
```
versus 
```java
    @Command(name="tail")
    void tail(@Option(names="-n", defaultValue="10") int n, File... files) {
        // ... do your thing
    }
```

The attached commit is a minimal implementation and test cases for this feature.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -255,11 +255,9 @@ public class CommandLine {
         if (getCommand() instanceof Method) {
              throw new UnsupportedOperationException("cannot discover methods of non-class: " + getCommand());
         }
-        for (Method method : getCommand().getClass().getDeclaredMethods()) {
-            if (method.isAnnotationPresent(Command.class)) {
-                CommandLine cmd = new CommandLine(method);
-                addSubcommand(cmd.getCommandName(), cmd);
-            }
+        for (Method method : getCommandMethods(getCommand().getClass(), null).keySet()) {
+            CommandLine cmd = new CommandLine(method);
+            addSubcommand(cmd.getCommandName(), cmd);
         }
         return this;
     }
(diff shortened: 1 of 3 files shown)


kravemir:
Quite interesting idea... Can it be used like `app interface list`, `app interface <ID> up/down`?

illes:
@kravemir : if it is possible using the `master` branch, it should be doable with this feature too, as the processing of `@Option`/`@Parameters` annotations is the same. Do you happen to have a working example we could try to adapt?

illes:
@kravemir :  with a bit of tweaking, this example works even without having to use `@ParentCommand`:
```java
@Command(name="app", subcommands={App.Interface.class})
class App {

    @Command(name="interface")
    class Interface {

        @Parameters(arity="0..1", description="network interface")
        String iface;

        @Command void list() { /* ... */ }
        @Command void up() { /* use this.iface */ }
        @Command void down() { /* use this.iface */ }

        // ...
```
Results in:
```
Usage: app interface [<iface>] [COMMAND]
      [<iface>]   network interface
Commands:
  up
  down
  list
```
Is this the use case you were thinking about?

kravemir:
@illes It's almost what I though about. But, I would like something more like:

```
Usage: app interface list
```

Together with:

```
Usage: app interface [iface] [COMMAND]
      [iface]   (required) network interface
Commands:
  up
  down
  reset
  ...
```

illes:
@kravemir : short answer: no, picocli does not offer that parsing logic, and my proposed changes do not affect parsing

picocli goes something like:
- for `iface` to come before the `up`/`down` command, it has to be a `@Parameter` to the parent `@Command`
- `iface` can be either required or not controlled by the compile time constraint `arity`, hence you cannot decide to make it required based on runtime information (the value of `COMMAND`)

as a workaround, you could e.g.:
- throw a `(Parameter)Exception` in `list()` if `iface` is defined
- create two separate `CommandLine` instances and chain them (try `CommandLine.parse(..)` on the one with `list`, if that fails try the other one, before actually executing via `run()`/`call()` )

remkop:
This looks very interesting! Thanks for the PR!

I need a bit of time to finish up some other things ([1](https://github.com/micronaut-projects/micronaut-core/pull/400) and [2](https://github.com/micronaut-projects/micronaut-core/issues/330) ) I'm working on, but I'll get back to this end of this week.

illes:
@remkop : Thanks for reviewing. I implemented the name change from `run` to `invoke` (also adding an `Object` return value), and added reflection logic helper `CommandLine.getCommandMethods(...)` to the library, unit tests have been updated to reflect/cover these changes. 

One way to evolve would be to add a `boolean` flag to `@Command` that would trigger `addMethodSubcommands()` automatically.

remkop:
Thanks for updating! I’ll take another look. 

By the way, I found this library https://github.com/tomitribe/crest which has a similar style of defining commands with methods. 

Not intending to imitate, but it might be useful to see how they approached things or solved certain issues. Not sure if they support subcommands for example. Anyway, their docs may be interesting.

remkop:
Sorry for the wait! This has finally bubbled to the top of my todo list now.

Good point about adding some way to make command methods subcommands of the command class. Like you suggested, one way is to have a `addMethodSubcommands` boolean on `@Command`. 

A potential alternative way would be to simply always add command methods as subcommands of the command class that contains the method _if the class is annotated with `@Command`_. 

If the class is _not_ annotated with `@Command`, the methods annotated with `@Command` are top-level commands (and the desired command needs to be called from the main method with `CommandLine.invoke`).

What do you think?

Can I ask a favour: would it be possible to rebase on master? (If that's too much hassle I can do a merge, but I like the cleaner history of a rebase.)

illes:
I like the idea: `addMethodSubcommands=true` added, incl. javadoc and tests. Rebased on `master` squashing fixups.

remkop:
Sorry for the late response: traveling with family and have intermittent access to internet and PC. 

Thanks for the changes!

remkop:
@illes  I went ahead and merged the PR. Many thanks!

There are a few changes I would like your opinion on:

* Do we need the `@Command(addMethodSubcommands)` attribute? I was thinking to add method commands automatically as subcommands if the class has a `@Command` annotation.
* Similarly, I was thinking to make the `CommandLine#addMethodSubcommands` method private.
* I am not sure about mixing in options/positional params from the containing class `@Command` into the method commands. I would prefer not to do this.
* That said, it would be good if command methods can mix in the standard help options. Currently if you annotate a method with `@Command(mixinStandardHelpOptions = true)`, and invoke that method with `CommandLine.invoke("method", MyClass.class, "--help")` you will get a NullPointerException. The NPE is easily fixed by the mixinStandardHelpOptions still need to be mixed in for command methods.
* We need documentation for command methods. Could be after [15.1 Annotated Methods](https://picocli.info/#_annotated_methods) or perhaps a separate page. Users probably are interested in a recommendation on when to use which annotation style.

I have some work in progress to fix the NPE for methods with `@Command(mixinStandardHelpOptions = true)` and will push that soon.

remkop:
Also, #22 will dovetail nicely into this PR: that will allow users to specify description text in a separate properties file (or other resource bundle) to keep the method parameter annotations brief and readable.

PR #423, opened 2018-07-26
Require at least Java 7?

simon04:
Java SE Public Updates has ended for Java 5 and Java 6. According to http://www.oracle.com/technetwork/java/eol-135779.html, Extended Support for Java 6 will end in December 2018. Thus, I propose to update this project to require at least Java 7.

- [x] Update build.gradle
- [ ] Update documentation

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,8 +23,8 @@ buildscript {
 allprojects {
     apply plugin: 'groovy'
     apply plugin: 'java'
-    sourceCompatibility = 1.5
-    targetCompatibility = 1.5
+    sourceCompatibility = 1.7
+    targetCompatibility = 1.7
 
     repositories {
         jcenter()


remkop:
My thinking was that there may be users that are forced to use an old version of Java and compiling against Java 5 makes the library available to the widest possible audience. (This is quite common in corporate environments. I have seen systems in my previous company that were still on Java 4(!) for various reasons.)

The drawback is that contributors to picocli cannot use the latest Java APIs, and in some places reflection is necessary to make certain features available to users (type conversion to `Path` and the Java 8 `java.time` classes, for example).

I'm okay with that trade-off since the "pain" is small and it benefits the users.

I may consider upgrading if this equation changes, for example if the Travis Continuous Integration that I've set up for picocli stops supporting Java 5. 

Please let me know if there are other considerations that I've missed.

remkop:
I hope I've been able to explain why I hesitate to require later versions of Java at this stage.
Closing this PR for now, feel free to reopen if there are other considerations.

remkop:
I appreciate the research, but unless you can prove there are zero people on Java 5 this does not alter the trade-off. :-) Please read my prior comments.

bbottema:
I have read your previous comments (of course), but I disagree with your assessment.

> My thinking was that there may be users that are forced to use an old version of Java and compiling against Java 5 makes the library available to the widest possible audience. (This is quite common in corporate environments. I have seen systems in my previous company that were still on Java 4(!) for various reasons.)

I myself have worked exclusively for enterprise clients (financial, government, education) and I haven't seen Java 5 for the past six years, with the exception of my current client which moved from Java 5 to Java 7 two and a half years ago when they finally ditched Websphere for Tomcat (now on Java 8).

And that's the thing, in my professional experience it is generally the application containers that dictate the supported Java version, because of which enterprise environments are often stuck at older Java versions. At the same time web containers host web applications that generally don't implement a command line interface, because their outward facing interface is exactly defined by a servlet specification (often with the addition of a message broker like MQ or JMS).

> The drawback is that contributors to picocli cannot use the latest Java APIs

I wouldn't exactly count Java 6 among the latest APIs. It's _twelve_ years [old](https://en.wikipedia.org/wiki/Java_version_history#Java_SE_6) technology. It stopped receiving public support in 2013 and will stop extended commercial (enterprise) support before the end of this year.

I also don't know of any enterprise level company that willingly uses a Java version for which no security updates are released anymore (for Java 5 extended support was dropped in 2015). Especially in the highly regulated financial sector this is a big no-no (but maybe this is different in Japan?)

Ever since enterprise companies have been moving away from behemoths like Websphere and WebLogic world-wide (citation needed), adoption speed has increased tremendously. My current client went from Java 5 to 8 (in some places 9) within two years since they finally moved away from Websphere. Following the Java 5 adoption trend in the past, I expect Java 6 users to drop to zero even faster (certainly within one year). 

> Please let me know if there are other considerations that I've missed.

So let me reiterate. 
1. My own professional experience tells me Java 5 is not in use for years now, including enterprise environments.
2. Even if Java 5 is still used in enterprise environments, my experience is that this is in web containers, which host web applications, which don't implements CLI's.
3. The reports found by a quick Google support these findings

Java 6 seems like a safe choice.

RobertZenz:
@bbottema What would be gained by upgrading to 6? I mean, how exactly would the codebase change?

bbottema:
@RobertZenz 

Nothing _has_ to change, since Java has always been fully backwards compatible.

Java 6 didn't come with sexy language features, but it does come with official support until the end of the year. Java 6 also boosts a [performance increase](https://docs.oracle.com/javase/6/docs/technotes/guides/performance/speed.html), but this is more about the JVM rather than the compiler.

And then the world is moving on dropping support for Java 5 left and right. Hence issues like @simon04's.

RobertZenz:
@bbottema What's the point in upgrading then? I mean, if the codebase does not change it is not even upgrading, it's an artificial limitation of the target platforms with no gains. And just because the target is 5 does not mean that it is build with 5.

bbottema:
@RobertZenz 

> What's the point in upgrading then?

For Java 6, not much. I'm just saying it would be a safe choice and since I was asked for proof, I elaborated a bit :smiley:

Anyway, I'm waiting for Java 7 and 8 to become the new Java 6 and then we can have this discussion in a more meaningful way.

RobertZenz:
> I'm just saying it would be a safe choice and since I was asked for proof, I elaborated a bit 

Okay.

 > Anyway, I'm waiting for Java 7 and 8 to become the new Java 6 and then we can have this discussion in a more meaningful way.

Well, with the fast release cycle, such discussions will only become more cumbersome in the future...and more confusing...and more often...

remkop:
Here are the trade-off points I've identified so far for compiling/building on Java 5:

*Advantages*
# Allows users to use picocli on JVMs from Java 5 and higher 

*Disadvantages*
#  Limits language features contributors can use (no multi catch or lambdas)
# Contributors need to use reflection when using APIs introduced in Java 6 or later
# Cannot use build tools or quality assurance tools that require later versions of Java

Of these disadvantages, the first two don't carry much weight for me. The last one does, depending on the tool. I really like the Travis CI and Jacoco coverage that comes integrated with GitHub. If those stop working because of Java 5 I'll probably start building on a later version of Java. I haven't looked at other tools like errorprone. If they add significant value I could consider building on a later Java version.

bbottema:
CodeClimate has worked well for me.

remkop:
Thanks Benny, I'll take a look at CodeClimate.

remkop:
Thanks, will take a look!

PR #425, opened 2018-07-31
Fix autocomplete script errors in zsh

anthonykeenan:
I've been testing the autocomplete script in `zsh`, and I came accross a few errors.

* `shopt` isn't available in `zsh`
* `compopt` doesn't seem to be understood in `zsh` even when using `bashcompinit`, `complete` with the same options works, I've added an alias when using `zsh` for `compopt`->`complete`
* added `setopt COMPLETE_ALIASES` when using zsh or the alias doesn't work properly when autocompleting

I've attempted to fix them by adding a check for the shell the script is being run under. I've tested the changes using Ubuntu Server 18.04 LTS and zsh 5.4.2, and Mac OS and zsh 5.5.1.


remkop:
Merged and added credits in the release notes.
Thanks for the contribution!

Issue #441, opened 2018-08-12
Add support for custom paramLabel


bbottema:
Ok so I have almost everything sorted out, except for one limitation.

Here's what I'm using currently to define an OptionSpec:

```
OptionSpec.builder(cliDeclaredOptionSpec.getName())
                        .type(List.class)
                        .auxiliaryTypes(String.class)
                        .arity(String.valueOf(possibleOptionValues.size()))
                        .paramLabel(determineParamLabel(possibleOptionValues))
                        .build()
```
This is working well for me, except that for every arity the paramLabel is repeated.

The only way to currently solve this, is by defining the type as String with an arity of 0 or 1 depending on whether there are parameters at all or not:

```
OptionSpec.builder(cliDeclaredOptionSpec.getName())
                        .type(String.class)
                        .arity(possibleOptionValues.isEmpty() ? "0" : "1")
                        .paramLabel(determineParamLabel(possibleOptionValues))
                        .build()
```
This solves the paramLabel issue, but now I have an even bigger problem, because now I need to manually extract the param tokens from the single string parameter. This is rather complex, since whitespace can be a separator as well as a valid value (you would know!).

What would make the entire solution work for me if I could do the following:

```
OptionSpec.builder(cliDeclaredOptionSpec.getName())
                        .type(List.class)
                        .auxiliaryTypes(String.class)
                        .arity(String.valueOf(possibleOptionValues.size()))
                        .customParamLabel(determineParamLabel(possibleOptionValues)) // <----
                        .build()
```
customParamLabel then would be used as-is without repeating.

What do you think?

bbottema:
I already have a custom synopsis. I'll see if I can include the params there manually. Can I prevent the OptionSpec from being rendered in synopsis?

bbottema:
> You can supply a custom synopsis if the auto-generated synopsis becomes too detailed.

It is not only a problem in the synopsis, it is also under the _Options:_ listing:

![image](https://user-images.githubusercontent.com/2115718/44003119-f51ddd70-9e4d-11e8-9619-8f200ba3d04e.png)

remkop:
Thanks for that example, now I understand.

It looks like what you really need is Composite Parameters (#358).

I will look into how easy it would be to provide a `customParamLabel` feature.

[...]

PR #445, opened 2018-08-15
#441: added support for fixing the paramLabel as-is and omit regular processing

bbottema:
Here's a simple fix that allows users to choose to omit regular paramLabel processing and fix it to render as-is instead (ANSI codes notwithstanding).

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -3941,7 +3941,7 @@ public class CommandLine {
             // help-related fields
             private final boolean hidden;
             private final String paramLabel;
-            private final boolean fixParamLabel;
+            private final boolean showParamSyntax;
             private final String[] description;
             private final Help.Visibility showDefaultValue;
 
@@ -3970,7 +3970,7 @@ public class CommandLine {
                 description = builder.description == null ? new String[0] : builder.description;
                 splitRegex = builder.splitRegex == null ? "" : builder.splitRegex;
                 paramLabel = empty(builder.paramLabel) ? "PARAM" : builder.paramLabel;
-                fixParamLabel = builder.fixParamLabel;
+                showParamSyntax = builder.showParamSyntax;
                 converters = builder.converters == null ? new ITypeConverter<?>[0] : builder.converters;
                 showDefaultValue = builder.showDefaultValue == null ? Help.Visibility.ON_DEMAND : builder.showDefaultValue;
                 hidden = builder.hidden;
@@ -4075,8 +4075,8 @@ public class CommandLine {
             public String paramLabel()     { return paramLabel; }
     
             /** Indicates whether paramLabel should be processed the regular way or that it should be rendered as-is.
-             * @see #paramLabel()  {@link #fixParamLabel()} */
-            public boolean fixParamLabel()     { return fixParamLabel; }
+             * @see #paramLabel()  {@link #paramLabel()} */
+            public boolean showParamSyntax()     { return showParamSyntax; }
     
             /** Returns auxiliary type information used when the {@link #type()} is a generic {@code Collection}, {@code Map} or an abstract class.
              * @see Option#type() */
@@ -4245,7 +4245,7 @@ public class CommandLine {
                 private boolean required;
                 private boolean interactive;
                 private String paramLabel;
-                private boolean fixParamLabel;
+                private boolean showParamSyntax = true;
                 private String splitRegex;
                 private boolean hidden;
                 private Class<?> type;
@@ -4271,7 +4271,7 @@ public class CommandLine {
                     setter = original.setter;
                     hidden = original.hidden;
                     paramLabel = original.paramLabel;
-                    fixParamLabel = original.fixParamLabel;
+                    showParamSyntax = original.showParamSyntax;
                     required = original.required;
                     interactive = original.interactive;
                     showDefaultValue = original.showDefaultValue;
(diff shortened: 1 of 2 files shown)


src/main/java/picocli/CommandLine.java
```diff
@@ -4332,6 +4340,9 @@ protected int hashCodeImpl() {
                 /** Sets the name of the option or positional parameter used in the usage help message, and returns this builder. */
                 public T paramLabel(String paramLabel)       { this.paramLabel = Assert.notNull(paramLabel, "paramLabel"); return self(); }
     
+                /** Sets the name of the option or positional parameter used in the usage help message, and returns this builder. */
```
bbottema: Just noticed I didn't update this javadoc entry...

bbottema:
Note: I went with a boolean flag that changes how `paramLabel` works, rather than adding a `customParamLabel`, otherwise you would have to add more logic to manage priority between the two.

I feel two String properties that affect the same output if is more difficult to explain to the library users than explicitly modifying a single existing String property with a boolean flag.

remkop:
One idea is to call the flag ‘paramLabelDecorations‘ or ‘paramLabelIndindicators‘ (default ‘true‘).

bbottema:
I'm fine with either, although personally I do prefer to use a _verb_ that describes what it does. So something like like `renderParamLabelDecorations `or `renderParamLabelIndindicators` (or `applyXYZ`)

remkop:
Yes, that’s good style. I just like starting with `paramLabel` so they show up together in the IDE. But maybe overthinking it.

bbottema:
Good point. Perhaps something like paramLabelShouldHaveDecorations, but maybe now I am over thinking it :smiley:

src/main/java/picocli/CommandLine.java
remkop: If you push additional commits to your branch, they will show up in the PR...

src/test/java/picocli/CommandLineModelTest.java
```diff
@@ -66,11 +66,14 @@ public void testModelUsageHelp() throws Exception {
...
         spec.addOption(OptionSpec.builder("-V", "--version").versionHelp(true).description("show help and exit").build());
         spec.addOption(OptionSpec.builder("-c", "--count").paramLabel("COUNT").arity("1").type(int.class).description("number of times to execute").build());
+        spec.addOption(OptionSpec.builder("-f", "--fix").paramLabel("FIXED(=BOOLEAN)").arity("1").fixParamLabel(true).required(true).description("run with fixed option").build());
         CommandLine commandLine = new CommandLine(spec);
         String actual = usageString(commandLine, Ansi.OFF);
         String expected = String.format("" +
-                "Usage: <main class> [-hV] [-c=COUNT]%n" +
+                "Usage: <main class> [-hV] [-c=COUNT] -f FIXED(=BOOLEAN)%n" +
                 "  -c, --count=COUNT   number of times to execute%n" +
+                "  -f, --fix FIXED(=BOOLEAN)%n" +
```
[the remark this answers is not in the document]
remkop: A hardcoded space is incorrect. Picocli currently inserts the value of `commandSpec.parser().separator()`. This behaviour needs to be preserved.

remkop:
As for the name of this switch, I'm thinking `showParamSyntax` may be a good name.

bbottema:
Applied the changes as discussed. The master build is failing though, so this fails as well since I updated the branch to master...

remkop:
Thanks. Changes look good. We should also add this attribute to the `@Option` and `@Parameters` annotations.

remkop:
What do you think of the idea to flip this around and call it `hideParamSyntax` (default `false`) instead? That makes it slightly more clear that this is shown by default.

remkop:
@bbottema  I think we are almost there. There is one test that fails now because the default changed from true to false. Will you be able to add the `hideParamSyntax` attribute to the `@Option` and `@Parameters` annotations?

bbottema:
Took care of the junit test, but I don't have time left to work on it for the moment. Got caught up in a nasty double byte character set [bug](https://github.com/bbottema/outlook-message-parser/issues/3) in a RTF to HTML converter 😮

bbottema:
I did what I could aside from the annotations stuff.

I'm calling it a day on this one because frankly speaking I find it difficult to learn the layers and relationships between the classes having to scroll all the time between them over almost 10k LOC (no overview). I tried a couple of times, but kept putting it down in a mild exasperation. File search is not helping much either since a word like "paramLabel" occurs 82 times when limited to whole words only in just the CommandLine.java, not counting comments. It doesn't help that even IntelliJ has buffering issues with this much code (and I have a monstrous custom built laptop).

Anyway, sorry for the rant. I really like what picocli is capable off, but it is too tedious for me personally to learn the intricacies of the library for a newcomer. I have half a mind of forking it and restructuring things a bit for my own use...

remkop:
Sorry for the late response: traveling with family, intermittent access to internet and PC. 

Thanks for the contributions so far! Picocli is a better library because of your work. 

I realize that the single source file is becoming a stumbling block for contributors. I’ll investigate whether it’s possible to do something in the build: have a Gradle task that picks the source apart into smaller chunks, let people work with these (ideally run the tests against them) and eventually recombine into a single source file again.

remkop:
I need a bit more time before I can get back to this PR.

Meanwhile, I added some class diagrams (https://github.com/remkop/picocli/tree/master/src/main/java/picocli/doc-files). Hope this is useful.

PR #448, opened 2018-08-17
Brief update to docs.

kalexmills:
Added comment about `@Unmatched` fields remaining `null` if no arguments are left unmatched.


remkop:
Thanks! Merged with slight modification. 
(Also updated index.adoc from which index.html is generated.)

PR #462, opened 2018-09-02
Fix option handling in autocompletion scripts

### LINKED ISSUE #461: Script auto-completion only ever completes options and never default bash completions
I have generated the bash completion script for my command. I have a command that looks like:

`command [OPTIONS] FILE...`

If I type `$ command <TAB>` I get `$ command -` and then hitting TAB twice to get `$ command -<TAB><TAB>` lists all of the short and long options. The proposed completions for the command options themselves look correct.

However, the completion only ever tries to complete the command options (as they begin with a `-`), and therefore, it never does not get a match and thus never falls back to the default completions.

It should be listing the default case which is files/directories and not supply a `-`. It should only try to complete the command options after I manually type a `-` (or `--`) on the command line. This is how (all) other command completions in bash seem to work.

Also, it would be nice to have `--<SPACE>` force it stop completing command options and fall back to the default competitions as this matches the behavior of the actual command.

dwalluck:
Closes #461


remkop:
I tested and it looks good.
Merged the PR.
Thank you very much!

Issue #321, opened 2018-04-03
Add support for IDefaultValueProvider 


remkop:
Add support for `IDefaultValueProvider` on `@Command` and potentially on individual `@Option` and `@Parameters`.

See discussion under #261.

Proposed API:
```java
interface IDefaultValueProvider {
    /** Returns the default value for an option or positional parameter or {@code null}.
     * The returned value is converted to the type of the option/positional parameter
     * via the same type converter used when populating this option/positional
     * parameter from a command line argument.
     * @param argSpec the option or positional parameter, never {@code null}
     * @return the default value for the option or positional parameter, or {@code null} if
     *       this provider has no default value for the specified option or positional parameter
     */
    String defaultValue(ArgSpec argSpec);
}
```

remkop:
Note to self: this ticket can/should be implemented before #322. Given the discussion on #280, there are some trade-offs with a `defaulValue` annotation attribute, while the benefits of this ticket are more clear.

remkop:
API-wise, I think the below is sufficient for the first cut:

```
public @interface Command {
  Class<? extends IDefaultValueProvider> defaultValueProvider() default NoDefaultProvider.class;
  // ...
}
private static class NoDefaultProvider implements IDefaultValueProvider {
    public String defaultValue(ArgSpec argSpec) { throw new UnsupportedOperationException(); }
}

public static class CommandSpec {
  public IDefaultValueProvider defaultValueProvider() { return defaultValueProvider; }
  public CommandSpec defaultValueProvider(IDefaultValueProvider  defaultValueProvider) {
    this.defaultValueProvider = defaultValueProvider;
    return this; 
  }
  // ...
}
```

No need to have an `defaultValueProvider` annotation attribute on @Option and @Parameters.

NicolasMassart:
If you can help me, I have difficulties to see how to retrieve the CommandSpec (to get my default value provider) from within the ArgSpec defaultValue() method...

remkop:
I was thinking to leave `ArgSpec.default()` untouched, and only add logic to the method in `Interpreter` where the default values are set before parsing the user input. Away from PC, from memory I think it was `setDefaultValue(ArgSpec)`. I was thinking to also pass the `IDefaultProvider` to that method.

NicolasMassart:
Ok, something like that ?

```
        private void applyDefault(IDefaultValueProvider defaultValueProvider,
            ArgSpec arg, List<ArgSpec> required) throws Exception {

[...]

PR #463, opened 2018-09-04
Fixed default provider initial and default values overriding

NicolasMassart:
Fixed misbehaviour that was introduced in PR #458, see #321
Now default provider overrides default or initial values only if the return
value of the provider is not null.


remkop:
Merged. Thank you!

NicolasMassart:
You're welcome.
Now the doc part...

PR #469, opened 2018-09-07
Fix option argument handling in autocompletion scripts

### LINKED ISSUE #466: Completion script should not generate suggestions for options with arguments that have no known completions
Follow-up issue resulting from discussion on #461:

options that take an argument for which no candidates are available (e.g., non-enum) should not generate any candidates. The current (incorrect) behaviour is to show subcommands or the default completions.

dwalluck:
Closes #466.


remkop:
Thank you! I hope I’ll be able to review this tonight.

remkop:
Merged. Thank you!

Issue #483, opened 2018-09-17
Auto detect Jansi


charphi:
It would be usefull to make `Ansi.AUTO` detect if Jansi is available and enabled at runtime.
Thus, we wouldn't have to specify it manually.

For example:
```java
AnsiConsole.systemInstall(); // Jansi magic
CommandLine.run(new WindowsJansiDemo(), args);
AnsiConsole.systemUninstall();
```
instead of
```java
AnsiConsole.systemInstall(); // Jansi magic
CommandLine.run(new WindowsJansiDemo(), System.err, Ansi.ON, args);
AnsiConsole.systemUninstall();
```

remkop:
Makes a lot of sense. 

What is a good way to detect that Jansi is installed? (I'm not sure that it is enough just to check that the Jansi jar is on the classpath...)

charphi:
It seems that you just need to compare AnsiConsole.out and System.out such as:
`AnsiConsole.out == System.out`

Here is a quick example code:
```java
static boolean isJansiEnabled() {
	try {
		Class<?> ansiConsole = Class.forName("org.fusesource.jansi.AnsiConsole");
		Field out = ansiConsole.getField("out");
		return out.get(null) == System.out;
	} catch (Exception ex) {
		return false;
	}
}
```

remkop:
Perfect!

Do you feel like providing a PR for this?

remkop:
#486 has been merged. Thanks for the contribution!

PR #486, opened 2018-09-17
Add Jansi auto detection.

charphi:
See #483


remkop:
Looks good!
Do you think you can add a unit test for this?

charphi:
I've added a test but it doesn't increase the code coverage.
I don't know how to properly test this since it depends on runtime context.

remkop:
@charphi  Sorry for my late response. Thanks for adding a test! I'll try to take a closer look later today.

PR #493, opened 2018-09-22
Added section 13.3 Subcommand aliases, and increased numbering of sub…

marinier:
…squent 13.x sections.


remkop:
The html file is generated from `index.adoc` during the build. (Execute `gradlew ascii` to omit other build tasks than generating docs.)

Can you modify the `index.adoc` instead (or in addition)? Otherwise your changes will be lost with subsequent updates.

marinier:
Sorry about that; I updated index.adoc (much easier!). For some reason, the `gradlew ascii` command doesn't seem to be regenerating index.html for me -- gradle seems to run fine (no errors), just no new output. Maybe it's because I'm running on Windows 10? Anyway, hopefully I got the format right even though I couldn't test the rendering.

marinier:
Nm, I wasn't looking in the build directory. It rendered fine.

Issue #468, opened 2018-09-07
Add support for completing positional parameters in completion script


remkop:
Follow-up issue resulting from discussion on #461:

add support for completing positional parameters in addition to named options. This may be challenging as it requires the completion script to determine the position index in user input that contains a mixture of subcommands, positional parameters, named options and option parameters.

dwalluck:
I would like to clarify something. Does picocli support subcommands inside other subcommands? How does picocli handle subcommands when there is also an argument to the parent command?

If I have a subcommand and a first parameter, it always seems to print "Missing required parameter".  I thought you had said on another issue that if there is a subcommand and a required argument in the parent command at position one that it would allow this and try to match the subcommand first.

I implemented a bash-completion script that mixes the completions for the commands with the completions with the first argument, but picocli itself doesn't seem to be able to handle it.

Would it be better to submit the pull request with the bash-completion generator even if it is not finished? I have a test program which is exhibiting this behavior.

remkop:
> Does picocli support subcommands inside other subcommands? 
Yes, nested subcommands are supported to any depth.

> How does picocli handle subcommands when there is also an argument to the parent command?
The parser will first try to match options, then subcommands, then positional parameters and finally treat the argument as unmatched.

> If I have a subcommand and a first parameter, it always seems to print "Missing required parameter". I thought you had said on another issue that if there is a subcommand and a required argument in the parent command at position one that it would allow this and try to match the subcommand first

It will try to match the subcommand first, and if successful the subcommand ends up consuming the remaining arguments. This then leaves the parent command without a value for its required parameter, hence the error. Users may invoke a subcommand but that does not mean they can omit required parameters. It sounds like this is the expected behaviour.

dwalluck:
Do you have bash-completion installed? It's probably best if you just try #502. No matter what number of parameters I pass, it seems to not parse successfully.

It does not support `Map` yet, but I think it supports all of the other types that it can support.

remkop:
I would expect this input to pass without "Missing required parameter" error:

```
java BashCompletionAutoCompleteTestApp file1 file2 bar -b baz foo
java BashCompletionAutoCompleteTestApp file1 file2 foo -f 255.255.255.255
```

I noticed one small this: the `Foo` command has two positional parameters but they have the same index (the default, which is `"0"`). I think you want to give them separate positional indices:

```
@Parameters(index = "0", arity = "1", paramLabel = "HOST_ONE", description = "SRC")  
[...]

PR #502, opened 2018-09-28
Add bash-completion file generator

dwalluck:


src/main/java/picocli/BashCompletionAutoComplete.java
```diff
@@ -0,0 +1,656 @@
...
+
+        return buff.toString();
+    }
+    private static void generateEnumCompetionCandidates(StringBuilder buff, ArgSpec f) {
+        if (f instanceof OptionSpec) {
+            buff.append("\n");
+            buff.append(format("    local %s_option_args\n", bashify(f.paramLabel().toLowerCase(ENGLISH))));
+            buff.append(format("    %s_option_args=\'%s\'\n",
+                    bashify(f.paramLabel().toLowerCase(ENGLISH)),
+                    concat(" ", Arrays.asList((Enum[]) f.type().getEnumConstants()), null, new EnumNameFunction()).trim()));
```
remkop: Doesn't `f.completionCandidates()` work also for ArgSpecs with an `enum` type? I thought that was fixed (so you should be able to combine `generateEnumCompetionCandidates` and `generateNonEnumCompetionCandidates` in one method).

dwalluck: I've removed it, although this code was copied from the original generator.

dwalluck:
The built-in `_get_first_arg` function doesn't seem to work if there is an option that takes an argument as this throws the index off by one every time it appears

remkop:
I guess it expects option values to be attached with the `=` separator:

```
<cmd> --option=value positional1 positional2
```

dwalluck:
> <cmd> --option=value positional1 positional2

This seems to make `_get_first_arg` work (it proposes the subcommand completions), but the index still appears to be off for the other args.

remkop:
I will try to take a look tomorrow but I to be honest I don't like having a dependency on `bash-completion`.

dwalluck:
> I will try to take a look tomorrow but I to be honest I don't like having a dependency on `bash-completion`.

It's an entirely separate class from the other generator.

Also, it should technically be possible to reimplement any needed functions and use those as a fallback assuming that the original functions work properly.

remkop:
@dwalluck Since the original driver for this feature (Add support for completing positional parameters in completion script #468) has been implemented in picocli 4.2, I am thinking to close this PR. 
Any objections?

dwalluck:
@remkop I recently saw #1097, so I was wondering if you're against bringing this back as a separate option so that the user could choose the pure-bash option or the bash-completion option with requires a dependency on https://github.com/scop/bash-completion. As this PR is almost two years old, I'll have to revisit what issues are not solved by the current pure-bash code.

remkop:
@dwalluck I much prefer not to have any dependencies. 

As part of the [4.4 milestone](https://github.com/remkop/picocli/milestone/66) I am planning to introduce completion scripts for several other shells (zsh, fish, clink and PowerShell being the main ones). So I’m definitely interested in improving this area.

When it comes to Bash, I thought 4.2 fixed all the issues we discussed, but I am interested in understanding what issues are not solved by the current solution.

dwalluck:
> @dwalluck I much prefer not to have any dependencies.

I am not sure I understand: It is not a dependency that picocli itself would inherit, but a dependency that lives on the user's system. I would think that it's not any more of a dependency than say, a Powershell user being required to have Powershell on his system.

> When it comes to Bash, I thought 4.2 fixed all the issues we discussed, but I am interested in understanding what issues are not solved by the current solution.

To be honest, I don't recall if I fully tested it, nor do I know for sure if all my issues made it into separate tickets. It's my fault for not having the time to monitor it closer. That was only one particular issue.

There may be other benefits of bash-completion in that it has some functions for completing some things like files, directories, hostnames, network interfaces, etc. (some of these are likely Linux-specific). Whenever I ran into an issue, I just saw myself re-implementing these features from scratch into the existing auto-completion code.

dwalluck:
> As part of the [4.4 milestone](https://github.com/remkop/picocli/milestone/66) I am planning to introduce completion scripts for several other shells (zsh, fish, clink and PowerShell being the main ones). So I’m definitely interested in improving this area.

Just some suggestions:

Perhaps having a common interface that each completer has to implement which satisfies various required features. This would keep the completion scripts separate, but would provide some common code or at least a way to tell what features they support.

Also, is there any way of testing the scripts themselves against the expected behavior? Often, a user will raise a new ticket about some bug or missing feature in the completion code, but it's not really clear how to tell if existing picocli features line up with the completions, especially as new picocli features get added.

remkop:
>>@dwalluck I much prefer not to have any dependencies.
>
>I am not sure I understand: It is not a dependency that picocli itself would inherit, but a dependency that lives on the user's system. I would think that it's not any more of a dependency than say, a Powershell user being required to have Powershell on his system.

I see what you are saying; picocli could support zsh, fish, clink, powershell, bash and bash-with-bash-completion, instead of just zsh, fish, clink, powershell, and bash. Perhaps calling this a dependency is the wrong way to look at this. That said, I am still not convinced that the additional functionality (if any) would offset the maintenance burden of essentially supporting another shell.

> There may be other benefits of bash-completion in that it has some functions for completing some things like files, directories, hostnames, network interfaces, etc.

The current solution already uses Bash built-ins to generate different completions for options and positional parameters of type `File`, `Path` and `InetAddress`, but if there are ways to improve this, I am interested!

>> As part of the 4.4 milestone I am planning to introduce completion scripts for several other shells (...)
>
> Perhaps having a common interface(...)

To be honest, I haven't started looking at this in detail, but I hope to be able to share a lot of the code among the different target shells.

> Also, is there any way of testing the scripts themselves against the expected behavior?

Yes! The picocli [4.2](https://github.com/remkop/picocli/releases/tag/v4.2.0) release added [tests for the completion scripts](https://github.com/remkop/picocli/tree/v4.2.0/src/test/dejagnu.tests).
These are written in Expect, and only work if Expect and DejaGnu are installed. Coming from the Java/JUnit world, this is a bit different and it took me some time to wrap my head around this, but it is very useful. It allows us to actually test that the completion script produces the expected completion candidates, given some partial user input, and the user hitting `<TAB>`. Going forward I plan to use this for testing completions where possible.

yschimke:
@remkop I'm happy to review any PRs for zsh.  I ended up liking zsh completion and getting a nice working file

https://github.com/yschimke/okurl/blob/master/zsh/_okurl

For a picocli annotated class https://github.com/yschimke/okurl/blob/master/src/main/kotlin/com/baulsupp/okurl/commands/CommandLineClient.kt

PR #510, opened 2018-10-09
Fix doc link for moved example files

anthonykeenan:
Noticed there were a couple of broken links on the website because the example files directory has been renamed.


remkop:
Thanks for the PR!

PR #523, opened 2018-10-21
Fill array values before setting them as value

knutwalker:
When using setter methods for array options, the values aren't set when the setter is invoked, making it impossible to do any kind of processing on the values other than storing the array, as every value is null.


knutwalker:
Sure, I've rebased the PRs

knutwalker:
Sure, I've rebased the PRs

PR #532, opened 2018-10-28
Add a way to get the argument attempted to be overwritten from an OverwrittenOptionException

acid1103:
Provides a nice way to get the `ArgSpec` of the option which was being overwritten before throwing an `OverwrittenOptionException`. The `ArgSpec` is retrieved through `OverwrittenOptionException#getOverwritten()`, which can be taken advantage of in a custom `IExceptionHandler2` to automate exception handling, generate custom error messages, etc.

I was in the process of writing a custom `IExceptionHandler2` and noticed some `ParameterException`s have convenience methods for accessing the root of the exception (see `MissingParameterException#getMissing()`) and thought `OverwrittenOptionException` was missing such a method. It's a tiny pr, but seems useful to me at least. Thanks for the consideration.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -7104,7 +7104,7 @@ public class CommandLine {
             if (initialized != null) {
                 if (initialized.contains(argSpec)) {
                     if (!isOverwrittenOptionsAllowed()) {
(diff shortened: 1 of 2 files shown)


remkop:
Thanks for the pull request!

PR #547, opened 2018-11-12
Print help infinite loop fix

patrickkuo:
Print help run into infinite loop when length of the text equals to `columnValue.maxLength` and offset is non zero.


remkop:
Merged.
Good catch! Thank you very much!

PR #553, opened 2018-11-21
fix(doc): link to CommandLine.java

simon04:
When viewing https://picocli.info/, the link resolves to the non-existing https://picocli.info/src/main/java/picocli/CommandLine.java


remkop:
Thank you for the pull request!
I merged it into master. Will update the link in the generated HTML when I get home.

remkop:
Finally updated the generated HTML as well... Sorry for the delay.

Issue #551, opened 2018-11-19
Support JCommander-style parameter files


triceo:
It is possible that this is a feature, but I did not find it in the documentation. See [my Github repo](https://github.com/triceo/picocli-bug-windows-path) for a complete reproducer. 

## Steps to reproduce

Create an argument file `argFile` such as this:

    -f "C:\Users\me\somefile.txt"

Run Picocli like this:

    CommandLine.run(..., "@argFile");

Observe that the value of the `String`-typed `Option` has been changed to `C:UsersmesomeFile.txt`.

## Additional information

- When the value is not quoted, everything works fine. This suggests a problem with not escaping the Windows backslash character. However, the quotes are used in order not to have to escape whitespace characters and therefore this workaround creates a different problem - either I escape the backslashes or the whitespace.
- When the value is kept quoted but passed directly on the command line (as opposed to the argument file), the problem no longer shows up. Unfortunately, the code will include the quotes in the resulting value. There is a second test included to showcase that problem.

remkop:
I have reproduced the problem. This seems to be the behaviour of `java.io.StreamTokenizer` that is used internally: it treats quoted Strings as if they are String literals in Java source code, where `\` characters need to be escaped with another `\`. 

This is not great. Looking into solutions...

remkop:
About the second issue you mention, that `"` quote values are preserved when specified on the command line: this is [by design but is configurable](https://picocli.info/#_quoted_values). 

You can switch this off with `commandLine.setTrimQuotes(false)`.

remkop:
Sorry I meant `commandLine.setTrimQuotes(true)`.

remkop:
Note to self: consider replacing use of `java.io.StreamTokenizer` with logic from https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/999dbd4192d0f819cb5224f26e9e7fa75ca6f289/src/jdk.compiler/share/classes/com/sun/tools/javac/main/CommandLine.java

remkop:
Hmm... I tested with the `Tokenizer` class used in Java 11 but it behaves the same as picocli: the 2nd argument in your example is parsed as `C:UsersmesomeFile.txt`.

I think it makes sense for picocli's @-file behaviour to be consistent with Java's @-file behaviour, what do you think?

remkop:
Thanks for the background. Let me check how JCommander does it.

remkop:
Looks like JCommander requires every argument to be on a separate line in the @argument-file.
[...]

PR #558, opened 2018-11-30
Add first failing test for #551

triceo:
The test for #551 is failing on `List`-based arguments. Before I move on with writing extra coverage, we should probably agree on how to deal with this. WDYT @remkop?

### DIFF
--- a/src/test/java/picocli/CommandLineTest.java
+++ b/src/test/java/picocli/CommandLineTest.java
@@ -58,15 +58,7 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.contrib.java.lang.system.ProvideSystemProperty;
 
-import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNotSame;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertSame;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
 import static picocli.CommandLine.Command;
 import static picocli.CommandLine.DuplicateOptionAnnotationsException;
 import static picocli.CommandLine.Help;
@@ -104,8 +96,10 @@ public class CommandLineTest {
     @Rule
     public final ProvideSystemProperty ansiOFF = new ProvideSystemProperty("picocli.ansi", "false");
 
-    @Before public void setUp() { System.clearProperty("picocli.trace"); }
-    @After public void tearDown() { System.setProperties(originalProperties); }
+    @Before
+    public void setUp() { System.clearProperty("picocli.trace"); }
+    @After
+    public void tearDown() { System.setProperties(originalProperties); }
 
     @Test(expected = NullPointerException.class)
     public void testConstructorRejectsNullObject() {
--- a/src/test/java/picocli/CommandLineTest.java
+++ b/src/test/java/picocli/CommandLineTest.java
@@ -3656,6 +3656,29 @@ public class CommandLineTest {
     }
 
     @Test
+    public void testAtFileSimplifiedWithQuotesTrimmed() {
+        System.setProperty("picocli.useSimplifiedAtFiles", "true");
+        class App {
+            @Option(names = "--quotedArg")
+            private String quoted;
+
+            @Option(names = "--urlArg")
+            private URL url;
+
+            @Option(names = "--unescapedBackslashArg")
+            private String unescaped;
+        }
+        File file = findFile("/argfile-simplified-quoted.txt");
+        final App app = new App();
+        final CommandLine cli = new CommandLine(app);
+        cli.setTrimQuotes(true);
+        cli.parse("@" + file.getAbsolutePath());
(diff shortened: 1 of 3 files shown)


src/test/java/picocli/CommandLineTest.java
```diff
@@ -59,7 +52,21 @@
...
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
```
triceo: IDEA be damned. I'll fix this.

src/test/java/picocli/CommandLineTest.java
```diff
@@ -3619,6 +3628,38 @@ public void testAtFileExpandedAbsolute() {
         assertEquals(Arrays.asList("1111", "2222", ";3333"), app.files);
     }
 
+    @Test
+    public void testAtFileSimplified() {
+        System.setProperty("picocli.useSimplifiedAtFiles", "true");
```
triceo: I can do this here, because of the change above.

triceo:
Also added a test to see if the simplified files work with quote trimming.

src/test/java/picocli/CommandLineTest.java
```diff
@@ -3619,6 +3623,61 @@ public void testAtFileExpandedAbsolute() {
...
+            @Option(names = "--urlArg")
+            private URL url;
+
+            @Option(names = "--unescapedBackslashArg")
+            private String unescaped;
+        }
+        File file = findFile("/argfile-simplified-quoted.txt");
+        final App app = new App();
+        final CommandLine cli = new CommandLine(app);
+        cli.setTrimQuotes(true);
```
triceo: @remkop Any chance this could also be set through a system property? This way, it prevents Picocli to be used in the simple way: `CommandLine.call(...)`

remkop: Yes sure. Do you want to provide a separate pull request for that?

remkop:
Thanks for the PR! It’s my bedtime tonight, I’ll take a look tomorrow.

src/test/java/picocli/CommandLineTest.java
triceo: I'll take a look in another PR.

remkop:
I took a look just now. The approach looks good, but the tests fail in the Travis CI build. Can you make the tests pass?

```
picocli.CommandLineTest > testAtFileSimplified FAILED
    picocli.CommandLine$UnmatchedArgumentException: Unmatched argument: yet something else
        at picocli.CommandLine$Interpreter.parse(CommandLine.java:6759)
        at picocli.CommandLine$Interpreter.parse(CommandLine.java:6615)
        at picocli.CommandLine.parse(CommandLine.java:722)
        at picocli.CommandLine.populateCommand(CommandLine.java:675)
        at picocli.CommandLineTest.testAtFileSimplified(CommandLineTest.java:3649)

picocli.CommandLineTest > testAtFileSimplifiedWithQuotesTrimmed FAILED
    java.lang.AssertionError: expected: java.lang.String<https://picocli.info/> but was: java.net.URL<https://picocli.info/>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:118)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at picocli.CommandLineTest.testAtFileSimplifiedWithQuotesTrimmed(CommandLineTest.java:3677)
```

remkop:
Sorry, I just re-read and saw your first comment. So, you are asking, if I have an option like this:

```
@Option(names = "-x") List<String> values;
```
and I give it multiple values, like `-x a b c`, why does picocli not add all values "a", "b" and "c" to the list but gives an 'Unmatched argument: b" error?

The answer is that [default arity](https://picocli.info/#_default_arity) for arrays and collection options is `1`. If an option needs to consume multiple arguments, you can specify the exact desired arity:

```
// consume 1 or 2 args for each -y option
// for example: <cmd> -y A -y A B -y E -y F G
// gives y = [A, A, B, E, F, G]
@Option(names = "-y", arity = "1..2") List<String> y;

// consume as many args as possible until the next option
// for example: <cmd> -z A B C D E -y Y
// gives z = [A, B, C, D, E] and y= [Y]
@Option(names = "-z", arity = "1..*") List<String> z;
```

triceo:
@remkop Thanks, I completely forgot about the arity. The test is fixed now.

Did you have any more tests in mind? If so, please share - otherwise I'll move on to creating the PR wrt. the quote-trimming property.

remkop:
I’m happy to merge this PR when the tests pass.

remkop:
Merged. Thanks for the PR!
I’ll update the release notes later.

remkop:
FYI: I created #561 as a more strategic solution for improving the convenience methods.

triceo:
@remkop Thanks for merging this PR, I also submitted #562 in the meantime. (Although #561 will definitely be useful and will supersede my solution.)

PR #562, opened 2018-12-01
Allow for enabling quote trimming via a system property

triceo:
The property is `picocli.trimQuotes`.

When set to `true` or when left empty, quote trimming will be enabled by default. Otherwise, the default is `false`. Regardless of the value of the property, the value can still be overriden through the usual API - therefore there is no break in backwards compatibility.


remkop:
Merged. Thanks for the PR!

Issue #530, opened 2018-10-26
Usage message customization (dependency injection and separation of concerns)


stechio:
Hi Remko,

first of all, kudos for the huge work put in this project, it feels very accurate and full of dedication!

Despite all the goodies, I've bumped against an annoying issue: its API seems a bit tightly coupled, there's less support to dependency injection than one might expect. Conversely, it seems to lack separation between presentation and data.

Let me explain: in my case, I needed to slightly customize the usage message layout (embellishment tweaks (expansions, indentations...) to make it more pleasing to the eye).
I found out that [the suggested route to expansions](https://picocli.info/#_section_headings) mixes line separation flags (`%n`) to actual contents (for example, `descriptionHeading="%nDescription:%n%n"`), whilst indenting specific contents seems impossible without resorting to hackish tricks (public `CommandLine.usage(..)` methods don't accept `Help` objects as argument, they all instantiate their own default implementation; unfortunately, the default `Help` class implements `description()` through a join that forces `table.indentWrappedLines=0` (sic!)).

The possibility to inject a derived `Help` object through `CommandLine.usage(..)` would have spared me the annoying hack to extract the default layout implementation (`CommandLine.usage(StringBuilder, Help)`) just to tweak its building.
Furthermore, IMO, layout formatting within content attributes (such as surrounding headings with newlines to adjust their vertical spacing, as described above) should be discouraged, since presentation (as you certainly know) should be separate from actual data, delegating it as much as possible to renderers: each and every content member (headings, section contents (header, synopsis, description, footer...)...) should be associated to its own renderer, without hard-coding string manipulation as seen in current `Help` implementation. About section reordering, I'd throw away the hard-coded members currently exposed by the `Help` class, in favor of a flexible map of members fluently manipulatable without having to hard code `StringBuilder.append(..)` call trains (as seen in `CommandLine.usage(StringBuilder, Help)`).

Please, let me know if I missed/misinterpreted anything.
thank you!

stechio:
In the meantime, **following your ideas I've just implemented the whole thing** -- it's *completely respectful of the existing API and idioms, yet flexible to customization*, really sweet!

Hard-coded rendering blocks have been refactored into their respective renderers, wiring back the existing entry points, so it's totally transparent to existing users.

All the existing junit tests passed (in the first run, I only had to fix the headingLength used in the detailedSynopsis algorithm), so I'm confident to send you a pull request quite soon.
cheers

PR #567, opened 2018-12-10
[#530] Usage message customization (2)

SysLord:
Issue #530 
Changes based on stechios changes and your suggestions.
Does this comply with your suggestions? Just say what you think.
First ever pull request, so I don't know what I am doing github-wise.
I assume that every new Help(...) inside of Help needs to be replaced by the factory?

For me all of this usage customization only has a value when Help.detailedSynopsis is split into multiple methods. This affects synopsis Help.synopsis(int). So I suggest making the switch between custom, abbreviated and detailed synopsis in Help.sections. Plus some code reuse to implement Help.synopsis(int) if the interface should not break. I think it fits this issue, or should i open a new issue for that?

My agenda: I like to see some parameters in usage instead of commandSpec.qualifiedName() and want to replace qualifiedName() with a subcommand-loop. e.g. we do 'thingy 119 update foo=13' and expect as subcommand help synopsis: thingy <id> update.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -8109,6 +8109,7 @@ public class CommandLine {
         private final ColorScheme colorScheme;
         private final Map<String, Help> commands = new LinkedHashMap<String, Help>();
         private List<String> aliases = Collections.emptyList();
+        private IHelpFactory helpFactory;
 
         private IParamLabelRenderer parameterLabelRenderer;
 
@@ -8144,7 +8145,8 @@ public class CommandLine {
             this.aliases.add(0, commandSpec.name());
             this.colorScheme = Assert.notNull(colorScheme, "colorScheme").applySystemProperties();
             parameterLabelRenderer = createDefaultParamLabelRenderer(); // uses help separator
-
+            this.helpFactory = commandSpec.commandLine() != null ? commandSpec.commandLine().getHelpFactory() : new DefaultHelpFactory();
+            
             this.addAllSubcommands(commandSpec.subcommands());
         }
 
@@ -8157,6 +8159,10 @@ public class CommandLine {
         /** Returns the {@code ColorScheme} model that this Help was constructed with.
          * @since 3.0 */
         public ColorScheme colorScheme() { return colorScheme; }
+        
+        /** Returns the {@code ColorScheme} model that this Help was constructed with.
+         * @since 2.9 */
+        private IHelpFactory getHelpFactory() { return helpFactory; }
 
         /** Option and positional parameter value label renderer used for the synopsis line(s) and the option list.
          * By default initialized to the result of {@link #createDefaultParamLabelRenderer()}, which takes a snapshot
@@ -8205,7 +8211,7 @@ public class CommandLine {
          * @return this Help instance (for method chaining) */
         Help addSubcommand(List<String> commandNames, CommandLine commandLine) {
             String all = commandNames.toString();
-            commands.put(all.substring(1, all.length() - 1), new Help(commandLine.commandSpec, colorScheme).withCommandNames(commandNames));
+            commands.put(all.substring(1, all.length() - 1), getHelpFactory().create(commandLine.commandSpec, colorScheme).withCommandNames(commandNames));
             return this;
         }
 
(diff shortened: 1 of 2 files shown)


SysLord:
Did not build. Shame on me.
What to do about the Java 1.5 compatibility?

src/main/java/picocli/CommandLine.java
```diff
@@ -8030,6 +8053,25 @@ public Help(CommandSpec commandSpec, ColorScheme colorScheme) {
...
+    	        () -> parameterList(),        			 //e.g. [FILE...] the files to convert
+    	        () -> optionListHeading(),    			 //e.g. %nOptions:%n%n
+    	        () -> optionList(),           			 //e.g. -h, --help   displays this help and exits
+    	        () -> commandListHeading(),   			 //e.g. %nCommands:%n%n
+    	        () -> commandList(),          			 //e.g.    add       adds the frup to the frooble
+    	        () -> footerHeading(),
+    	        () -> footer()
+    	    );
+        }
+
```
remkop: We want user to be able to plug in their own section renderers. My initial thought was that users would subclass `Help` and override this method, but now I'm thinking it may be nicer to do something like this:

```
private Map<String, ISectionRenderer> sectionMap = createDefaultSectionMap();

/**
 * Returns the map of section keys and renderers used to construct the usage help message.
 * The usage help message can be customized by adding, replacing and removing section renderers from this map.
 * Sections can be reordered with {@link #setSectionKeys(List) setSectionKeys}.
 * Sections that are either not in this map or not in the list returned by {@link #getSectionKeys() getSectionKeys} are omitted.
 * @since 3.9
 */
public Map<String, ISectionRenderer> getSectionMap() { return sectionMap; }

private Map<String, ISectionRenderer> createDefaultSectionMap() { 
    Map<String, ISectionRenderer> result = new HashMap<String, ISectionRenderer>();

    // TODO we should probably have constants for the default section keys
    result.put("headerHeading", new ISectionRenderer() { public String render(Help help) {
            return help.headerHeading(); } });
    result.put("header", new ISectionRenderer() { public String render(Help help) {
            return help.header(); } });
    result.put("synopsisHeading", new ISectionRenderer() { public String render(Help help) {
            return help.synopsisHeading(); } });
    result.put("synopsis", new ISectionRenderer() { public String render(Help help) {
            return help.synopsis(help.synopsisHeadingLength()); } });
    // ...
    return result;
}

```

src/main/java/picocli/CommandLine.java
```diff
@@ -7925,6 +7943,11 @@ public void init(CommandLine helpCommandLine, Help.Ansi ansi, PrintStream out, P
         void init(CommandLine helpCommandLine, Help.Ansi ansi, PrintStream out, PrintStream err);
     }
 
+    
+    public interface IHelpSectionRenderer {
+    	String render();
```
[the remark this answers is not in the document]
remkop: After thinking some further the 2nd signature looks better. What do you think?

```
// TODO make Help.commandSpec() method public if we choose this signature.
String render(Help help);
```

remkop:
> For me all of this usage customization only has a value when Help.detailedSynopsis is split into multiple methods. This affects synopsis Help.synopsis(int). So I suggest making the switch between custom, abbreviated and detailed synopsis in Help.sections. Plus some code reuse to implement Help.synopsis(int) if the interface should not break. I think it fits this issue, or should i open a new issue for that?

Yes, I think it may make sense to open a separate ticket for that. Essentially you want the detailedSynopsis method separated into the logic that generates the command name and the logic that generates the option and positional parameter details, correct?

The work for this ticket will certainly enable that, I don't think we're very far away. Will you be able to add some unit tests?

remkop:
Do you think you will be able to work on this pull request further? I can do a separate 3.9 release for this if we can polish this a bit. Otherwise it will have to wait until 4.0 or later.

src/main/java/picocli/CommandLine.java
```diff
@@ -8010,7 +8145,8 @@ public Help(CommandSpec commandSpec, ColorScheme colorScheme) {
             this.aliases.add(0, commandSpec.name());
             this.colorScheme = Assert.notNull(colorScheme, "colorScheme").applySystemProperties();
             parameterLabelRenderer = createDefaultParamLabelRenderer(); // uses help separator
-
+            this.helpFactory = commandSpec.commandLine() != null ? commandSpec.commandLine().getHelpFactory() : new DefaultHelpFactory();
```
SysLord: Not sure if this is the right way to get the helpfactory.

remkop: Yes, that looks fine. You raise a good point though: the `CommandLine` class is mostly a facade, and it may be good to move some of this down to the `CommandSpec` class. I will go ahead with the merge first, and I may make some changes afterwards.

remkop:
Merged. Thank you for the pull request! I may make some further changes.
You mentioned you wanted to make some changes to the API to enable customizing the `detailedSynopsis`. Would you mind creating an issue for that?

SysLord:
First, thank you for your excellent support.
>> Would you mind creating an issue for that?
Yes, maybe today or tomorrow.

PR #587, opened 2019-01-03
JLine3 example: use defaultParser to split readLine output

mattirn:
Use defaultParser to split readLine output in order to manage correctly escaped chars and quoted arguments


remkop:
Nice one! 
This has been merged and will be included in the upcoming 3.9 release.

PR #602, opened 2019-01-14
Make CommandLine in AutoComplete use correct IFactory implementation

### LINKED ISSUE #601: AutoComplete fails if using a custom IFactory implementation even if provided with --factory
Our structure is as follows:

```
Class (command)

- Main.class (main)
  -- Sub.class (main sub)
    --- Sub$Command1.class (main sub com1)
    --- Sub$Command2.class (main sub com2)
```

All subcommands are declared in the `@Command` annotation.

When running the AutoComplete command to generate the file we get a stack trace with error:
```
Cannot instantiate subcommand package.Sub$Command1: the class has no constructor
Caused by: java.lang.NoSuchMethodException:package.Sub$Command1.<init>()
```

- This class has one constructor with parameters annotated with `@Inject`. 
- We use our own implementation of `IFactory` we've passed to the command with `--factory`.
- The `Main.class` is the command we're trying to generate the autocompletion script for. 

It seems like the problem is that the `CommandLine` implementation that is created by the `AutoComplete` command does not use the factory passed to it.

See: https://github.com/remkop/picocli/blob/82f7c75b54d11ab3047bffc5f1a428ff68e5ea0c/src/main/java/picocli/AutoComplete.java#L134-L140

I've built picocli locally changing this line:
```diff
-CommandLine commandLine = new CommandLine(instance);
+CommandLine commandLine = new CommandLine(instance, factory);
```

This resolved the issue for us and I will create a pull request for it soon. Let me know if I'm missing any side effects this change may produce that require testing.

MikolajK:
In some cases using the DefaultFactory when instantiating CommandLine could cause subcommands that require a specific implementation to fail to be created. This commit makes the CommandLine instance use the IFactory implementation passed by the user to the AutoComplete command (if any) which should resolve this issue.

Closes #601

### DIFF
--- a/src/main/java/picocli/AutoComplete.java
+++ b/src/main/java/picocli/AutoComplete.java
@@ -137,7 +137,7 @@ public class AutoComplete {
                 }
                 Class<?> cls = Class.forName(commandLineFQCN);
                 Object instance = factory.create(cls);
-                CommandLine commandLine = new CommandLine(instance);
+                CommandLine commandLine = new CommandLine(instance, factory);
 
                 if (commandName == null) {
                     commandName = commandLine.getCommandName(); //new CommandLine.Help(commandLine.commandDescriptor).commandName;


remkop:
Merged into master. 
Thank you for the pull request!

remkop:
Updated the release notes.

PR #606, opened 2019-01-15
Subcommand demo example

deining:
Citation from #223:

> If anyone has subcommand examples they would not mind contributing, all pull requests are welcome.

Hopefully the example I authored can be considered here. I guess it can be helpful for potential new users of the excellent piciocli library.


remkop:
Updated the release notes.

Issue #620, opened 2019-02-02
Feature suggestion: allow non-static inner classes to be subcommands of their outer class


ghost:
For example, something like (not tested)

```
@Command(name="outer", subcommands={Inner.class})
public class Outer {

    @Command(name="inner")
    public class Inner {}
}
```

Currently this doesn't work, since `Inner` requires an instance of `Outer`. Would it be possible to make it so that when an instance of `Outer` is run via the `CommandLine` it will create an instance of `Inner` and use it as a subcommand? This would probably only really work if `Inner` is a subcommand of `Outer`, as opposed to a subcommand of a completely separate class. 

I know `Inner` could be made static, but then it loses the ability to use private variables of `Outer`, which is useful. Another possible way to do something like this (which I have not tested) might be to use a method as the command and have it run a new `Inner` object with all args it recieves. But even if that would work (and I'm not sure), it would be simpler to be able to be able to just put it in the `subcommands` section.

remkop:
I think you can already accomplish this by using a factory that has a reference to the outer class.
Picocli has some tests that do this.
Please take a look at https://github.com/remkop/picocli/blob/master/src/test/java/picocli/InnerClassFactory.java

PorygonZRocks:
That looks like it's what I need, thanks!

I'm thinking that with a loop checking `getEnclosingClass` it might be possible to make picocli automatically detect if a subcommand is an inner class of the main command, and automatically do something like that. If I did that and made a PR (no promises, but hopefully), would that be fine, or would you rather not have something like that in the default code?

remkop:
I think the enclosing class may not be sufficient: we need a reference to an *instance* of the enclosing class. Or am I missing something?

remkop:
So I don’t think this can be automatic but if I’m wrong please let me know. I always consider pull requests.

PorygonZRocks:
Assuming there's a way to get the actual object from a `Command` (which I haven't checked), you should be able to run `getEnclosingClass` on the subcommand's object, and check it against the class of the command object. If they're the same, use the command object to create the subcommand. Sort of example of what I mean:

```
    Object cmdObject = cmd.getObject();
    Object subClass = sub.getClass();
    Object subObject = null;
    if (cmdObject.class == subClass.getEnclosingClass()) {
        subObject = subClass.getInstance(cmdObject);
    }
```
 Obviously the real code would probably look pretty different, but do you get what I mean?

PR #621, opened 2019-02-02
Change `DefaultFactory` to be based on `InnerClassFactory`, use it

PorygonZRocks:
I basically just copied `create` from [`InnerClassFactory`](https://github.com/remkop/picocli/blob/master/src/test/java/picocli/InnerClassFactory.java), added a couple methods for backwards compatibility (they just use `null` for the object, which I believe `InnerClassFactory` catches and works with). I also changed almost all uses in picocli to use the new constructor (although there were a couple uses in static methods I couldn't change). 

**NOTE: I AM NOT ABLE TO TEST**. I don't remember if there's auto tests on PRs here or not, but if not, someone else will need to test, as I unfortunately can't. 

Another note: this does NOT affect custom factories like I had originally hoped to be able to do, only the default one.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -160,7 +160,7 @@ public class CommandLine {
      * @throws InitializationException if the specified command object does not have a {@link Command}, {@link Option} or {@link Parameters} annotation
      */
     public CommandLine(Object command) {
-        this(command, new DefaultFactory());
+        this(command, new DefaultFactory(command));
     }
     /**
      * Constructs a new {@code CommandLine} interpreter with the specified object (which may be an annotated user object or a {@link CommandSpec CommandSpec}) and object factory.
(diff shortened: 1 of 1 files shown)


PorygonZRocks:
See #620 for an example of a use for this

remkop:
Thanks for the PR. I am away from my PC but looking at it on my phone I have a better idea of what you have in mind now. 

I’m not sure yet if this is a good idea or not. 

I need to spend some more time on this to understand better whether the new factory’s behavior is consistent and easy to understand or whether there are edge cases where its behavior will be surprising for casual users.
(E.g., what if the command hierarchy has several levels of non-static inner class commands? )

The other thing is, is this a common use case? Why would you want to define commands using non-static inner classes? (What problem are we trying to solve?)

Note that picocli offers a `@ParentCommand` annotation and a `@Spec` annotation that allows commands access to other command instances in the hierarchy, even if they are static inner classes. Would this be sufficient or are there cases where non-static inner classes are preferable to static inner classes with these annotations?

That said, the changes look relatively straightforward at first glance. 

These are some initial thoughts. I need to think about this some more.

PR #643, opened 2019-03-13
Change % to %% when using ${DEFAULT-VALUE} in option description

StefRe:
Percent signs shall be escaped in the description text as it is going
to be passed to String.format. If the default value contains (single) %
signs then using the built-in DEFAULT-VALUE variable would lead to a
warning message. This is avoided by escaping any % signs in the
DEFAULT-VALUE variable when used in the description.

This is related to #615 and its fix. I have an option with a default value of "%s - page %d of %d" and using ${DEFAULT-VALUE} in the description leads to the warning message. A workaround would have been to not use ${DEFAULT-VALUE} but instead writing the default value with doulbe % signs in the description but I definitely prefer the use of the variable just in case the default changes some time later.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -5329,7 +5329,7 @@ public class CommandLine {
                 String defaultValueString = defaultValueString();
                 String[] result = new String[desc.length];
                 for (int i = 0; i < desc.length; i++) {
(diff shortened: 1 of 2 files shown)


remkop:
Thank you for the PR!
This has been merged and will be included in the next release.

Issue #536, opened 2018-10-30
Optional interactive


Zethson:
Hi,

I'm having the following issue: 
I have a password CommandLine.Option, which is currently interactive -> thus it asks the user for his password. 
However, I would like to skip the interactive question, if a password was already provided.

Is it possible to consume the interactive request?

remkop:
@Zethson  At the moment picocli does not support this. There is a [section](https://picocli.info/#_interactive_password_options) in the manual about combining interactive use with non-interactive use.

Basically, if you need this, one idea is to have separate options, like `--password:hidden` for the interactive option, and `--password:clear` that accepts the password in cleartext as an option parameter on the command line. The latter is also suitable for batch scripts.

Zethson:
I see, thanks!
I still would like to kindly request this feature. 

It's more or less the only thing that I miss from [Click](https://click.palletsprojects.com/en/7.x/) .

remkop:
Understood. Not sure when I'll get to it.
I'm a bit worried about introducing ambiguity, for example:

```java
class App {
    @Option(names = "--password", interactive = true)
    String password;

    @Parameters
    String[] otherParams;
}
```

Now, if the end user specifies some input:

```
<command> --password aaa bbb ccc
```
Currently picocli will interactively ask for a password and put `aaa`, `bbb` and `ccc` in the `otherParameters` array.

This will change after this feature request is implemented: for the same input, the `aaa` parameter will now be interpreted as a password parameter value. Not sure yet how to deal with this...

Zethson:
What Click does is that every parameter that is NOT passed and labeled as interactive is prompted. Everything that was passed is not going to be prompted.
<code>  --password aaa bbb ccc</code> 
is going to set aaa as value for password   
 
<code>  <command> aaa bbb ccc</code>
will open the interactive prompt and ask for the password, since no value for password was passed.

remkop:
I guess I will either have to live with potentially breaking existing applications, or introduce some extra flag that needs to be set for picocli to accept values following `--password` as the password value.

Zethson:
I'm afraid yes.

I prefer the potentially breaking change, but this is of course up to you.

Thanks!

remkop:
JCommander actually implemented this feature request:

* https://github.com/cbeust/jcommander/issues/70
* https://github.com/cbeust/jcommander/issues/195
* https://github.com/cbeust/jcommander/issues/333

[...]

Issue #358, opened 2018-04-13
Support repeating composite arguments (was: Parameter options/modifiers)


kravemir:
Picocli supports mixing options and parameters: http://picocli.info/#_mixing_options_and_positional_parameters

I would like to know what options have been specified before certain parameter. Consider such command:

```
print --paper A4 A.pdf --count 3 B.pdf --rotate left C.pdf -- D.pdf --rotate right E.pdf
```

Where:

* `--paper` is an global option,
* `--count` and `--rotate` are "local" options, or parameter modifiers,
*  `--` reset local options.

So, that command would evaluate to:

* globals: `paper = A4`,
* parameter `A.pdf`: `count = default, rotate = default`,
* parameter `B.pdf`: `count = 3, rotate = default`,
* parameter `C.pdf`: `count = 3, rotate = left`,
* parameter `D.pdf`: `count = default, rotate = default`,
* parameter `E.pdf`: `count = default, rotate = right`.

Is there support for such things in picocli?

remkop:
Brainstorming a bit: one idea is to introduce a new `@Composite` annotation, that would allow you to define a group of options and positional parameters that need to be recognized together. Something like the below...

```java
@Command(name = "print)
class PrintCommand {
  enum PaperFormat { A1, A2, A3, A4 }

  @Option(names = "--paper")
  PaperFormat format;

  @CompositeDelegate
  Document[] documents;
}

@Composite
class Document {
  @Parameters(index = "0", arity = "1")
  File file;

  @Option(names = {"-c", "--count"})
  int count = 1; // the default

  enum Rotate { left, right, straight }

  @Option(names = "--rotate")
  Rotate rotate = Rotate.straight; // the default
}
```

kravemir:
Composite positional parameters is a good idea. I would rather keep naming consistent, and named it `@CompositeParameters`.

### Modifiers idea

I was also thinking about modifiers vs options. Where modifier would apply to all further parameters, and is reset by `--`, but option is passed only to one parameter following it. Ie. set `count` only for one document, or for all following documents.

Depends on use case, but it might be good to leave a choice to an application user, instead of application developer or library developer (forced style).

One idea is to introduce `modifier_names` to `@Option`:

```java
  @Option(names = {"-c", "--count"}, modifier_names = {"---count"})
  int count = 1; // the default
```

So, command-line application user can use `-c` and `--count` to specify count for one following document, but can also use `---count` to specify count for all following documents until reset.

### Modifier reset functionality

[...]

PR #648, opened 2019-03-23
Introduce a failing test for char[] type conversion

triceo:
When I have a `char[]`-typed `@Option` in my command-line object, I get an error when trying to set it from the command line:

    Invalid value for option '-charArray' (<charArrayField>): 'abcd' is not a single character

It appears that Picocli treats `char[]` as `char`. The expected behavior instead would be that the input `String` is converted to `char[]` using its `toCharArray()` method.

Couple more notes:

- Using `char[]` as the data type is a pretty common thing when the parameter is a password. The array can be manually zeroed when the password is no longer needed, thus completely removing it from memory while its parent object still lives.
- Using `String` for the same purpose is sub-optimal, as JVM does all kinds of magic with `Strings` and the password could, for example, be interned - irrevocably burning it into the JVM memory for the entire runtime of the application.
- The original code is lifted directly from a JCommander-based implementation, which handled this situation properly.
- I attempted to fix the problem, but the type conversion code seems very convoluted. Without gaining a deep understanding of all the code branches, I risk causing more problems than I solve.

### DIFF
--- a/src/test/java/picocli/CommandLineTypeConversionTest.java
+++ b/src/test/java/picocli/CommandLineTypeConversionTest.java
@@ -64,6 +64,7 @@ public class CommandLineTypeConversionTest {
         @Option(names = "-Byte")          Byte aByteField;
         @Option(names = "-char")          char charField;
         @Option(names = "-Character")     Character aCharacterField;
+        @Option(names = "-charArray")     char[] charArrayField;
         @Option(names = "-short")         short shortField;
         @Option(names = "-Short")         Short aShortField;
         @Option(names = "-int")           int intField;
@@ -493,6 +494,16 @@ public class CommandLineTypeConversionTest {
             fail("Invalid format was accepted");
         } catch (CommandLine.ParameterException expected) {
             assertEquals("Invalid value for option '-char': 'aa' is not a single character", expected.getMessage());
+        }
+    }
+    @Test
+    public void testCharArrayConverter() {
+        try {
+            final SupportedTypes cli = new SupportedTypes();
+            CommandLine.populateCommand(new SupportedTypes(), "-charArray", "abcd");
+            assertEquals(new char[]{'a', 'b', 'c', 'd'}, cli.charArrayField);
+        } catch (Exception exception) {
+            fail("Unexpected exception while converting char[] type: " + exception.getMessage());
         }
     }
     @Test


remkop:
Hi,

What you're saying makes total sense. It has been on the todo list to improve this but other things took priority.

As you pointed out, the type conversion logic in picocli is not easy, because it needs to support adding values to multi-value types like Map, Collection and arrays, as well as simply replacing values of single-value fields. However, I have some idea for adding support for password `char[]` values without changing the type conversion logic too much.

I am considering combining a fix for this with #536 to improve password handling overall and cutting a separate 3.9.x maintenance release for this.

How urgent is this for you? I am currently working on #358, which is quite involved and will probably take some more time. I'd like to finish that and do a picocli-4.0-alpha-1 release before context-switching to the topic of passwords.

triceo:
@remkop I can wait. Combining this with #536 makes a lot of sense.

remkop:
@triceo Would you be okay with getting this in 4.0.0-alpha-2, or do you need a backport to the 3.9.x branch? I would prefer to spend as much time as possible on 4.0...

triceo:
@remkop That depends on how many incompatible changes there are in 4.0, and when that is coming out. Right now, I don't know anything about the release.

remkop:
So far it’s all backwards compatible. 

The final GA release is a few months away, but the alpha releases are high quality and pass all 1500+ unit tests. 

I’m releasing this functionality as alpha releases so that I have the option to change the API if peer review turns up any issues.

triceo:
@remkop Unfortunately, the release rules won't let me depend on a non-final artifact. That said, I understand your motivation for not putting this into a 3.9 - I'll wait for a final 4.0 then.

remkop:
> the release rules won't let me depend on a non-final artifact. 

Understood. Thanks for the clarification. I’ll try to do a separate 3.9.x release for these then.

remkop:
@triceo I’ve pushed some commits to the 3.X branch to improve support for interactive (password) options. I’ll create a 3.9.6 release for this soon, maybe later today. 

This includes the `char[]` type support you requested, but I’ve limited my changes so far to only impact _interactive_ options. 

So, for “normal” (non-interactive) options, using a field of type `char[]` still works similar to using a field of type `int[]`: picocli expects each option parameter to be a single value that can be converted to a `char` or `int` and added to the array. Your PR adds a test for “normal” (non-interactive) options, and this test will still fail, so I cannot merge this PR yet. 

It does make sense to go one step further and treat `char[]` more like Strings than like `int[]` also for “normal” (non-interactive) options. (I’ll probably hold off on this until 4.0.) Making this change would potentially break existing applications so we need to provide a setting that allows users to switch back to the old behavior. Any ideas for a good name for such a configuration?

triceo:
Hmm... `picocli.disableCharArrayToStringConversion`? 
A bit long, but descriptive.

remkop:
Adding some ideas:
* `charArraysCanCaptureStrings` 
* `charArraysCaptureSingleCharacters`

remkop:
I just took another look at this, but I'm not sure that I want to go ahead with this:

The main goal (support for char[] types for interactive password options) has been achieved in 3.9.6.

This ticket is still open because we are considering to go one step further and also allow normal (non-interactive) options to capture multi-char (String) values in a char[] type. On the face of it this would make things more consistent, but I hesitate to go ahead:

* I'm not sure if there is any user demand for this (nobody asked for this)
* it would introduce more special logic for `char[]` in a few places, like `CommandLine$Interpreter.getTypeConverter` and `CommandLine$Interpreter.applyValuesToArrayField` (this is a minor issue, just mentioning it)
* I am unsure how to deal with multiple values: append to the array, or overwrite the previous value? For other array types, the expected behaviour is to append, while for single value types the expected behaviour is to overwrite. Would we need to introduce another parser switch for this? Seems overkill since nobody asked for this...

So I will probably postpone this remaining work until someone asks for this feature.

Thoughts?

triceo:
@remkop It is true that, since 3.9.6, my use case is solved and I don't need anything more. If doing this feature in its entirety is too much for the existing system, feel free to postpone/ignore it. :-)

remkop (approved):

remkop:
Finally merged! 😅 

PR #652, opened 2019-04-01
Update index.adoc

deining:
Fixing minor glitches


remkop:
Merged to master. Thanks for the contribution!

Issue #495, opened 2018-09-26
Add module picocli-jpms (was: Include module-info.class for full JPMS modularization and jlink support)


remkop:
Currently the picocli jar is an automatic module (manifest has `Automatic-Module-Name: info.picocli`). 

> A JLink trimmed binary image needs explicit modules. This means the jar needs to include a `module-info.class`.

Update: this is a common misunderstanding. See the comment below for a tutorial for creating a JLink binary image with an automatic module. 

**module-info.class trade-offs**

It is [possible](https://stackoverflow.com/questions/47359025/is-it-possible-to-mix-java-8-and-java-9-source-code-in-the-same-project-without) to compile only the module-info.java file with Java 9 and include it in the picocli jar where all other classes are compiled with Java 5.

However, there are quite a few tools that scan all classes in a Jar and that would choke on the module-info.class (whether it is in the root of the jar or in `META-INF/versions/9/`). For example: [jandex](https://issues.jboss.org/browse/JANDEX-40), and [Apache Commons BCEL](https://bz.apache.org/bugzilla/show_bug.cgi?id=60688).

Consider producing a `picocli-jpms-3.x.x.jar` that has the same contents as `picocli-3.x.x.jar` but in addition a `module-info.class`. Users looking to create a JLink custom runtime image can use the picocli-jpms jar.

**Solution**

The comment below is a brief tutorial for creating a JLink binary image with an automatic module. No `module-info.class` required in picocli or in the application. 

Todo: do a more formal write-up and add it to the picocli documentation.

remkop:
It turns out that an application does not need to be a full-fledged JPMS module with a `module-info.java` file to generate a custom `jlink` runtime image for your application.

Example application:

```java
package hello;

import picocli.CommandLine;
import picocli.CommandLine.Command;
import picocli.CommandLine.Option;

@Command(name = "hello", description = "picocli demo", version = "3.6.1", mixinStandardHelpOptions = true)
public class HelloWorld implements Runnable{

    @Option(names = {"-u", "--user"}, description = "Specify a user. Default is ${DEFAULT-VALUE}.")
    String user = "unknown";

    @Override
    public void run() {
        System.out.printf("Hello, %s", user);
    }

    public static void main(String[] args) {
        CommandLine.run(new HelloWorld(), args);
    }
}
```

Example `build.gradle`:

```build
group 'experiments'
version '1.0-SNAPSHOT'

sourceCompatibility = 1.8

repositories {
    jcenter()
}

[...]

PR #662, opened 2019-04-12
Jmps java 9

Warkdev:
Adding JPMS module définition to picocli.
Adding new subproject picocli-core-module

### DIFF
--- a/picocli-core-module/build.gradle
+++ b/picocli-core-module/build.gradle
@@ -4,14 +4,15 @@ plugins {
     id 'maven-publish'
     id 'com.jfrog.bintray'
     id 'java-library'
+    id 'org.beryx.jar' version '1.1.3'
 }
 
 group 'info.picocli'
 description 'Picocli JPMS module.'
 version "$projectVersion"
 ext.moduleName = 'info.picocli'
-sourceCompatibility = JavaVersion.VERSION_1_9
-targetCompatibility = JavaVersion.VERSION_1_9
+sourceCompatibility = JavaVersion.VERSION_1_8
+targetCompatibility = JavaVersion.VERSION_1_8
 
 task copyRootProjectCode(type: Copy) {    
     from("${rootProject.projectDir}/src/main/java"){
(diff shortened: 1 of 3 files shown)


picocli-core-module/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+    }
+    into "${projectDir}/src/main/java"
+}
+
+compileJava {
+    inputs.property("moduleName", moduleName)   
+    doFirst {
+        options.compilerArgs = [
+            '--module-path', classpath.asPath,
+        ]
```
remkop: Looks like this fails the build on Java 8... I see this:

```
> Task :picocli-core-module:compileJava FAILED
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':picocli-core-module:compileJava'.
> invalid flag: --module-path
```

Maybe we can work around that with something like this:

```
doFirst {
    if (org.gradle.api.JavaVersion.current().isJava9Compatible()) {
        options.compilerArgs = [
            '--module-path', classpath.asPath,
        ]
    }
    classpath = files()
}

remkop:
I went ahead and merged the PR. 
Thank you for the hard work!
With the version check the build now passes on Java 8, so it's looking good!

remkop:
Link to the original issue:
https://github.com/remkop/picocli/issues/495

picocli-core-module/build.gradle
jwheeler91: Java 8 and below won't understand the --module-path argument so that makes sense to me...

jwheeler91: If this core module is aimed squarely at the JPMS features, why do we need to support java 8 anyway as it the main jar will support java 8?

remkop: The plugin that’s used allows us to build a Java 9 module with just Java 8. It uses a byte code manipulation library for this. That makes it easier for contributors to set up the environment and build the project so I like that.

jwheeler91: Cool. I see. Nice work

Warkdev: @jwheeler, I'm not in this weekend, feel free to update the gradle file as remkop suggested. 😊

remkop:
I went ahead and merged the PR, and already made the above changes. I'll put further comments on #495.

PR #671, opened 2019-04-17
Fix shellcheck warning in autocompletion script

AlcaYezz:
Update autocompletion script generation:
- Change variable scope to local
- Change prev_word variable to be set only if needed
- Change curly brace struct to if / for struct
- Change COMPREPLY populating method


remkop:
Great, thank you for the contribution!

FYI: I am planning to release 4.0-alpha-2 tonight to coincide with Nicolas Peters' [DevoxxFR lightning talk about picocli and GraalVM](https://cfp.devoxx.fr/2019/talk/XJZ-0197/Des_applications_en_ligne_de_commande_avec_Picocli_et_GraalVM__).

If I am unable to review your PR in time for alpha-2, it will go into alpha-3.

remkop:
@AlcaYezz Thanks for your patience, I finally had some time to review the proposed changes.

----
Source changes:

* I like the `for` loop improvements; the `for ... ; do ... done` is better.
* I also like introducing `if; then ... ; fi` to replace the `{ ... && ... }` constructs.
* Making variables `local` and lower-case: looks good.
* Avoid declaring the `PREV_WORD` variable is it is not used: looks good.

I am not sure about the change from 

```
COMPREPLY=( $(compgen -W \"${COMMANDS}\" -- ${CURR_WORD}) )
```
to
```
read -d -a COMPREPLY < <(compgen -W \"${commands}\" -- \"${curr_word}\")
```

What is the problem we are trying to fix here? (Also, what does the `-d` option without a `delim` parameter do?) Looking at other completion scripts on the web I get the impression that the original syntax is commonly used.

----
Testing the result:

To do a quick manual test, I sourced the `picocompletion-demo_completion.bash` file that is generated by `AutoCompleteTest#nestedSubcommands()`, but it seems that the completion is no longer working as before:

```
INPUT                                 EXPECTED OUTPUT
picocompletion-demo <TAB>          -> picocompletion-demo sub
picocompletion-demo sub<TAB><TAB>  -> sub1 sub2  #(on next line)
picocompletion-demo sub2 <TAB>     -> picocompletion-demo sub2 subsub
picocompletion-demo sub2 -<TAB>    -> -d  --directory  --num2  #(on next line)
```

With the PR, what actually happens is that `picocompletion-demo <TAB>` produces 

```
picocompletion-demo sub1
sub2
```

(So both sub1 and sub2 are completed, with sub2 on the next line.)
If I hit `<TAB>` again after that, the subcommands for sub2 are both completed, again one on the next line:

```
picocompletion-demo sub1
sub2 subsub1
subsub2
```

Also, I noticed that options are not completed anymore.
For example, `picocompletion-demo -<TAB>` should produce
`-h         --help     -V         --version`
on the next line, but no options are shown.

remkop:
@AlcaYezz , have you had a chance to look at my comments?

AlcaYezz:
Yes sorry for the delay, I just take a look.

What I do with the COMPREPLY change is fixing this warning:

    COMPREPLY=( $(compgen -W "${FLAG_OPTS} ${ARG_OPTS}" -- ${CURR_WORD}) )
                ^-- SC2207: Prefer mapfile or read -a to split command output (or quote to avoid splitting).

Anyway you are right a delim parameter is missing and everything work with it.

---

I can either fix the read command or revert to the first one. What did you think?

remkop:
Thanks for providing that error number. I took a look at https://github.com/koalaman/shellcheck/wiki/SC2207 and now I understand better. Let's use the `read` command.

AlcaYezz:
There is one last warning that I miss on the first try about the read command.

      read -d ' ' -a COMPREPLY < <(compgen -W "${_AB_C_option_args}" -- "${curr_word}")
      ^--[SC2162](https://github.com/koalaman/shellcheck/wiki/SC2162): read without -r will mangle backslashes.

We shouldn't find backslash in command name / option / attribute so we can add this option without problem, don't we ?

remkop:
What happens if we use `read -r`? Would that cause any problem?

AlcaYezz:
I have done some more search and it should be added `read -r` avoid more processing on the input than the required splitting.

remkop:
Ok, so the plan is to use `read -d ' ' -ra` ?
Sounds good to me.

remkop:
Hi @AlcaYezz , will you be able to take care of this, or do you want me to take a look?

PR #770, opened 2019-07-20
Polish

snicoll:
Rather that defining the 3 annotations to enable configuration, auto-configuration and component scan, we use `@SpringBootApplication`.

We don't recommend using field injections and we prefer if the `SpringApplication` is as simple as possible as it is the root of the context. This PR moves the logic of executing the command in a separate class.

### DIFF
--- a/picocli-spring-boot-starter/README.md
+++ b/picocli-spring-boot-starter/README.md
@@ -37,28 +37,12 @@ This will bring in the `info.picocli:picocli` and `org.springframework.boot:spri
 ## Example Application
 
 ```java
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.CommandLineRunner;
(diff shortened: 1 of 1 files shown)


remkop:
@snicoll Thanks for improving the example!
Question: How will the `SpringApplication.run` method know to invoke the `MyApplicationRunner.run` method, since the `MyApplication` class has no reference to `MyApplicationRunner`? Does it scan the classpath for classes implementing `CommandLineRunner`?

snicoll:
Ah, sorry. No Spring doesn't scan the classpath unless told to. Adding `@SpringBootApplication` means that classpath occurs on the package of the app (and all sub-packages) but you need a stereotype to indicate it should be handled as a bean.

I forgot that and submitted #771 to fix it. Thanks for the review

remkop:
I see. Thank you for the clarification!

PR #796, opened 2019-08-26
Imperative mood for HelpCommand header

gitfineon:
-Align with other headers and descriptions
-Adjust some comments


remkop:
Merged. Thank you!

PR #811, opened 2019-09-10
Propagate resource bundle to subcommands recursively

frontfact:


remkop:
Thank you for the pull request! Good catch!
Would it be possible to add a unit test that demonstrates the problem?

frontfact:
> Would it be possible to add a unit test that demonstrates the problem?

I need to fix my Netbeans' JUnit installation first, because for some reason I can't run unit tests for now.

By the way, how does this work: does a new commit in my branch is reflected in this pull request automatically ?

thank you

remkop:
Yes, if you push more commits to the same branch they will show up in this pull request.
Good luck with NetBeans! (Programming is 10% inspiration and 90% getting your environment set up...)

remkop:
I see more commits were added. Thank you!
Please let me know when you’re finished.

frontfact:
I'm finished, let me know if it's not enough or if I missed something, thanks

remkop:
Merged into master. Thank you for the contribution!

PR #824, opened 2019-09-20
Add command() to ArgSpec

hmakholm:
There seems to be no public way to get from an `ArgSpec` to the command model it belongs to. Expose the existing field with a getter method.

My use case for this is so I can write a single generic `IDefaultValueProvider` that is sensitive to which subcommand it is getting a default for. This makes it simpler to implement [Bazel-style config files](https://docs.bazel.build/guide.html#bazelrc-syntax-and-semantics) that provide a uniform way to override defaults for all command-line options.

(If I'm overlooking something obvious, please advise).

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -7366,6 +7366,10 @@ public class CommandLine {
              * @since 4.0 */
             public ArgGroupSpec group() { return group; }
 
+            /** Returns the command this option or positional parameter belongs to.
+             * @since 4.1 */
+            public CommandSpec command() { return commandSpec; }
+
             /** Returns the untyped command line arguments matched by this option or positional parameter spec.
              * @return the matched arguments after {@linkplain #splitRegex() splitting}, but before type conversion.
              *      For map properties, {@code "key=value"} values are split into the key and the value part. */
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -7367,6 +7367,10 @@ public class CommandLine {
             public ArgGroupSpec group() { return group; }
 
             /** Returns the command this option or positional parameter belongs to.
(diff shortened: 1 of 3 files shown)


remkop:
I can see how this would be useful.

The reason I did not expose this earlier was that I was worried about an unusual use case: In theory, it is possible to build an `OptionSpec` or a `PositionalParamSpec` and add it to _multiple_ different `CommandSpec` instances. This would be unusual, but it is possible. In that case, the `command()` method will return the _last_ command that it was added to (and there is no way to get a list of the commands it was added to previously).

However, I think it is more important that normal use cases like the one you mentioned are possible and deal with unusual use cases by simply documenting how it works. :smile: 

Long story short: I like your change and I plan to merge it for picocli 4.1. :+1: 

Can you do me a favor and update the javadoc to mention that, in case an `ArgSpec` was added to multiple commands, the method returns the last command that it was added to. Also, please mention that the method may return `null` if the `ArgSpec` has yet been added to any command, or if the `ArgSpec` has been `remove()`-ed from its command.

Finally, can I ask you to add tests for the statements made in the javadoc? (Maybe add to `ModelOptionSpecTest` and `ModelPositionalParamSpecTest`.)

hmakholm:
Sounds reasonable. Should a similar warning apply in the javadoc for `ArgSpec.group()`?

remkop:
You are right that `ArgSpec.group()` is similar. Let me think a bit about the javadoc: perhaps it is better not to mention such strange use cases. I am not sure that I want to add validation to prevent it outright, but I certainly don't want to encourage it either. Mentioning it in the javadoc makes it sound a bit like it is a legitimate thing to do...

remkop:
Thank you for adding the test! :-) 👍

remkop:
Merged. Thanks again for the contribution!
(I will update the release notes when I get to my PC.)

PR #866, opened 2019-11-14
Fixing documentation: adding missing chapter 9.8 'Handling errors'

deining:
I realised that chapter 9.8 'Handling errors' was missing in the documentation, for whatever reason, and I don't think this is on purpose.
I bumped to latest versions of asciidoctor gradle plugin and asciidoctorj and rebuild the asciidocs, now the chapter is there (not sure what the reason for the missing chapter really was, though).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -237,7 +237,7 @@ javadoc {
 javadoc.dependsOn(jar)
 javadoc.dependsOn('asciidoctor')
 asciidoctorj {
-    version = '1.5.5'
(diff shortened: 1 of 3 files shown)


remkop:
Hi!
The "Handling errors" section has not been published yet intentionally because it (specifically `ShortErrorMessageHandler`) uses API that is not available yet in picocli-4.0.4. 
I was planning to publish this with picocli-4.1.
Apologies for the confusion!

PR #874, opened 2019-11-23
Javadoc issues

deining:
Hi,

while reading the API docs of the newly introduced PropertiesDefaultProvider, I realized that parts of the javadoc were missing. The attached patch fixes this issue.
A second patch fixes all javadoc errors that were present in the project.

Thanks for considering my input.


remkop:
Merged into master. 
Thank you!

PR #883, opened 2019-11-26
Quick guide: replace deprecated method calls

deining:


remkop:
Merged into master. 
I’ll update the release notes later. 
Thank you!

Issue #885, opened 2019-11-28
Documentation for programmatic API uses deprecated AbstractParseResultHandler


rbresalier:
The documentation for the programmatic API, https://picocli.info/picocli-3.0-programmatic-api.html, shows coding using AbstractParseResultHandler. However the apidocs state that AbstractParseResultHandler is obsolete: https://picocli.info/apidocs/picocli/CommandLine.AbstractParseResultHandler.html

So should the programmatic API docs be updated for the non-deprecated way (or remove AbstractParseResultHandler)? Or is AbstractParseResultHandler really deprecated and that should be removed from the apidocs?

remkop:
That is a good point, thank you!

`AbstractParseResultHandler` has been replaced by [IExecutionStrategy](https://picocli.info/apidocs/picocli/CommandLine.IExecutionStrategy.html) in picocli 4.0. I have no plans to remove `AbstractParseResultHandler` from the library in the near future; I want picocli to be backwards compatible as much as possible.

That said, the docs need updating. Are you interested in providing a pull request to update the programmatic API guide for picocli 4.0?

remkop:
@rbresalier There have been some updates. Can you check the latest version and provide feedback? Thanks!

rbresalier:
Thanks @remkop ! It is a bit hard to review it because I first went to https://github.com/remkop/picocli/wiki/Programmatic-API, but that is not updated yet. So I clicked on the 3 commits and tried to read that, it does look like you updated it. Is there a page with the final result similar to https://github.com/remkop/picocli/wiki/Programmatic-API that I could review?

Some words about why and how I use the Programmatic API.

In my applications, I have a common library that adds and handles a few options that are common to all of my applications that use the library. Then each application adds its own custom options. So part of the options are specified in a common library, and others are specified in the application that uses the library.

I could not see how to accomplish this with the annotations API, it seems you can't have some options defined in one common source library and the rest of the options defined in custom applications. This is why I chose the programmatic API.

I'm also coming from Python using the argparse library, where I could handle common options quite easily, similar to CommandSpec and parseResult=parseArgs().

So the way I do this is:

1) Create a CommandSpec in the application
2) Application adds its options to CommandSpec
3) Application calls a function in the common library which does
  3.1 Adds its common options to the CommandSpec
  3.2 calls parseResult = parseArgs()
  3.3 handles the common options, mainly setting some common flags and internal variables
  3.4, Returns parseResult back to the caller application.
[...]

PR #889, opened 2019-11-30
Programmatic API guide for picocli 4.1 (#885)

deining:


remkop:
Merged. Thank you!

I will probably replace the old 3.0-programmatic-api document with a link to the latest one.

remkop:
Thanks a lot for the PR! Awesome work!

I changed it a bit to use lambdas which is nice and concise. 

I renamed the file to remove the version number and redirect the old https://picocli.info/picocli-3.0-programmatic-api.html to the new doc. 
That should make maintenance easier in the future. 

I also shortened the first example and expanded the example in the Execute section (near the end). 

Let me know what you think.

deining:
> I changed it a bit to use lambdas which is nice and concise.
> I also shortened the first example and expanded the example in the Execute section (near the end).

Great improvements, thanks for reviewing my PR.
Glad to see the document updated now.

One small glitch I came across:

The usage help message of the first example states:

```
Usage: <main class> [-hV] [-c=COUNT] FILES...
      FILES...        The files to process
  -c, --count=COUNT   number of times to execute
  -h, --help          Show this help message and exit.
  -V, --version       Print version information and exit.
```

If I run the command and specify `--help` as only argument, I'm getting no output (expected: help messaage).
If I run the command and specify `--help foo` as commad line arguments, I'm getting as output: `0 foo`

Is this the intended behaviour?

remkop:
@deining Good catch! That was a bug: our IExecutionHandler should also handle requests for help. Fixed now.

PR #893, opened 2019-12-02
Minor improvements and corrections to groovy examples

deining:
I realised that the README.md of the `picocli-groovy` lists an example that states in the description of the option `algorithm`: 
 
`See [1] for more details.`

There is no footer with a reference [1], though.

Removed this passage.


remkop:
Merged, thank you!!
I still need to update the release notes for this.

PR #911, opened 2020-01-01
Fix syntax error in javadoc

deining:
Also added @Command annotation for sake of clarity.


remkop:
Thank you!!
And happy New Year! :-)

PR #913, opened 2020-01-02
More Java and Kotlin examples

deining:
This patch provides further example files, coded both in Java and Kotlin. References to the newly added files were added to the manual, too.

This patch also corrects the Kotlin MyApp example which currently does not compile:

```
companion object {
    @JvmStatic fun main(args: Array<String>) {
        CommandLine.run(MyApp(), *args)
    }
}
```

Inside the companion object, `CommandLine.run()` is called, however, MyApp implements `Callable` and not `Runnable`.
Since both methods `run()` and `call()` are deprecated, I changed this to `CommandLine(MyApp()).execute(*args)`.

In order to prevent these kind of errors in the future, I would propose to add a task `compileKotlin` to the file `build.gradle` of the `picocli-examples` module.

### DIFF
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -329,7 +329,7 @@ Another idea is to add a separate different option (that could be named `--passw
 
 A command that combines either of these with an interactive `--password` option (with the default `arity = "0"`) allows end users to provide a password without specifying it in plain text on the command line. Such a command can be executed both interactively and in batch mode.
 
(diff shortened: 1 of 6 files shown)


remkop:
@deining Compiling the Kotlin examples is a good idea, not sure why we were not doing this already... :-)

Fixed in #915. 
Thanks for raising this!

PR #914, opened 2020-01-03
Picocli examples: adding minimal i18n demo application (Java + Kotlin)

deining:


remkop:
Very cool example app! Nice work!
Thank you as always!

PR #919, opened 2020-01-13
Added example Gradle project with Kotlin and Graal Native Image

OndrejMalek:
Gradle project setup. Example code copied from picocli-examples/src/main/kotlin/picocli/examples/kotlin/Checksum.kt


remkop:
Merged into master.

PR #963, opened 2020-02-25
Fix broken link

vladimirf7:


remkop:
Merged into master. 
Good catch, thank you!
Keep them coming! :-)

PR #987, opened 2020-04-10
Bump JLine to 3.14.1 and fix #969

### LINKED ISSUE #969: autocomplete & suggestions for nested subcommands not working in Picocli Shell JLine3?
For Picocli Shell JLine3, Is there any support for suggestions & autocomplete of nested subcommands? I tried to accomplish just that by adding a subcommand to one of the subcommands (the "cmd" named command in this case) of the example code at https://github.com/remkop/picocli/tree/master/picocli-shell-jline3 but it did not work. I.e., I can autocomplete the first subcommand but no subcommand of that subcommand.

Based on the description of PicocliCommands it should work.

Thanks in advance!

/ Niklas

mattirn:
Actually  fix #969 only the command tab completion part.

Method `commandDescription()` still fails to return sub-command descriptions. In order to fix this we need next JLine release.


remkop:
@mattirn Merged into master. Thank you for the contribution!

> Method commandDescription() still fails to return sub-command descriptions. In order to fix this we need next JLine release.

Do you mean JLine 3.14.1, or another future JLine release?

mattirn:
JLine 3.14.2 + small modification in picocli-shell-jline3. 

I have already modified JLine and tested also picocli-shell-jline3 using JLine 3.14.2-SNAPSHOT

remkop:
@mattirn Understood. Thanks for the quick response! Looking forward to it!

Issue #991, opened 2020-04-15
Detecting terminal width fails on non-English Windows versions


S-Gaertner:
Dynamically detecting the terminal width (e.g. when specifying `usageHelpAutoWidth` on a command) fails on non-English Windows versions and runs into a 2 second timeout.

The output of `cmd.exe /c mode con` is the following on a German Windows:
```
Status von Gerät CON:
---------------------
    Zeilen:          9001
    Spalten:         148
    Wiederholrate:   31
    Verzögerungszeit:1
    Codepage:        850
```

This obviously fails to match the regex pattern in `CommandLine.java:6573`:
```
Pattern pattern = Pattern.compile(".*olumns(:)?\\s+(\\d+)\\D.*", Pattern.DOTALL);
```

The timeout occurs because termination of the helper process is not detected directly but by monitoring the `size` value - which does not change in this case.

I'd like to propose two improvements:
* A more robust regex that also suits international Windows versions. That might be tricky though because the general structure of the output might not be comparable between different localizations.
* A second monitoring variable for detecting process termination that is not used for any other purpose. This allows to fail faster and thus improves user experience.

remkop:
Thank you for raising this!
I like your proposed improvements. Will you be able to provide a pull request?

S-Gaertner:
Sure, I can do that!

remkop:
Merged the PR into master.
Thank you again for raising this and thank you for the contribution!

PR #993, opened 2020-04-16
#991 more robust terminal width detection for international Windows versions

S-Gaertner:
* changed terminal width detection regex pattern for Windows
* check for process termination in order to fail faster

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -6679,7 +6679,9 @@ public class CommandLine {
                                 txt += " " + line;
                             }
                             tracer.debug("getTerminalWidth() parsing output: %s%n", txt);
-                            Pattern pattern = Pattern.compile(".*olumns(:)?\\s+(\\d+)\\D.*", Pattern.DOTALL);
+                            Pattern pattern = (Help.Ansi.isWindows() && !Help.Ansi.isPseudoTTY())
+                                    ? Pattern.compile(".*?:\\s*(\\d+)\\D.*?:\\s*(\\d+)\\D.*", Pattern.DOTALL)
+                            		: Pattern.compile(".*olumns(:)?\\s+(\\d+)\\D.*", Pattern.DOTALL);
                             Matcher matcher = pattern.matcher(txt);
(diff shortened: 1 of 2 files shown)


src/main/java/picocli/CommandLine.java
```diff
@@ -6697,7 +6699,7 @@ public void run() {
                 do {
                     if (size.intValue() >= 0) { break; }
                     try { Thread.sleep(25); } catch (InterruptedException ignored) {}
-                } while (System.currentTimeMillis() < now + 2000);
+                } while (System.currentTimeMillis() < now + 2000 && t.isAlive());
```
remkop, line 6702: Looks good. Were you able to test that this prevents the 2 second timeout when the pattern cannot match anything (e.g. with the old pattern on a German language environment)?

S-Gaertner: Yes, I manually tested that on my system with the old pattern. The 2 second delay was gone as soon as I introduced the alive test for the thread.

remkop: Great, thank you!

remkop:
Overall looks good. I have some other work in progress but I will try to merge this soon.

S-Gaertner:
You're welcome! Thank you for this great project :)

remkop:
Can I ask:

* What is it that you like about picocli?
* What would you do if picocli did not exist?

S-Gaertner:
What I like about picocli:
* It allows complex use cases, yet is pretty easy to use - especially with the convenience methods.
* It has zero dependencies - I am reluctant to include dependencies that themselves entail more dependencies.
* It has superb documentation.
* It is open source, mature, active, feature-rich and has a friendly community.

What I would do without picocli:
* I already have rolled my own command line parser for simple uses cases. Now my use cases grew more complex, so I probably would have adapted it to some degree and also simplified my use cases (and at the same time would provide more user training for the probably more complicated command line invocation).
* I also would have looked into alternatives like JCommander.

remkop:
@S-Gaertner Great feedback, thank you! :+1:

PR #997, opened 2020-04-20
Fix not shown parameter heading under specific condition

### LINKED ISSUE #984: Parameters heading not shown when `@filename` is the only parameter
This issue was introduced by #795 (Show `@filename` in options list).

The problem is that the `@filename` entry in the usage help is listed under positional parameters, but when there are no other positional parameters, the Parameters heading won't be shown.

Test to reproduce the problem:

```java
@Test
public void testAtFileParameterListHeadingShownIfNoOtherPositionalParameters() {
    @Command(name = "A", mixinStandardHelpOptions = true,
            showAtFileInUsageHelp = true,
            parameterListHeading = "Parameters:%n",
            optionListHeading = "Options:%n",
            description = "... description ...")
    class A { }

    String actual = new CommandLine(new A()).getUsageMessage();
            "Usage: A [-hV] [@<filename>...]%n" +
            "... description ...%n" +
            "Parameters:%n" +
            "      [@<filename>...]   One or more argument files containing options.%n" +
            "Options:%n" +
            "  -h, --help             Show this help message and exit.%n" +
            "  -V, --version          Print version information and exit.%n");
    assertEquals(expected, actual);
}
```

This test fails in picocli 4.2 because the actual usage help message is not displaying the `Parameters:` heading:

```
Usage: A [-hV] [@<filename>...]
... description ...
      [@<filename>...]   One or more argument files containing options.
Options:
  -h, --help             Show this help message and exit.
  -V, --version          Print version information and exit.
```

Wycers:
Fix #984 
Add judgment on atFileParameterList.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -13775,11 +13775,19 @@ public class CommandLine {
         }
 
         /**
+         * Returns true if has file parameter in the list, otherwise false.
+         * @return
+         */
+        public boolean hasAtFileParameterList() {
+            return commandSpec.parser.expandAtFiles() && commandSpec.usageMessage.showAtFileInUsageHelp();
+        }
+
+        /**
          * Returns the section of the usage help message that lists the @-file and its description.
          * @return the section of the usage help message that lists the @-file and its description
          */
         public String atFileParameterList() {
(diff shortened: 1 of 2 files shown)


remkop:
Merged into master.

PR #1002, opened 2020-04-22
Fix #988

### LINKED ISSUE #988: Option group sections should include subgroup options
Currently, if an argument group has an option and a nested sub-group, the usage help will only show the options that are _directly_ in the group under the heading for that group.

Instead, the usage help should also include the options for the nested subgroups, unless this subgroup has its own heading.

Wycers:
Fix #988 by using a new method to work out options need or not need to be considered.


remkop:
Sorry for the delay in reviewing your PR.
Merged into master.
Thank you very much!

Issue #845, opened 2019-10-31
Use ANSI colors and styles in error handlers


remkop:
Use ANSI colors and styles in error handlers. For example:

* show invalid usage error messages in bold and red by default
* TBD what style to use for the stack trace printed for execution exceptions and unhandled exceptions

It may make sense to update the `CommandLine.Help.ColorScheme` class to make these configurable.

jerrylususu:
Hello! I'm considering to contribute to this project, and I'd like to know whether the issue is open to pull request. Is there any suggestions for implementing this feature?

remkop:
Hello, thank you for considering contributing! 
PRs are welcome, especially if they have tests and ideally documentation. 
I haven’t thought about implementation, but it may not be easy... :-)

jerrylususu:
I'm trying to implement this function, but I meet some design decisions, and would like to know what's the best way in your mind.

My current plan is:
1. Add 2 attributes `errors` and `stackTraces` to `ColorScheme`
2. Change `internalHandleParseException` method, let writer write `colorScheme.errorText(ex.getMessage())`, instead of just `ex.getMessage()`, so argument parse error can get the ANSI color required.
3. Create a default implementation of `IExecutionExceptionHandler`, customize a `ColoredPrintWriter` with `ColorScheme` or `IStyle[]` in constructor, and when invoking `ex.printStackTrace`, instead of using `cmd.getErr()`, use the new ColoredPrintWriter instead.

However, designing in such way will lead to some problems:
1. `ColorScheme` contains color settings beyond help info, so it might be necessary to move `ColorScheme` to the same layer of `Help`, to show that `ColorScheme` represents general settings of ANSI colors. But doing so will introduce some serious incompatibility problems when users want to update picocli to new version.
2. As the exception message is part of the stack trace, hacking `PrintWriter` will make the exception message has the same style as the stack trace, which might not be intended behavior.

I would like to know what do you think, so I can move forward. Thanks!

By the way, here is a peek of my current progress. I'm using `Style.fg_red` and `Style.bold` for parse errors, and `Style.italic` for stack traces.

![image](https://user-images.githubusercontent.com/17522475/80103404-c38bb600-85a6-11ea-8257-f71e87301942.png)

jerrylususu:
[...]

PR #1008, opened 2020-04-25
Attempting to fix issue #845

jerrylususu:
Changes:
1. Add `errors` and `stackTraces` to Help.ColorScheme
2. Change the `internalHandleParseException` method to use `colorScheme.errorText(ex.getMessage())`, instead of just `ex.getMessage()`
3. Add a method called `exceptionToColorString` to convert exceptions to ANSI colored strings.
4. Change the `handleUnhandled` method to use `exceptionToColorString`

Tests: existing tests updated, no new test introduced
Documentation: JavaDoc updated, no change to other documentations

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -1653,7 +1653,7 @@ public class CommandLine {
             internalHandleParseException(ex, new PrintWriter(err(), true), colorScheme()); return returnResultOrExit(null); }
 
         static void internalHandleParseException(ParameterException ex, PrintWriter writer, Help.ColorScheme colorScheme) {
-            writer.println(ex.getMessage());
+            writer.println(colorScheme.errorText(ex.getMessage()));
             if (!UnmatchedArgumentException.printSuggestions(ex, writer)) {
                 ex.getCommandLine().usage(writer, colorScheme);
             }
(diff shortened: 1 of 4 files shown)


src/main/java/picocli/CommandLine.java
```diff
@@ -1929,10 +1929,30 @@ public int execute(String... args) {
...
 
+    /**
+     * Convert an {@code Exception} to a {@code String} , with message and stack traces extracted and colored
+     * according to {@code ColorScheme}.
+     * @param ex the {@code Exception} to be converted
+     * @param colorScheme the {@code ColorScheme} to use
+     * @return converted and colored {@code String}
+     */
+    private static String exceptionToColorString(Exception ex, Help.ColorScheme existingColorScheme){
+        Help.ColorScheme colorScheme = new Help.ColorScheme.Builder(existingColorScheme).applySystemProperties().build();
```
remkop, line 1945: Out of interest: why not use the specified `existingColorScheme` directly? Is there a reason to make a copy?

src/main/java/picocli/CommandLine.java
```diff
@@ -1929,10 +1929,30 @@ public int execute(String... args) {
         }
     }
     private static int handleUnhandled(Exception ex, CommandLine cmd, int defaultExitCode) {
-        ex.printStackTrace(cmd.getErr());
+        cmd.getErr().print(exceptionToColorString(ex, cmd.getColorScheme()));
+        cmd.getErr().flush();
```
remkop, line 1933: Is this call to `flush()` necessary?

src/main/java/picocli/CommandLine.java
jerrylususu: Sorry... Just noticed this after a few days. One reason I'm using a copy here is not sure if the system property about the color setting might be changing during runtime, and another reason is this is how `Help` constructor works, so I'd like to comply with what's existing here...

src/main/java/picocli/CommandLine.java
jerrylususu: I added this line to fix a bug I find during testing. It seems that Java's `PrintWriter` won't flush until a line separator is met, so I'd like to add one `flush()` here to make sure the message is indeed printed. In retrospect, this line might not be necessary, as error messages generated with `exceptionToColorString` should already have line separators in it. But... just to be safe?

Issue #765, opened 2019-07-05
split regexp with escaped chars: usage help


deining:
I couldn't reopen #763, so I'm creating a new issue here:
Thanks for helping me out. There is still a minor issue when using split regex:

```
package test;
import picocli.CommandLine;
import picocli.CommandLine.Command;
import picocli.CommandLine.Option;

public class ExecuteTest {
    
    static CommandLine cmd;

    public static void main(String[] args) {
        cmd = new CommandLine(new TestCommand());
        cmd.usage(cmd.getOut());
    }

    @Command(name = "ExecuteTest")
    static class TestCommand implements Runnable {
        
        @Option(names = "x", split = "\\|")
        static String[] parts = {};

        @Override
        public void run() {
        }
    }
}
```

This results in 

```
Usage: ExecuteTest [x=<parts>[\|<parts>...]]...
      x=<parts>[\|<parts>...]
```

IMHO, the backslash shouldn't appear in the usage message since it might confuse the user.

remkop:
I agree. I was aware of this but you're right, it should be fixed.

remkop:
I started to look at this again. I remember now why I didn't do anything about this earlier, it's actually quite tricky. Taking a look at the [spec](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html), you'll see that there are escaped characters that have special meaning and should not be "unescaped":

Regex|Meaning
-- | --
\d | A digit: [0-9]
\D | A non-digit: [^0-9]
\s | A whitespace character: [ \t\n\x0B\f\r]
\S | A non-whitespace character: [^\s]
\w | A word character: [a-zA-Z_0-9]
\W | A non-word character: [^\w]
\b | A word boundary
\B | A non-word boundary
\A | The beginning of the input
\G | The end of the previous match
\Z | The end of the input but for the final terminator, if any
\z | The end of the input

Simply removing the leading backslash from such regular expressions would be even more confusing... For example, changing `\d` to `d` completely changes the meaning...

So, here are the options that I can think of (let me know if you can think of alternatives):
* leave all regular expressions as is
* replace only `\|` by `|` and `\\` by `\` and leave all other regular expressions as is

I'm not sure that the second solution won't cause other problems later. I don't like the current situation but I'm not confident that we have a good (or even acceptable) solution yet...

deining:
This is tricky, indeed.

> So, here are the options that I can think of
> (let me know if you can think of alternatives):

[...]

PR #1017, opened 2020-04-30
Fix issue #765

Hannnnnn:
Fix issue: https://github.com/remkop/picocli/issues/765
I add usageSplit property for Option and Parameters.
If the user want to provide a user-friendly usage information without escaped char in regex.
He or she could add usageSplit in  Option and Parameters.
If it is not set by user, usageSpitRegex will be the same as splitRegex in class Argspec and class Builder. So it does not affect the orginal tests.


src/main/java/picocli/CommandLine.java
```diff
@@ -8166,6 +8177,7 @@ private static String describe(ArgSpec argSpec, String separator, String value)
                     description = option.description();
                     descriptionKey = option.descriptionKey();
                     splitRegex = option.split();
+                    usageSplitRegex = option.usageSplit().equals("") ? splitRegex : option.usageSplit();
```
NewbieOrange, line 8180: Why not isEmpty()?

src/main/java/picocli/CommandLine.java
```diff
@@ -8166,6 +8177,7 @@ private static String describe(ArgSpec argSpec, String separator, String value)
                     description = option.description();
                     descriptionKey = option.descriptionKey();
                     splitRegex = option.split();
+                    usageSplitRegex = option.usageSplit().isEmpty() ? splitRegex : option.usageSplit();
```
remkop, line 8180: Let's just store the annotation value (and we can decide later whether to use `splitRegex` or `usageSplitRegex`):

```
usageSplitRegex = option.usageSplit();
```

src/main/java/picocli/CommandLine.java
```diff
@@ -8197,6 +8209,7 @@ private static String describe(ArgSpec argSpec, String separator, String value)
                         description = parameters.description();
                         descriptionKey = parameters.descriptionKey();
                         splitRegex = parameters.split();
+                        usageSplitRegex = parameters.usageSplit().equals("") ? splitRegex : parameters.usageSplit();
```
remkop, line 8212: Let's just store the annotation value (and we can decide later whether to use `splitRegex` or `usageSplitRegex`):

```
usageSplitRegex = parameters.usageSplit();
```

src/main/java/picocli/CommandLine.java
```diff
@@ -8370,7 +8383,12 @@ private static String inferLabel(String label, String fieldName, ITypeInfo typeI
                 public T converters(ITypeConverter<?>... cs) { this.converters = Assert.notNull(cs, "type converters").clone(); return self(); }
 
                 /** Sets a regular expression to split option parameter values or {@code ""} if the value should not be split, and returns this builder. */
-                public T splitRegex(String splitRegex)       { this.splitRegex = Assert.notNull(splitRegex, "splitRegex"); return self(); }
```
remkop, line 8373: Let's leave the existing `splitRegex(String)` setter as it is, and add a separate new setter `ArgSpec.Builder::usageSplitRegex(String)`.

We also need new getters `String ArgSpec.Builder::usageSplitRegex()` and  `String ArgSpec::usageSplitRegex()`.

src/main/java/picocli/CommandLine.java
```diff
@@ -14510,7 +14528,7 @@ public Text renderParameterLabel(ArgSpec argSpec, Ansi ansi, List<IStyle> styles
                 if (argSpec.hideParamSyntax()) { return colorScheme.apply((argSpec.isOption() ? separator() : "") + argSpec.paramLabel(), styles); }
 
                 Text paramName = colorScheme.apply(argSpec.paramLabel(), styles);
-                String split = argSpec.splitRegex();
+                String split = argSpec.usageSplitRegex();
```
remkop, line 14531: This is where we choose whether to use usageSplitRegex or splitRegex:

```java
String split = isEmpty(argSpec.usageSplitRegex()) ? argSpec.splitRegex() : argSpec.usageSplitRegex();
```

src/test/java/picocli/test/UsageSplitExecution.java
```diff
@@ -0,0 +1,24 @@
...
+      public void run() {
+
+      }
+    }
+
+    public static void main (String []args) {
+      CommandLine commandLine = new CommandLine(new UsageSplitCommand());
+      commandLine.usage(commandLine.getOut());
+    }
+}
```
remkop, line 24: Let's make this a JUnit test:

```java
package picocli;

import org.junit.Test;
import picocli.CommandLine.Command;
import picocli.CommandLine.Option;

import static org.junit.Assert.assertEquals;
import static picocli.CommandLine.Help.Ansi.OFF;

public class UsageSplitExecutionTest {
    @Test
    public void testSimple() {
        @Command(name = "UsageSplit")
        class Simple {
            @Option(names = "v", split = "\\|", usageSplit = "|")
            String args[] = {};
        }

                "Usage: UsageSplit [v=<args>[|<args>...]]...%n" +
                "      v=<args>[|<args>...]%n" +
                "%n");
        String actual = new CommandLine(new Simple()).getUsageMessage(OFF);
        assertEquals(expected, actual);
    }
    
    // more tests for using ArgSpec.Builder usageSplit getter and setter
}
```

Hannnnnn:
Thank you for reply.
I have followed you advices, change the code and add some tests.

remkop:
Merged into master. 
Thank you for helping making picocli a better library!

PR #1019, opened 2020-05-02
DOC: Fix PrintExceptionMessageHandler example

hosmanadam:


remkop:
@hosmanadam Thanks for the PR!

The `index.html` file is generated from the `index.adoc` file. We should fix the `index.adoc` file too or your changes will be overwritten the next time index.html is generated.

To test, run:

```
gradlew ascii
```

Then check the generated HTML  in build/docs/html5/index.html
Thanks again for helping to make picocli a better library!

hosmanadam:
@remkop Thanks for the heads up, I made that change as well.
Least I can do, thanks for an awesome tool :)

remkop:
Merged and updated release notes.

Can I ask:

* what is it that you like about picocli?
* what would you do if picocli did not exist?

hosmanadam:
@remkop Sure. I like that it works well and is conveniently annotation-configured, which is a familiar workflow when you come from Spring. If it didn't exist, I'd build our CLI in [Click](https://click.palletsprojects.com/) instead, but then no one else on the team could touch it, because they're Java guys.

remkop:
Thank you!

Enjoy picocli! 😉

PR #1020, opened 2020-05-02
Added Case-insensitive support for Option and related ParseResult, also with some test cases

ifsheldon:
Added `@caseInsensitve` annotation to `Option`.

Added proxy classes to specially handle method requests(`put()`, `containsKey()`, etc.) to `optionsByNameMap`, `negatedOptionsByNameMap` and `posixOptionsByKeyMap` in `CommandSpec` so that logics of codes that access these three `Map` can be left unmodified.

Also, in `CommandSpec`, added codes that specially handle finding case-insensitive options in `findOption()`, so that codes in `ParseResult` can be left unmodified while still provide case-insensitive support.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -3524,6 +3524,17 @@ public class CommandLine {
         boolean versionHelp() default false;
 
         /**
+         *
+         * Set whether to allow an option to be used regardless of cases.
+         * <p>If an option is case-insensitive, for example {@code @Option(names = "option", case-insensitive = true) boolean option;}
+         * then names like "OptiOn" or "opTIoN" are acceptable</p>
+         * <p>However, a {@link DuplicateOptionAnnotationsException} will be thrown when defining multiple options(and at least one is set to case-insensitive) with the same name(regardless cases)</p>
+         * @return whether an option can be called regardless cases.
+         * @since 4.2.1
+         */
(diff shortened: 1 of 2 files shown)


ifsheldon:
PS: the test case `testTracingDebugWithSubCommands()` in `CommandLineTest` fails on my computer, for there is one extra output `[picocli DEBUG] Checking required args for parent command 'git' (user object: picocli.Demo$Git@)...` between `[picocli DEBUG] Found subcommand 'commit' (command 'git-commit' (user object: picocli.Demo$GitCommit@))` and `[picocli DEBUG] Initializing command 'git-commit' (user object: picocli.Demo$GitCommit@): 8 options, 1 positional parameters, 0 required, 0 groups, 0 subcommands.`

It seems to me acceptable, but you may want to check whether this is an issue.

remkop:
@ifsheldon Thanks again for helping to explore the solution space and clarify the requirements!
Your contribution helped made Case Insensitivity (#9) a better feature and is credited in the [release notes](https://github.com/remkop/picocli/blob/master/RELEASE-NOTES.md#4.3.0-fixes).

remkop:
> the test case testTracingDebugWithSubCommands() in CommandLineTest fails on my computer, for there is one extra output [picocli DEBUG] Checking required args for parent command 'git' (user object: picocli.Demo$Git@)...

Sorry about that: I pushed some changes but without updating that test.

PR #1021, opened 2020-05-02
Another implementation of case-insensitive support for subcommands and options (#9)

NewbieOrange:
This is an alternative PR to #1020 , follows a somewhat different approach, trying to solve #9 .

Currently it is only a draft, feedbacks are welcomed. Tests are partially added, not fully tested.

No previous tests are broken due to the new changes.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -761,12 +761,12 @@ public class CommandLine {
 
     /** Returns whether the command should be case-insensitive in subcommands. The default is {@code false}.
      * @return {@code true} if subcommands can be specified that don't match the {@code names()} value of the registered ones, {@code false} otherwise;
-     * e.g., for a subcommand of names help, inputs like help, hElp and HELP are all recognized as it if {@code true}.
+     * e.g., for a subcommand of names help, inputs like help, HeLp and HELP are all recognized as it if {@code true}.
      * @since 4.3 */
     public boolean isCaseInsensitiveSubcommands() { return getCommandSpec().caseInsensitiveSubcommands(); }
 
     /** Sets whether the command should be case-insensitive in subcommands. The default is {@code false}.
-     * When set to true, for example, for a subcommand of names help, inputs like help, hElp and HELP are all recognized as it if {@code true}.
+     * When set to true, for example, for a subcommand of names help, inputs like help, HeLp and HELP are all recognized as it if {@code true}.
      * <p>The specified setting will be registered with this {@code CommandLine} and the full hierarchy of its
      * subcommands and nested sub-subcommands <em>at the moment this method is called</em>. Subcommands added
      * later will have the default setting. To ensure a setting is applied to all
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -5340,9 +5340,9 @@ public class CommandLine {
          * This class provides a case-aware Linked HashMap. Supports both case-sensitive and case-insensitive modes.
          * @param <V> type of the value
          */
-        private static class LinkedCaseAwareMap<V> implements Map<String, V> {
-            private final LinkedHashMap<String, V> targetMap = new LinkedHashMap<String, V>();
-            private final HashMap<String, String> keyMap = new HashMap<String, String>();
+        private static class LinkedCaseAwareMap<K, V> implements Map<K, V> {
+            private final LinkedHashMap<K, V> targetMap = new LinkedHashMap<K, V>();
+            private final HashMap<K, K> keyMap = new HashMap<K, K>();
             private final Locale locale;
             private boolean caseInsensitive = false;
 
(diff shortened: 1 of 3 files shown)


src/main/java/picocli/CommandLine.java
```diff
@@ -5565,6 +5781,28 @@ protected CommandSpec commandLine(CommandLine commandLine) {
...
+             * @since 4.3 */
+            public CommandSpec caseInsensitive(boolean caseInsensitive) { caseInsensitiveCommands(caseInsensitive); caseInsensitiveOptions(caseInsensitive); return this; }
+            
+
+            /** Returns whether the commands is case-insensitive.
+             * @since 4.3 */
+            public boolean caseInsensitiveCommands() { return commands.isCaseInsensitive(); }
+            /** Sets the case-insensitivity of commands.
+             * @since 4.3 */
+            public CommandSpec caseInsensitiveCommands(boolean caseInsensitiveCommands) { commands.setCaseInsensitive(caseInsensitiveCommands);  return this; }
```
NewbieOrange, line 5797: Probably catch the IllegalStateException from LinkedCaseAwareMap and rethrow a wrapped exception here?

src/main/java/picocli/CommandLine.java
```diff
@@ -5565,6 +5781,28 @@ protected CommandSpec commandLine(CommandLine commandLine) {
...
+
+            /** Returns whether the options is case-insensitive.
+             * @since 4.3 */
+            public boolean caseInsensitiveOptions() { return optionsByNameMap.isCaseInsensitive(); }
+            /** Sets the case-insensitivity of options.
+             * @since 4.3 */
+            public CommandSpec caseInsensitiveOptions(boolean caseInsensitiveOptions) { optionsByNameMap.setCaseInsensitive(caseInsensitiveOptions); negatedOptionsByNameMap.setCaseInsensitive(caseInsensitiveOptions); posixOptionsByKeyMap.setCaseInsensitive(caseInsensitiveOptions);  return this; }
```
NewbieOrange, line 5804: Same as above.

remkop:
I’ll do another check tomorrow. 

Can you think of tests for any edge cases introduced by this change? Like ambiguities that didn’t exist before? I can think of cases where the combination of POSIX options now resembles a long option, or a long option with the same name as a subcommand, it may be good to have tests for those. 

Can anyone think of other tests?

NewbieOrange:
> 
> 
> I’ll do another check tomorrow.
> 
> Can you think of tests for any edge cases introduced by this change? Like ambiguities that didn’t exist before? I can think of cases where the combination of POSIX options now resembles a long option, or a long option with the same name as a subcommand, it may be good to have tests for those.
> 
> Can anyone think of other tests?

I did some preliminary tests, yes the combination of POSIX options can resemble a long option, but this behaviour seems to be the same as the original, i.e. without case-insensitive set, it is still possible to have the aforementioned situation.

NewbieOrange:
The new LinkedCaseAwareMap should be transparent to the rest of the code, and as long as picocli is using Map#get and Map#containsKey to check for duplications, the bahaviour should be reasonable.

There will of course be new ambiguities, like -ABCD and -a, -b, -c, -d will be ambiguous with case-insensitive set, but this behaviour should be intuitive as it is the same as -abcd and -a, -b, -c, -d without case-insensitivity.

There might be some edge cases tho, but none that I am aware of, yet.

remkop:
I see this test is failing in Travis CI, can you take a look?

```
picocli.ModelCommandSpecTest > testCommandSpecAddSubcommand_DisallowsDuplicateSubcommandAliases FAILED
    org.junit.ComparisonFailure: expected:<...'a' for subcommand '[x]' is already used by...> but was:<...'a' for subcommand '[a]' is already used by...>
        at org.junit.Assert.assertEquals(Assert.java:115)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at picocli.ModelCommandSpecTest.testCommandSpecAddSubcommand_DisallowsDuplicateSubcommandAliases(ModelCommandSpecTest.java:1184)
```

NewbieOrange:
Fixed the failing test. Also, I've updated the duplicate subcommand / option exception message to use the previous registered ones case-sensitive name, e.g. If there is already `-h` and we add option `-H`, it will now throws `Option name '-h' is used`.

remkop:
Great, thank you! I think we are getting close.

Codecov reports that code coverage will go down if this PR is merged.

Can you add tests for the code changes that are [not or only partially covered by the tests yet](https://codecov.io/gh/remkop/picocli/pull/1021/diff?src=pr&el=tree#diff-c3JjL21haW4vamF2YS9waWNvY2xpL0NvbW1hbmRMaW5lLmphdmE=)?

NewbieOrange:
Since LinkedCaseAwareMap is a private class (and probably should be), some code are not called by picocli (not public methods that can reach these lines) and they are there for potential future error-checking.

I could add some tests using reflection on the custom map, but does Codecov count that as 'coverage'?

NewbieOrange:
For example, uncovered lines are checks for non-convertible classes, null keys, removing elements, which are not called by any other picocli code.

NewbieOrange:
Now the coverage shall increase after this PR is merged! 😃

NewbieOrange:
Pushing a small commit to try to trigger Codecov again for a correct report.

remkop:
That is great! Thank you very much.
This is ready to be merged.

Do you think you will be able to add some documentation to the [parser  configuration section](https://picocli.info/#_parser_configuration) of the user manual (in the `docs/index.adoc` file)? If not, I can do it.  😄

NewbieOrange:
I am not a native English speaker and not really familiar with asciidoc 😅.

It would probably be better if you could add those documentation.

remkop:
okay, no worries.

jerrylususu:
> okay, no worries.
> Thank you for the contribution!

I'm glad to help on writing documentation. Is `docs/index.adoc` the only file need to change? However, I need some time to look into the changes and figure out how to document it clearly.

remkop:
Yes index.adoc is the main file for documentation. 
I’ll also make an entry in RELEASE-NOTES.md. 

To test, run:

```
gradlew ascii
```

Then check the generated HTML  in build/docs/html5/index.html
Thanks again for helping to make picocli a better library!

PR #1023, opened 2020-05-04
Finished docs on case sensitivity (#9), and add some examples.

jerrylususu:
Hope this helps.


remkop:
@jerrylususu Nice work!
Thank you very much!

PR #1024, opened 2020-05-05
fix issue #995

### LINKED ISSUE #995: Reset to initial value for multi-value options/positional params
The initial value for an array, collection or map option is not preserved and applied if the user input omits the option/positional on the command line.

Picocli currently clears (actually creates a new instance of) collections and arrays
(see comment "existing values are default values if `initialized` does NOT contain argsSpec"  in `CommandLine::applyValuesToArrayField`).

But the value is only cleared and reset to en empty array or collection if the user actually specifies a value...
The initial value is not preserved and applied if the user input omits the option/positional.

Related to #877 and #967.

Wycers:
fix #995

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -12391,6 +12391,13 @@ public class CommandLine {
                 throw new MissingParameterException(CommandLine.this, argSpec, "Expected parameter " + desc + "for " + optionDescription("", argSpec, -1) + " but found '" + args.peek() + "'");
             }
         }
+
+        /**
+         * Checks whether the initial value for the collection is changed while collection is changed. Recover the initial value if is changed unexpectedly.
+         * @param args the command line arguments
+         * @param initialized set of initialized that parsed from the CommandLine
+         * @param collection collection that parsed from consumeArguments
+         */
         private void fixInitialValueIfSetted(Stack<String> args, Set<ArgSpec> initialized,Collection<Object> collection) {
             if(args.empty()) {
                 List<ArgSpec> list = new ArrayList<ArgSpec>(initialized);
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -12391,6 +12391,15 @@ public class CommandLine {
                 throw new MissingParameterException(CommandLine.this, argSpec, "Expected parameter " + desc + "for " + optionDescription("", argSpec, -1) + " but found '" + args.peek() + "'");
             }
         }
+        private void fixInitialValueIfSetted(Stack<String> args, Set<ArgSpec> initialized,Collection<Object> collection) {
+            if(args.empty()) {
+                List<ArgSpec> list = new ArrayList<ArgSpec>(initialized);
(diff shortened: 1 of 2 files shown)


remkop:
@Wycers @Linyer-qwq Thank you for the pull request!

I added some more tests (currently `@Ignored`) that currently fail, with `@Parameters`-annotated `Set`, `Map` and `EnumSet` fields.

Your fix surprised me: I thought something more difficult would be necessary. 👍 😅 

I did more analysis and added a comment on #995. The fix may be even simpler than your proposal, perhaps this is enough:

```
// applyValuesToCollectionField (and similar change in applyValuesToMapField)
if (collection == null || ( /*!collection.isEmpty() &&*/  !initialized.contains(argSpec))) {
```

Can you explore this?

Wycers:
Thanks for adding more tests, we will follow up.

remkop:
@Wycers @Linyer-qwq any update?

I am planning to finish up some items over the weekend (hopefully including a fix for #995) and release picocli 4.3. 

Please let me know if you are still working on this.

Wycers:
We're still working on this, but it's a little difficult for us.

No worries if you want to fix it. We're happy to see that happens!

Linyer-qwq:
Maybe you could give us a little more time....XD

remkop:
@Wycers @Linyer-qwq Thank you for the quick reply!

> Maybe you could give us a little more time....XD

Okay. Can you give me some indication of how much time?

I suggest the following approach:

* I did some analysis and wrote a summary in [this comment](https://github.com/remkop/picocli/issues/995#issuecomment-624064425) in #995. Because of that analysis, I believe the below may be a good way to fix it:
```
// applyValuesToCollectionField (and similar change in applyValuesToMapField)
if (collection == null || ( /*!collection.isEmpty() &&*/  !initialized.contains(argSpec))) {
```
* This fix will likely cause some new test failures, specifically, tests that verify that empty initial instances of collections or maps are reused. For example:
```
@Parameters List<File> list = new ArrayList(); // this instance used to be reused but after our change it is not reused
```
* With the new logic we will no longer reuse and modify that instance. When the first value is matched we will create a new ArrayList (or other collection or Map), and put the values into that new instance.
* So, the tests that verify reuse will fail, and that is okay, because we do not want to reuse this instance. For those tests, we need to change the expectation. (Expect _no reuse_.)
* However, if _no value is matched_, then we expect that our initial value is unchanged.
* There is one test with an `ArrayBlockingQueue` that fails because there is public no-arg constructor. That test can be `@Ignore`-ed.

Linyer-qwq:
Sorry to reply so late. I am following that "change the expectation. (Expect no reuse.)" So most of the failed testcase pass now. And now I am working on fix testcase "testMapFieldHappyCase". I thought i will work out it tonight.

remkop:
No problem at all. 
Take your time; I’ll probably do the release on Sunday.

Wycers:
We give it a new try... (#1035)

remkop:
I merged #1035, so I will close this pull request.
(This PR is still mentioned in the release notes for picocli 4.3 though.)

PR #1029, opened 2020-05-06
Check for duplicate negated when adding non-negatable option (#1022)

### LINKED ISSUE #1022: Duplicate negated options are accepted
During the tests for the case-insensitive issue, it seems that I found one somewhat interesting issue(bug?). 

```java
    @Test
    public void testDuplicateNegatedOptionsRejected()
    {
        @CommandLine.Command(name = "negatable-options-demo")
        class NegatableOptionsDemo
        {
            @CommandLine.Option(names = "--verbose", negatable = true)
            boolean verbose= false;
            @CommandLine.Option(names = "--no-verbose")
            boolean nverbose = false;
        }
        NegatableOptionsDemo d = new NegatableOptionsDemo();
        CommandLine cmd = new CommandLine(d);
        cmd.execute("--verbose");
        assertTrue(d.verbose);
        assertFalse(d.nverbose);
        cmd.execute("--no-verbose");
        assertFalse(d.verbose); // it seems both fields have been set
        assertTrue(d.nverbose); // it seems both fields have been set
    }
```

I think this test should throw a duplication execption, but it passed.

NewbieOrange:
Fixes #1022 , a new test is added.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -6005,10 +6005,8 @@ public class CommandLine {
                         throw DuplicateOptionAnnotationsException.create(existingName, option, existing);
                     }
                     // #1022 checks if negated option exists with the same name
-                    String negatedName = negatableOptionTransformer().makeNegative(name, this);
-                    String existingNegatedName = optionsByNameMap.getCaseSensitiveKey(negatedName);
-                    OptionSpec existingNegated = negatedOptionsByNameMap.get(negatedName);
-                    if (existingNegated == null) { existingNegated = optionsByNameMap.get(negatedName); }
+                    String existingNegatedName = negatedOptionsByNameMap.getCaseSensitiveKey(name);
+                    OptionSpec existingNegated = negatedOptionsByNameMap.get(name);
                     if (existingNegated != null && existingNegated != option) {
                         throw DuplicateOptionAnnotationsException.create(existingNegatedName, option, existingNegated);
                     }
@@ -6039,7 +6037,7 @@ public class CommandLine {
                             tracer.debug("Option %s is negatable, but has no negative form.%n", name);
                         } else {
                             tracer.debug("Option %s is negatable, registering negative name %s.%n", name, negatedName);
-                            String existingName = optionsByNameMap.getCaseSensitiveKey(negatedName);
+                            String existingName = negatedOptionsByNameMap.getCaseSensitiveKey(negatedName);
                             OptionSpec existing = negatedOptionsByNameMap.put(negatedName, option);
                             if (existing == null) { existing = optionsByNameMap.get(negatedName); }
                             if (existing != null) {
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -6115,19 +6115,6 @@ public class CommandLine {
                 }
             }
 
-            private void validateNegatedOptions() {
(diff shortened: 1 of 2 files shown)


ifsheldon:
```java
@Test(expected = CommandLine.DuplicateOptionAnnotationsException.class)
    public void duplicateCaseInsensitiveNegatedOptionsNotAccepted()
    {
        @CommandLine.Command(name = "negatable-options-demo")
        class NegatableOptionsDemo
        {
            @CommandLine.Option(names = "--verbose", negatable = true)
            boolean verbose= false;
            @CommandLine.Option(names = "--nO-veRbose")
            boolean nverbose = false;
        }
        NegatableOptionsDemo d = new NegatableOptionsDemo();
        CommandLine cmd = new CommandLine(d).setOptionsCaseInsensitive(true);
        cmd.execute("--verbose");
        fail();
    }
```
Test failed

NewbieOrange:
`RegexTransformer.createDefault()` does not seem to consider --NO, --No, --nO.. so the current PR may not solve those cases. (partially, so if --verbose is non-negatable and --NO-verbose is negatable there are issues, otherwise if it is negatable --verbose and non-negatable --NO-verbose then the exception is correctly thrown)

Any thoughts? Should the transformer also considers the upper-case '--NO'?

remkop:
Thanks for bringing this up!
I created a separate ticket https://github.com/remkop/picocli/issues/1030 for this.

NewbieOrange:
Now this seems to pass all the tests I could think of.

NewbieOrange:
After some thoughts, I think replacing the `negatableOptionTransformer` will likely cause issues.

For example, if the user has their custom transformer set, then we cannot replace theirs. There are several possible choices:

1. Checks whether the transformer equals to the default ones (need to override `equals()` function in class `RegexTransformer`), and only replace if it does, which makes the behaviour more complicated.
2. Replace the default transformer with the new case-insensitive one.
3. Do not replace the transformer, only resets `negatedOptionsByNameMap` in `optionsCaseInsensitive()`. Probably keep the new transformer method in case users want it?

remkop:
@NewbieOrange You raise a good point, but I disagree with your conclusion.

Yes, ideally we want to build the library such that users can call its functions _in any order_, and it will always work correctly. But in reality this is not always possible.

I believe the simplest solution is what we discussed earlier: replace the transformer with a case-insensitive one when `setOptionsCaseSensitive(true)` is called. But we should tell users that this happened.

I propose that we add javadoc to these setter methods: `CommandLine::setOptionsCaseSensitive`, `CommandSpec::optionsCaseSensitive` and `CommandSpec::negatableOptionTransformer` to tell users that this interdependency exists.

Something like: (in setOptionsCaseSensitive)

> Note that changing case sensitivity will also make change the case sensitivity of {@linkplain Option#negatable() negatable} options: any custom {@link INegatableOptionTransformer} that was previously installed will be replaced by the case-insensitive version of the default transformer. To ensure your custom transformer is used, install it last, after changing case sensitivity.

...and a similar message in the javadoc for the other setter methods.

NewbieOrange:
@remkop I think picocli 4.x does allow methods to be called in arbitary order? Making `setOptionsCaseSensitive` an exception kind of breaks the consistency, since users may assume the setters don't have dependencies to other setters.

That said, I've pushed a commit with the new javadoc you proposed.

remkop:
Merged into master.
Thank you very much for the thorough work you did on this pull request! 
I'm impressed!

PR #1032, opened 2020-05-07
Add test case for help.hasAtFileParameter

licia-tia:
Hi, I wrote a test case for help.hasAtFileParameter.


licia-tia:
I'll add more tests and pull request them together :-)

PR #1035, opened 2020-05-08
Remkop#995 v2

Linyer-qwq:
Hi! Does that meet your expectation?  I will be glad if that helps.


remkop:
Merged into master. Thank you @Linyer-qwq, @licia-tia and @Wycers !

PR #1047, opened 2020-05-15
Add abbreviation recognition for subcommands and options

### LINKED ISSUE #10: Support recognizing abbreviated options and commands
http://bailando.berkeley.edu/papers/psb03.pdf

### LINKED ISSUE #732: Support unambiguous abbreviations
Some programs allow subcommands to be abbreviated with any unambiguous abbreviation of the subcommand name.  I would like to request the ability to allow this in picocli.

The key here is that these are not explicitly listed abbreviations, but rather any unambiguous abbreviation of the command name or possibly even its aliases.

For example, in Mercurial, `hg showconfig` can be run with `hg showc`, `hg show` or `hg sho`.  However `hg sh` and `hg s` are ambiguous and generate the following errors.  (At least they do with the set of Mercurial plugins I have installed.  Yours may be different.)
```
$ hg sh
hg: command 'sh' is ambiguous:
    shelve showconfig
$ hg s
hg: command 's' is ambiguous:
    serve shelve showconfig split squash stabilize status strip summary
```

Another example is the `btrfs` filesystem command in which `btrfs fi usa /` and `btrfs f u /` are the same as `btrfs filesystem usage /`.

This concept could also extend to option names.  For example Python's `argparse`, supports this with the `allow_abbrev` setting.

In the cases I have seen this implemented, "unambiguous abbreviation" really means "unambiguous prefix", but there might be generalizations.  (This is not critical to me.  It is just a possible point in the design space.) For example, you might allow "foo-bar" to be abbreviated as "f-b" (under the rule that "-" separated words can be abbreviated individually) or "fb" or "fr" (under the rule that abbreviations are subsequences).  The latter might be impractical to use (e.g., "fr" becomes ambiguous between "free" and "foo-bar").  I only mention it as it is what is used by Emacs Helm, albeit for interactive input not command lines.

NewbieOrange:
Currently using code snippet from @remkop #732 with minor changes.

For ease of debugging, the `AbbreviationMatcher` class ~~is in separate file for _now_~~.

~~This is a draft, no javadoc are added _yet_, and tests may not cover all cases~~.

Feedbacks are welcomed!

This draft's goal is to resolve #10 and resolve #732.

Edit: All javadoc have been added, and diff coverage is 94.79%.


src/main/java/picocli/NameMatcher.java
```diff
@@ -0,0 +1,80 @@
...
+                    matchCount++;
+                }
+                if (matchCount == abbreviatedKeyChunks.size()) {
+                    candidates.add(key);
+                }
+            }
+        }
+        if (candidates.size() > 1) {
+            String str = candidates.toString();
+            throw new IllegalArgumentException(abbreviation + " is not unique: it matches '" +
```
remkop, line 75: Let's throw a `ParameterException` so that the user will see a nice error message.
The error message should probably start with the word "Error", so we get `"Error: " + abbreviation + " is not unique: ` (...)

src/main/java/picocli/NameMatcher.java
```diff
@@ -0,0 +1,80 @@
+package picocli;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+class NameMatcher {
```
remkop, line 7: I am not sure yet where these methods should live...

* they should be easy to test, so package-protected methods of a public or package-protected class
* they should not be in an external top-level class, but somewhere in the `CommandLine.java` source file
* they should not make the picocli jar bigger than necessary (unsure how much difference it makes in size if we add a new static inner class for these methods, but perhaps `ParserSpec` or `CommandLine` is a good place... still thinking...)

src/test/java/picocli/CommandLineTest.java
```diff
@@ -3893,6 +3894,119 @@ public void testCaseInsensitiveParseResult() {
         assertEquals(42, returnCode);
     }
 
+    @Test
+    public void testAbbrevSubcommands() throws Exception {
```
remkop, line 3898: We should also keep the unit tests for the `splitIntoChunks` and `match` methods that are defined in the 
https://github.com/remkop/picocli/issues/732 code snippets if possible.

NewbieOrange:
This currently does not work with case-insensitive mode, because the match function does not take it into consideration.

For POSIX options, the current implementation will prioritize full name, and in this case, also the abbreviation. For example, if we have three options: `-A`, `-B` and `--ABC`, if the input is `-AB`, this PR will match `--ABC` instead of `-A` and `-B`, which is counter-intuitive.

Also, if there are options --a-B, --aB, -a, -B... this will currently be unable to process. (Always throws a ParameterException!)

I think the behaviour of this feature should be: `full-match > posix-match > abbrev-match`? If this is the case, we may make the entire part of pre-processing options to a method, and call with original `arg` first, then try to match POSIX, if both failed, call the method again with `matched arg`.

src/main/java/picocli/NameMatcher.java
NewbieOrange: This class will need information from `CommandSpec` for case-insensitivity, and if it is not in `CommandSpec`, it should be in `CommandLine` to be able to access the getter.

NewbieOrange: Or we can (must) pass the case-insensitivity as an argument to the methods, since the case-insensitivity might be different for subcommands and options, while we only have one `match` method.

NewbieOrange:
Another way is to limit abbreviations to long options only (options starts with `--`), so it will probably not conflict with POSIX options.

NewbieOrange:
Not sure how to `splitIntoChunks` with case-insensitivity...

src/main/java/picocli/CommandLine.java
```diff
@@ -12377,6 +12437,16 @@ private void processArguments(List<CommandLine> parsedCommands,
...
                 // or an option may have one or more option parameters.
                 // A parameter may be attached to the option.
+                if (commandSpec.parser().abbreviatedOptionsAllowed()) {
+                    Set<String> aggregatedOptionNames = new LinkedHashSet<String>();
+                    aggregatedOptionNames.addAll(commandSpec.optionsMap().keySet());
+                    aggregatedOptionNames.addAll(commandSpec.negatedOptionsMap().keySet());
+                    try {
+                        arg = AbbreviationMatcher.match(aggregatedOptionNames, arg);
+                    } catch (IllegalArgumentException ex) {
+                        throw new ParameterException(CommandLine.this, "Error: " + ex.getMessage(), ex);
```
remkop, line 12447: We can pass the `CommandLine` instance as a parameter to the `match` method so it can throw a `ParameterException` instead of an `IllegalArgException`

src/main/java/picocli/AbbreviationMatcher.java
```diff
@@ -0,0 +1,80 @@
+package picocli;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+class AbbreviationMatcher {
+    static List<String> splitIntoChunks(String command) {
```
remkop, line 8: Let's move these methods to `CommandLine`. We should not create a separate source file for this. Sorry I was unclear.

src/main/java/picocli/CommandLine.java
```diff
@@ -12356,6 +12409,13 @@ private void processArguments(List<CommandLine> parsedCommands,
                 }
 
                 // if we find another command, we are done with the current command
+                if (commandSpec.parser().abbreviatedSubcommandsAllowed()) {
+                    try {
+                        arg = AbbreviationMatcher.match(commandSpec.subcommands().keySet(), arg);
+                    } catch (IllegalArgumentException ex) {
+                        throw new ParameterException(CommandLine.this, "Error: " + ex.getMessage(), ex);
```
remkop, line 12416: We can pass the `CommandLine` instance as a parameter to the `match` method so it can throw a `ParameterException` instead of an `IllegalArgException`

src/main/java/picocli/AbbreviationMatcher.java
NewbieOrange: I knew, just make it separate temporarily to make development easier since the editor is more responsive. Will move it inside after the development is done.

src/main/java/picocli/CommandLine.java
NewbieOrange: That kind of breaks the encapsulation?

NewbieOrange:
Any thoughts on behaviour about case-insensitivity and POSIX options mentioned above?

remkop:
Bedtime for me. I’ll take another look tomorrow.

src/main/java/picocli/CommandLine.java
remkop: You mean the `CommandLine.Interpreter::tryConvert` method, right? (line numbers may have changed). Good point!

One big difference is that `ITypeConverter` is a client-facing interface: it is intended to be implemented by users of the library. So we want to make it as simple and easy as possible: "given a String, produce a different value, or throw any exception you want". This interface would be a lot less easy to use if applications could only throw a `ParameterException`, so we put some extra effort in the framework to make it easier for applications.

The `match` method is internal, and is not meant to become public API. It will not be used outside of `CommandLine`, so I am okay with less encapsulation.

src/main/java/picocli/CommandLine.java
```diff
@@ -12377,6 +12437,22 @@ private void processArguments(List<CommandLine> parsedCommands,
                 // A single option may be without option parameters, like "-v" or "--verbose" (a boolean value),
                 // or an option may have one or more option parameters.
                 // A parameter may be attached to the option.
+                if (commandSpec.parser().abbreviatedOptionsAllowed() && arg.startsWith("--")) {
```
remkop, line 12440: Sometimes long (non-POSIX) options can start with a single hyphen `-`, or some other prefix (picocli allows any prefix). Perhaps we can think of some other way to exclude POSIX options.

src/main/java/picocli/NameMatcher.java
remkop: Passing case-insensitivity as an argument the the `match` method makes sense to me.

src/main/java/picocli/CommandLine.java
NewbieOrange: We might enforce the scope of options abbreviation to ones that start with dual hyphens? Otherwise it will inevitably cause ambiguity.

Or we can note in the document that abbreviation only works fully when `setPosixClusteredShortOptionsAllowed(false)`.

Another way is to provide an option for the users to choose what they want...

NewbieOrange: Or make an `abbreviatedTransformer`

remkop:
I haven't had a chance yet to sit down and think clearly about the interaction between case insensitive mode, posix short options, negatable options and abbreviated options. It is possible that some combinations will always give counter-intuitive results.

I would be fine with reducing scope, like leaving negatable options out of scope: for example, we could recognize the abbreviated original option, but the negatable form must always be specified fully and cannot be abbreviated. Just an idea.

The examples you provided are useful to help list up and think through the possible combinations. I just need to find a block of time where I can focus on this. I may not be able to do so this weekend.

Abbreviated commands are easier. One idea is to (initially?) only provide this feature for subcommands.

NewbieOrange:
Actually the negatable form options does work since 747720c.

Feel free to take your time, I am not in a hurry. 😀

NewbieOrange:
It makes sense to me that we only split blocks using hyphen (ignore cammelCase) in case-insensitive mode.

NewbieOrange:
Also, if there are options -A, -B and -AB, if user inputs -AB, picocli will match -AB instead of -A and -B (full option name match has higher priority than POSIX cluster-options). So we can define a priority for options matching and let the users to avoid ambiguity.

NewbieOrange:
Hi @remkop, any new idea on this?

remkop:
Sorry I’ve had other commitments, I’ll try to take another look tomorrow morning.

remkop:
I finally had time to take a look.
Overall, this is very good progress!

* let's move the abbreviation-related tests from CommandLineTest to AbbreviationMatcherTest
* can we add tests for abbreviated options in case-insensitive mode?

On the topic of matching abbreviated options:

* I don't think we should restrict this to options that start with `--`
```java
//picocli.CommandLine.Interpreter#processArguments
if (commandSpec.parser().abbreviatedOptionsAllowed() && arg.startsWith("--")) {
```
should be
```java
if (commandSpec.parser().abbreviatedOptionsAllowed()) {
```

* There may be a bug in our implementation: abbreviations should not abbreviate the non-alphanumeric prefix. If an option is defined as `--x-y` then even in abbreviated mode, we should not match `-x`, `-x-y` or `-xY`, only `--x`, `--x-y` and `--xY`. If we fix that, it should resolve the problem you mentioned:
> ... if there are options `--a-B`, `--aB`, `-a`, `-B`... this will currently be unable to process if user inputs `-aB`. (Always throws a ParameterException!)

* The previous point would also resolve the interaction you mentioned between abbreviations and POSIX options. In your example with three options `-A`, `-B`, and `--ABC`, user input `-AB`, should be matched being matched as `-A` and `-B`, not as the long option `--ABC`, because the non-alphanumeric prefix is significant. However, user input `--AB` will be matched as the long option `--ABC`, not as the POSIX options  `-A` and `-B`. This is unambiguous.

* In case-insensitive mode, users can only abbreviate options with hyphens, so if an option is defined as `-CamelCaseOption` then users can only abbreviate as `-c-c-o`; we will not match `-CCO` as an abbreviation. That is a limitation of combining abbreviations with case-insensitive options. As long as we document this I believe this is fine.

remkop:
Correction: I said

> if an option is defined as `-CamelCaseOption` then users can only abbreviate as `-c-c-o` ...

This is incorrect. In case-insensitive mode, we cannot distinguish chunks by case any more, so we can recognize `-CamelCas` or `-came` as an abbreviation, but not `-c-c-o` or `-CCO`.

We can still recognize chunks (even in case-insensitive mode) if the option is defined in kebab-case, _and_ the user specifies chunks in kebab-case. 

I believe the below tests are reasonable:

```java
@Test
public void testAbbrevOptionsCaseInsensitive1() {
    class App {
        @Option(names = {"-CamelCaseOption"}) boolean camelCaseOption;
    }

    try {
        new CommandLine(new App()).setAbbreviatedOptionsAllowed(true).setOptionsCaseInsensitive(true).parseArgs("-CCO");
        fail("Expected exception");
    } catch (CommandLine.UnmatchedArgumentException ex) {
        assertEquals("Unknown option: '-CCO'", ex.getMessage());
    }

    try {
        new CommandLine(new App()).setAbbreviatedOptionsAllowed(true).setOptionsCaseInsensitive(true).parseArgs("-c-c-o");
        fail("Expected exception");
    } catch (CommandLine.UnmatchedArgumentException ex) {
        assertEquals("Unknown option: '-c-c-o'", ex.getMessage());
    }
    App app = new App();
    new CommandLine(app).setAbbreviatedOptionsAllowed(true).setOptionsCaseInsensitive(true).parseArgs("-came");
    assertTrue(app.camelCaseOption);
}

@Test
public void testAbbrevOptionsCaseInsensitive2() {
    class App {
        @Option(names = {"-kebab-case-option"}) boolean kebabCaseOption;
    }

    try {
        new CommandLine(new App()).setAbbreviatedOptionsAllowed(true).setOptionsCaseInsensitive(true).parseArgs("-KCO");
        fail("Expected exception");
    } catch (CommandLine.UnmatchedArgumentException ex) {
        assertEquals("Unknown option: '-KCO'", ex.getMessage());
    }
    App app = new App();
    new CommandLine(app).setAbbreviatedOptionsAllowed(true).setOptionsCaseInsensitive(true).parseArgs("-k-c-o");
    assertTrue(app.kebabCaseOption);

    app = new App();
    new CommandLine(app).setAbbreviatedOptionsAllowed(true).setOptionsCaseInsensitive(true).parseArgs("-kebab");
    assertTrue(app.kebabCaseOption);
}
```

NewbieOrange:
Now we enforce the non-alphabetic prefix matching on abbrev options, which (hopefully) solves the previous issues.

Case-insensitive abbrev options tests from @remkop added, and they all passed. 😄

NewbieOrange:
Where should `AbbreviationMatcher` go though? Another `static class` in `CommandLine.java` seems good.

remkop:
I'm fine with making `AbbreviationMatcher` a private static class in `CommandLine`.

Question: does this test pass?

```java

@Test
public void testAbbrevOptionsAmbiguous() {
    @Command
    class App {
        @Option(names = "--a-B")
        public boolean a_b;

        @Option(names = "--aB")
        public boolean aB;

        @Option(names = "-a")
        public boolean a;

        @Option(names = "-B")
        public boolean B;
    }

    App app = new App();
    new CommandLine(app).setAbbreviatedOptionsAllowed(true).parseArgs("-aB");
    assertFalse(app.a_b);
    assertFalse(app.aB);
    assertTrue(app.a);
    assertTrue(app.B);

    app = new App();
    new CommandLine(app).setAbbreviatedOptionsAllowed(true).parseArgs("--aB");
    assertTrue(app.a_b);
    assertTrue(app.aB);
    assertFalse(app.a);
    assertFalse(app.B);
}
```

NewbieOrange:
If there are options -A, -B and -AB, the ambiguity may be unavoidable. If we have document about this it should be fine.

The current impl will pass the following test:
```java
    @Test
    public void testPOSIXClusterOptionsWithAbbrevOptions() {
        @Command
        class App {
            @Option(names = "-A")
            public boolean a;

            @Option(names = "-B")
            public boolean b;

            @Option(names = "-AB")
            public boolean ab;
        }

        CommandLine commandLine = new CommandLine(new App());
        commandLine.setAbbreviatedOptionsAllowed(true);

        ParseResult result = commandLine.parseArgs("-AB");
        assertFalse(result.hasMatchedOption("-A"));
        assertFalse(result.hasMatchedOption("-B"));
        assertTrue(result.hasMatchedOption("-AB"));

        result = commandLine.parseArgs("-A");
        assertTrue(result.hasMatchedOption("-A"));
        assertFalse(result.hasMatchedOption("-B"));
        assertFalse(result.hasMatchedOption("-AB"));
    }
```

NewbieOrange:
> 
> 
> I'm fine with making `AbbreviationMatcher` a private static class in `CommandLine`.
> 
> Question: does this test pass?
> 
> ```java
> @Test
> public void testAbbrevOptionsAmbiguous() {
>     @Command
>     class App {
>         @Option(names = "--a-B")
>         public boolean a_b;
> 
>         @Option(names = "--aB")
>         public boolean aB;
> 
>         @Option(names = "-a")
>         public boolean a;
> 
>         @Option(names = "-B")
>         public boolean B;
>     }
> 
>     App app = new App();
>     new CommandLine(app).setAbbreviatedOptionsAllowed(true).parseArgs("-aB");
>     assertFalse(app.a_b);
>     assertFalse(app.aB);
>     assertTrue(app.a);
>     assertTrue(app.B);
> 
>     app = new App();
>     new CommandLine(app).setAbbreviatedOptionsAllowed(true).parseArgs("--aB");
>     assertTrue(app.a_b);
>     assertTrue(app.aB);
>     assertFalse(app.a);
>     assertFalse(app.B);
> }
> ```

It does not. When parsing `--aB`, a_b is not matched.

NewbieOrange:
It will pass if the second part first assertion is `assertFalse(app.a_b)`. Which does make sense to me.

remkop:
Package private is fine. :-)

About the test, if we pass `--AB` or `--a-b` then both `a_b` and `aB` options are matched?

NewbieOrange:
> --a-b

If we pass --AB then an `UnmatchedArgumentException` is thrown: Unknown option: '--AB'

If we pass --a-b then it causes a `ParameterException`: --a-b is not unique: it matches '--a-B', '--aB'

NewbieOrange:
If we pass --aB **and** --a-B then both options are matched.

NewbieOrange:
So the cuurent impl will find exact match first, if there is exact match, then any ambiguity is not considered. Otherwise, if there is ambiguity, a `ParameterException` is thrown.

NewbieOrange:
The lastest code should have addressed most issues.

remkop:
Great, thank you! I’m not sure if I’ll be able to review today, I’ll try to get it done this weekend.

remkop:
I went ahead and merged the PR.
Thank you very much for the contribution!
I’ll add documentation in the next few days.

NewbieOrange:
Great, glad I could contribute to the picocli community! 😀

Issue #1048, opened 2020-05-15
Cause exception not printed by default execution exception handler


remkop:
When an exception occurs in the business logic, this is handled by the execution exception handler. By default, this prints the full stack trace of the exception.

Since 4.3.0, only the stack trace of the "outer" exception is printed. If that exception has any "cause" exceptions, these are incorrectly omitted.

With the work done in #484, #845 and #1008 (colors in exception handler), we no longer call `exception.printStackTrace()` but instead have a custom implementation that has the same effect but uses ANSI colors. This implementation incorrectly omits the stack trace for any nested "cause" exceptions (and their nested cause, if any).

A workaround is to register a custom execution exception handler as follows:

```java
new CommandLine(new MyApp())
        .setExecutionExceptionHandler(
            (Exception ex, CommandLine commandLine, ParseResult parseResult) -> {
                        ex.printStackTrace();
                        return commandLine.getCommandSpec().exitCodeOnExecutionException();
            })
        .execute(args);
```

jerrylususu:
Sorry for the bug introduced. Working on a permanent fix.

jerrylususu:
One easy fix would redirect all the output from ex.printStackTrace() to a string, as proposed in #845. However, doing so may harm the ability to split exception messages from stack traces.

jerrylususu:
Seems extend StringWriter to write Style.on() before and Style.off() after is one viable solution. Testing locally.

jerrylususu:
Mostly done...
![image](https://user-images.githubusercontent.com/17522475/82067105-d8212100-9702-11ea-9604-2c944e2a3430.png)

jerrylususu:
By the way, here is a strange table about the compatibility of ANSI style.

![image](https://user-images.githubusercontent.com/17522475/82113371-1a407600-9788-11ea-9adb-2cec5e88fcc0.png)

remkop:
Thanks for the help!
I just released picocli 4.3.2 with this fix.

About the terminals, there is not much we can do about that in the application. The best we can do is select colors and styles that look acceptable in most consoles. (Which I think we did.) :-)

jerrylususu:
Sorry for the bug introduced! Hope 4.3.2 will be the last patch version of 4.3, and wish everything goes well!

remkop:
No worries!
That’s software!

PR #1049, opened 2020-05-15
Fix issue #1048

jerrylususu:
Injecting ANSI characters into ColorSchemeStringWriter, instead of reimplementing `printStackTrace()`.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -2026,11 +2026,6 @@ public class CommandLine {
             this.colorScheme = colorScheme;
         }
 
-        public ColorSchemedStringWriter(int initialSize, Help.ColorScheme colorScheme) {
-            super(initialSize);
-            this.colorScheme = colorScheme;
-        }
-
         @Override
         public void write(String str, int off, int len) {
             List<IStyle> styles = str.startsWith("\t") ? colorScheme.stackTraceStyles() : colorScheme.errorStyles();
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -2009,7 +2009,6 @@ public class CommandLine {
      */
     private static String exceptionToColorString(Exception ex, Help.ColorScheme existingColorScheme){
         Help.ColorScheme colorScheme = new Help.ColorScheme.Builder(existingColorScheme).applySystemProperties().build();
-        String newLine = System.getProperty("line.separator");
         StringWriter stringWriter = new ColorSchemedStringWriter(colorScheme);
         ex.printStackTrace(new PrintWriter(stringWriter));
         return stringWriter.toString();
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -2007,7 +2007,7 @@ public class CommandLine {
      * @param existingColorScheme the {@code ColorScheme} to use
      * @return converted and colored {@code String}
      */
-    private static String exceptionToColorString(Exception ex, Help.ColorScheme existingColorScheme){
+    private static String exceptionToColorString(Exception ex, Help.ColorScheme existingColorScheme) {
         Help.ColorScheme colorScheme = new Help.ColorScheme.Builder(existingColorScheme).applySystemProperties().build();
         StringWriter stringWriter = new ColorSchemedStringWriter(colorScheme);
         ex.printStackTrace(new PrintWriter(stringWriter));
@@ -2033,21 +2033,9 @@ public class CommandLine {
 
         @Override
         public void write(String str, int off, int len) {
-            if(str.startsWith("\t")){
-                // that's stacktrace
-                super.write(Style.on(colorScheme.stackTraceStyles().toArray(new IStyle[0])));
-            } else {
-                super.write(Style.on(colorScheme.errorStyles().toArray(new IStyle[0])));
-            }
-            super.write(str, off, len);
-            if(str.startsWith("\t")){
-                super.write(Style.off(colorScheme.stackTraceStyles().toArray(new IStyle[0])));
-            } else {
-                super.write(Style.off(reverseArray(colorScheme.errorStyles().toArray(new IStyle[0]))));
-            }
-            super.write(Style.reset.off());
+            List<IStyle> styles = str.startsWith("\t") ? colorScheme.stackTraceStyles() : colorScheme.errorStyles();
+            super.write(colorScheme.apply(str.substring(off, len), styles).toString());
         }
-
     }
 
     private <T> T enrichForBackwardsCompatibility(T obj) {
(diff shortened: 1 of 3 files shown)


jerrylususu:
The root cause of issue #1048 is lack of testing on this situation. I added a test on the nested exceptions in `ColorSchemeTest`, but in retrospective maybe putting it into `ExecuteTest` is a better option. Also, current modifications to make all test past feels strange to me, as sometimes default parameter of `ColorScheme` needs to be hardcoded into tests in `ExecuteTest` , which indicates high coupling. Perhaps some more thoughts on the tests is required.

src/test/java/picocli/ExecuteTest.java
```diff
@@ -480,15 +480,17 @@ public int getExitCode(Throwable t) {
         int exitCode = cmd.execute();
         assertEquals(1234, exitCode);
 
-        assertThat(this.systemErrRule.getLog(), startsWith("java.io.IOException: error"));
+        assertThat(this.systemErrRule.getLog(), startsWith(Help.Ansi.ON.new Text(format(
+                "@|fg(red),bold java.io.IOException: error|@"
+        )).toString()));
```
remkop, line 485: Why does the stack trace contain ANSI escape codes, when the ExecuteTest switches off ANSI colors explicitly?

The below test rule defined in ExecuteTest ensures that system property `picocli.ansi` is set to `false` before each test. That should result in no ANSI escape codes being emitted. But you have changed the test expectations to expect ANSI escaped codes, and these tests pass... This is strange. Why is this?


```java
public class ExecuteTest {
    @Rule
    public final ProvideSystemProperty ansiOFF = new ProvideSystemProperty("picocli.ansi", "false");
```

src/main/java/picocli/CommandLine.java
```diff
@@ -2009,13 +2009,45 @@ private static int handleUnhandled(Exception ex, CommandLine cmd, int defaultExi
...
+
+        public ColorSchemedStringWriter(int initialSize, Help.ColorScheme colorScheme) {
+            super(initialSize);
+            this.colorScheme = colorScheme;
         }
-        return stringBuilder.toString();
+
+        @Override
+        public void write(String str, int off, int len) {
+            if(str.startsWith("\t")){
```
remkop, line 2036: Code style comment:
There is a space missing between the `if` and the opening brace `(`, and there is also a space missing between the closing brace `)` and the opening curly `{`. 

This is the case for all "if" statements, but not for "for" statements or "else" statements. Please put spaces to adhere to the style of the existing code. :-)

src/main/java/picocli/CommandLine.java
```diff
@@ -2009,13 +2009,45 @@ private static int handleUnhandled(Exception ex, CommandLine cmd, int defaultExi
...
+                // that's stacktrace
+                super.write(Style.on(colorScheme.stackTraceStyles().toArray(new IStyle[0])));
```
remkop, line 2038: It is not a good idea to unconditionally emit ANSI escape codes.
The overall idea is a good one, but it is better to construct `Ansi.Text` objects for the existing `Ansi` value.
The `ColorScheme.apply` method is useful here:

```java
@Override
public void write(String str, int off, int len) {
    List<IStyle> styles = str.startsWith("\t") ? colorScheme.stackTraceStyles() : colorScheme.errorStyles();
    super.write(colorScheme.apply(str.substring(off, len), styles).toString());
}
```

src/test/java/picocli/ExecuteTest.java
jerrylususu: Sorry, don't notice this... I'll revert my changes to this file.

src/main/java/picocli/CommandLine.java
jerrylususu: Thanks for the advice! That seems much better!

src/test/java/picocli/ExecuteTest.java
jerrylususu: It seems my original implementation don't check if ANSI is enabled or disabled. Current version uses Ansi.Text.toString, which will check it now.

src/main/java/picocli/CommandLine.java
jerrylususu: Sorry for that. I will run the code formatter in IDEA before committing.

jerrylususu:
Changed according to the comment. Now uses `Ansi.Text.toString` to only output ANSI control characters when ANSI is enabled. Changes to the `ExecuteTest` are reverted.

src/test/java/picocli/ExecuteTest.java
jerrylususu: Changes to `ExecuteTest` reverted.

src/main/java/picocli/CommandLine.java
```diff
@@ -2007,15 +2007,35 @@ private static int handleUnhandled(Exception ex, CommandLine cmd, int defaultExi
...
+     */
+    static class ColorSchemedStringWriter extends StringWriter {
+
+        Help.ColorScheme colorScheme;
+
+        public ColorSchemedStringWriter(Help.ColorScheme colorScheme) {
+            this.colorScheme = colorScheme;
+        }
+
+        public ColorSchemedStringWriter(int initialSize, Help.ColorScheme colorScheme) {
```
remkop, line 2029: Perhaps we can omit this constructor.

src/test/java/picocli/ColorSchemeTest.java
```diff
@@ -149,6 +154,102 @@ public void testColorScheme_stringAnsiOn() {
         assertEquals("Hello #UNDERLINE ME# please.", colorScheme.string("Hello @|underline UNDERLINE ME|@ please."));
     }
 
+    @Test
```
remkop, line 157: Can you add tests with Ansi.ON?
to verify that parts of the message have different colors and can be controlled by the ColorScheme?
Maybe test with a custom colorscheme also?

src/main/java/picocli/CommandLine.java
jerrylususu: The `initialSize` one? Seems not used by any code. I think it can be safely omitted.

src/test/java/picocli/ColorSchemeTest.java
jerrylususu: Sure.

jerrylususu: Could you provide such an example of invoking with `Ansi.ON`?

remkop: ok.
I added `ExecuteTest::testIssue1048CauseException`. This fails with current master but passes when your PR is applied.

Note the commented out `//System.setProperty("picocli.ansi", "true");` before `CommandLine::execute` is called. 

That is what I mean: let's add 2 more tests, both with `-Dpicocli.ansi=true`, one using the standard color scheme, and one with a customized color scheme. 

In the verification part of the test you would assert that the output contains the expected ANSI escape codes (similar to some tests you had previously): we want to verify that different styles are applied to the stack trace part and to the error part.

remkop: I pushed one more commit to make it easier to add tests.

jerrylususu: Thanks! Now I understand. I'm working on it.

jerrylususu:
Added some new test as requested.

Issue #1058, opened 2020-05-19
Synopsis of man page generator: incorrect asciidoctor call


deining:
**Synopsis of man page generator:**

```
Usage: gen-manpage [-fhVv] [--[no-]exit] [-c=<factoryClass>] [-d=<outdir>] [-t=<template-dir>] [@<filename>...] <classes>...
...
...
Converting to Man Page Format

Use the `asciidoctor` tool to convert the generated AsciiDoc files to man pages in roff format:

`asciidoctor --backend=manpage --source-dir=SOURCE_DIR --destination-dir=DESTINATION`
```
The asciidoctor command is missing the required parameter `FILE...`, should be: 

`asciidoctor --backend=manpage --source-dir=SOURCE_DIR --destination-dir=DESTINATION *.adoc`

I submitted a PR meanwhile (#1059).

**Additional question:**

I do have a localized, picocli based application. Can I use the man page generator to generate *.adoc files in my various language variants?

remkop:
Thank you for pointing out the problem in the docs and for the PR to fix it!

> I do have a localized, picocli based application. Can I use the man page generator to generate *.adoc files in my various language variants?

I haven't actually tried this, but yes, that should be possible. When you invoke the `ManPageGenerator` command, if you set the target locale by specifying the `user.language`, `user.country`, and `user.variant system` properties, the generated usage help should use the resource bundle based on the user locale.

(We should probably add a section about this to the user manual for generating man pages.)

deining:
> I haven't actually tried this, but yes, that should be possible. When you invoke the `ManPageGenerator` command, if you set the target locale by specifying the `user.language`, `user.country`, and `user.variant system` properties, the generated usage help should use the resource bundle based on the user locale.

I can confirm that this works, indeed.
 
> (We should probably add a section about this to the documentation for generating man pages.)

I try to author a PR later on.

**Additional note:**

Man page generation is a great feature. Currently, it is somehow hidden, however. Can't we make it more accessible by giving an picocli based application the ability to generate its own man page on demand (via subcommand). Something like:

```
@Command(name = "mycommand", subcommands = GenerateManPage.class)
public class MyApp implements Runnable {
...
}
```

Can this be implemented? Or does this feature exist already? Would be a great addition IMHO.

remkop:
>  Can't we make it more accessible by giving an picocli based application the ability to generate its own man page on demand (via subcommand).

I see. I hadn't thought of that but that is a nice idea.
[...]

PR #1059, opened 2020-05-19
Man page generator: fix incorrect asciidoctor call in synopsis

deining:
Fix for #1058.


remkop:
Merged into master.

Issue #1063, opened 2020-05-20
ManPageGenerator does not ignore "hidden" options


bdemers:
I was expecting the ManPageGenerator to ignore "hidden" options the same way that the standard `--help` renderer ignores them, for example:

```java
@Option(names = "--color", hidden = true)
public void setColor(ColorOptions color) { 
  // ...
}
```

This does NOT render when `--help` is used, but it IS rendered to the manpage doc:

```txt
--color=<color>
```

bdemers:
NOTE: hidden subcommands are rendered as well

remkop:
@bdemers Thank you for reporting this!
This is an oversight and a bug.

remkop:
Will you be able to provide a pull request for this?

bdemers:
I think I've got it, doing a bit more testing and I'll send a PR

remkop:
Merged the PR. Thanks for the contribution!

There may be an edge case where options in an argument group (with header) are not hidden, while all other options are hidden; would be good to have a test case for that... 
If you can take a look at this that would be great!

PR #1064, opened 2020-05-21
Excludes hidden options, params, and subcommands from man page generation

bdemers:
Fixes: #1063

### DIFF
--- a/picocli-codegen/src/main/java/picocli/codegen/docgen/manpage/ManPageGenerator.java
+++ b/picocli-codegen/src/main/java/picocli/codegen/docgen/manpage/ManPageGenerator.java
@@ -381,6 +381,14 @@ public class ManPageGenerator {
         IParameterRenderer parameterRenderer = spec.commandLine().getHelp().createDefaultParameterRenderer();
 
         List<OptionSpec> options = new ArrayList<OptionSpec>(spec.options()); // options are stored in order of declaration
+
+        // remove hidden options
+        for (Iterator<OptionSpec> iter = options.iterator(); iter.hasNext();) {
+            if (iter.next().hidden()) {
+                iter.remove();
+            }
+        }
+
         List<ArgGroupSpec> groups = optionListGroups(spec);
         for (ArgGroupSpec group : groups) { options.removeAll(group.options()); }
 
@@ -431,14 +439,12 @@ public class ManPageGenerator {
     }
 
     private static void writeOption(PrintWriter pw, IOptionRenderer optionRenderer, IParamLabelRenderer paramLabelRenderer, OptionSpec option) {
-        if (!option.hidden()) {
-            pw.println();
-            Text[][] rows = optionRenderer.render(option, paramLabelRenderer, COLOR_SCHEME);
-            pw.printf("%s::%n", join(", ", rows[0][1], rows[0][3]));
-            pw.printf("  %s%n", rows[0][4]);
-            for (int i = 1; i < rows.length; i++) {
-                pw.printf("+%n%s%n", rows[i][4]);
-            }
+        pw.println();
+        Text[][] rows = optionRenderer.render(option, paramLabelRenderer, COLOR_SCHEME);
+        pw.printf("%s::%n", join(", ", rows[0][1], rows[0][3]));
+        pw.printf("  %s%n", rows[0][4]);
+        for (int i = 1; i < rows.length; i++) {
+            pw.printf("+%n%s%n", rows[i][4]);
         }
     }
 
--- a/picocli-codegen/src/main/java/picocli/codegen/docgen/manpage/ManPageGenerator.java
+++ b/picocli-codegen/src/main/java/picocli/codegen/docgen/manpage/ManPageGenerator.java
@@ -461,7 +461,7 @@ public class ManPageGenerator {
     }
 
     static void genPositionalArgs(PrintWriter pw, CommandSpec spec) {
-        List<PositionalParamSpec> positionals = new ArrayList<>(spec.positionalParameters());
+        List<PositionalParamSpec> positionals = new ArrayList<PositionalParamSpec>(spec.positionalParameters());
         // remove hidden params
         for (Iterator<PositionalParamSpec> iter = positionals.iterator(); iter.hasNext();) {
             if (iter.next().hidden()) { iter.remove(); }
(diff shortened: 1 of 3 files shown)


src/main/java/picocli/CommandLine.java
```diff
@@ -6693,6 +6705,7 @@ void initDefaultValueProvider(Class<? extends IDefaultValueProvider> value, IFac
             void updateHelpCommand(boolean value)           { if (isNonDefault(value, DEFAULT_IS_HELP_COMMAND))             {isHelpCommand = value;} }
             void updateSubcommandsRepeatable(boolean value) { if (isNonDefault(value, DEFAULT_SUBCOMMANDS_REPEATABLE))      {subcommandsRepeatable = value;} }
             void updateAddMethodSubcommands(boolean value)  { if (isNonDefault(value, DEFAULT_IS_ADD_METHOD_SUBCOMMANDS))   {isAddMethodSubcommands = value;} }
+            void updateHidden(boolean value)                { if (isNonDefault(value, DEFAULT_HIDDEN)) { isHidden = value; } }
```
bdemers, line 6708: @remkop I didn't see another way of access the `hidden` prop, but I'm probably missing something as it is used for the help text output.
If you have any alternative idea let me know, and I could remove this method (and revert the MINOR version bump)

remkop: @bdemers The `@Command(hidden)` value is captured as a property of the command's `UsageHelpMessageSpec`. I admit this is not easy to find. Perhaps the javadoc for `picocli.CommandLine.Command#hidden` should have a `@see picocli.CommandLine.Model.UsageMessageSpec#hidden()` reference to make this a bit easier to find. Can you do that as part of the PR?

The `Help` class has its own map of subcommands, and when this map is built up in `CommandLine.Help#addAllSubcommands`, hidden subcommands are excluded. This means that the logic for `Help` is simpler, because it does not need to care about hidden subcommands any more, but can be a bit puzzling if you miss this. This should be mentioned in the javadoc for `Help#addAllSubcommands` and the javadoc for `Help#subcommands`. Can you include that in the PR?

Good finds, thanks! 👍

picocli-codegen/src/main/java/picocli/codegen/docgen/manpage/ManPageGenerator.java
```diff
@@ -441,12 +449,14 @@ private static void writeOption(PrintWriter pw, IOptionRenderer optionRenderer,
     }
 
     private static void writePositional(PrintWriter pw, PositionalParamSpec positional, IParameterRenderer parameterRenderer, IParamLabelRenderer paramLabelRenderer) {
-        pw.println();
-        Text[][] rows = parameterRenderer.render(positional, paramLabelRenderer, COLOR_SCHEME);
-        pw.printf("%s::%n", join(", ", rows[0][1], rows[0][3]));
-        pw.printf("  %s%n", rows[0][4]);
-        for (int i = 1; i < rows.length; i++) {
-            pw.printf("+%n%s%n", rows[i][4]);
+        if (!positional.hidden()) {
```
remkop, line 452: Should we have a similar if statement in `writeOption` : `if (!option.hidden()) ...`, 
or should we exclude hidden `PositionalParamSpec` before the `writePositional` method is called?
I would like it if we can do this in a consistent manner for options, positionals and commands.

src/main/java/picocli/CommandLine.java
remkop: Since we can use `picocli.CommandLine.Model.UsageMessageSpec#hidden`, all changes to `CommandLine` can be reverted.

remkop:
FYI: This issue may be relevant: https://github.com/remkop/picocli/issues/1071

bdemers:
Still one more comment to address, but I reverted the new method and version bump

picocli-codegen/src/main/java/picocli/codegen/docgen/manpage/ManPageGenerator.java
bdemers: +1, keeping the check in `writeOption`/`writePositional` is easier, as `writePositional` is called from multiple places.

However, the check for the subcommands gets a little trickier, the subcommands need to be removed before processing to avoid the tag/"Commands" heading.

Thoughts?

remkop:
Thanks for the updates! Please let me know if this is ready for review.

bdemers:
Hey @remkop it's ready. though I did have a question/comment: https://github.com/remkop/picocli/pull/1064/files#r430546512

picocli-codegen/src/main/java/picocli/codegen/docgen/manpage/ManPageGenerator.java
remkop: After thinking about this some more, I think your original approach to filter out hidden args first is better.

The reason is the same as for subcommands: if all options or positional parameters are hidden, we don't want to include the tag and heading in the output. (see also https://github.com/remkop/picocli/issues/1071)

bdemers:
@remkop good call, I updated the test to make sure the Arguments header is not included when all of the args are hidden

picocli-codegen/src/main/java/picocli/codegen/docgen/manpage/ManPageGenerator.java
```diff
@@ -381,6 +381,14 @@ static void genOptions(PrintWriter pw, CommandSpec spec) {
         IParameterRenderer parameterRenderer = spec.commandLine().getHelp().createDefaultParameterRenderer();
 
         List<OptionSpec> options = new ArrayList<OptionSpec>(spec.options()); // options are stored in order of declaration
+
+        // remove hidden options
+        for (Iterator<OptionSpec> iter = options.iterator(); iter.hasNext();) {
+            if (iter.next().hidden()) {
+                iter.remove();
+            }
+        }
```
remkop, line 390: I believe this is the right approach, but we should do this at the top of this method, before rendering the `== Options` header.

bdemers: Ahh, I missed that one, i just pushed an update

remkop:
It may be good to have a test for the case where all options are hidden.

remkop:
I went ahead and merged this into master.

PR #1070, opened 2020-05-25
Code cleanup

NewbieOrange:
This PR removes redundant modifiers and initializations, unused variables, incorrect javadoc references, etc.

The `static` modifiers in inner interfaces are removed since they are static by default, and removing them should not break binary compatibility (the md5sums are the same). Let me know if those are preferred to be kept.

### DIFF
--- a/src/main/java/picocli/AutoComplete.java
+++ b/src/main/java/picocli/AutoComplete.java
@@ -235,7 +235,7 @@ public class AutoComplete {
         }
     }
 
-    private static interface Function<T, V> {
+    private interface Function<T, V> {
         V apply(T t);
     }
 
@@ -264,7 +264,7 @@ public class AutoComplete {
         public String apply(CharSequence value) { return value.toString(); }
     }
 
-    private static interface Predicate<T> {
+    private interface Predicate<T> {
         boolean test(T t);
     }
     private static class BooleanArgFilter implements Predicate<ArgSpec> {
@@ -720,14 +720,12 @@ public class AutoComplete {
             return "";
         }
 
-        StringBuilder buff = new StringBuilder(1024);
-        buff.append("\n");
-        buff.append("  compopt +o default\n");
-        buff.append("\n");
-        buff.append("  case ${prev_word} in\n");
-        buff.append(optionsCases);
-        buff.append("  esac\n");
-        return buff.toString();
+        return "\n"
+                + "  compopt +o default\n"
+                + "\n"
+                + "  case ${prev_word} in\n"
+                + optionsCases
+                + "  esac\n";
     }
 
     private static String generateOptionsCases(List<OptionSpec> argOptionFields, String indent, String currWord) {
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -7993,7 +7993,7 @@ public class CommandLine {
             private List<String> stringValues = new ArrayList<String>();
             private List<String> originalStringValues = new ArrayList<String>();
             protected String toString;
-            private List<Object> typedValues = new ArrayList<Object>();
+            private final List<Object> typedValues = new ArrayList<Object>();
             Map<Integer, Object> typedValueAtPosition = new TreeMap<Integer, Object>();
 
             /** Constructs a new {@code ArgSpec}. */
@@ -8976,13 +8976,13 @@ public class CommandLine {
         public static class OptionSpec extends ArgSpec implements IOrdered {
             public static final String DEFAULT_FALLBACK_VALUE = "";
             static final int DEFAULT_ORDER = -1;
-            private String[] names;
-            private boolean help;
-            private boolean usageHelp;
-            private boolean versionHelp;
-            private boolean negatable;
-            private String fallbackValue;
-            private int order;
+            private final String[] names;
+            private final boolean help;
+            private final boolean usageHelp;
+            private final boolean versionHelp;
+            private final boolean negatable;
+            private final String fallbackValue;
+            private final int order;
 
             public static OptionSpec.Builder builder(String name, String... names) {
                 String[] copy = new String[Assert.notNull(names, "names").length + 1];
@@ -9255,7 +9255,7 @@ public class CommandLine {
         public static class PositionalParamSpec extends ArgSpec {
             private Range index;
             private Range capacity;
-            private Range builderCapacity;
+            private final Range builderCapacity;
 
             /** Ensures all attributes of this {@code PositionalParamSpec} have a valid value; throws an {@link InitializationException} if this cannot be achieved. */
             private PositionalParamSpec(Builder builder) {
@@ -9788,12 +9788,12 @@ public class CommandLine {
                 private Range multiplicity = Range.valueOf("0..1");
                 private boolean validate   = true;
                 private int order          = DEFAULT_ORDER;
-                private List<ArgSpec> args = new ArrayList<ArgSpec>();
-                private List<ArgGroupSpec> subgroups = new ArrayList<ArgGroupSpec>();
+                private final List<ArgSpec> args = new ArrayList<ArgSpec>();
+                private final List<ArgGroupSpec> subgroups = new ArrayList<ArgGroupSpec>();
 
                 // for topological sorting; private only
                 private Boolean topologicalSortDone;
-                private List<Builder> compositesReferencingMe = new ArrayList<Builder>();
+                private final List<Builder> compositesReferencingMe = new ArrayList<Builder>();
 
                 Builder() { }
                 Builder(IAnnotatedElement source) {
@@ -11455,9 +11455,9 @@ public class CommandLine {
             private boolean usageHelpRequested;
             private boolean versionHelpRequested;
             boolean isInitializingDefaultValues;
-            private List<Exception> errors = new ArrayList<Exception>(1);
+            private final List<Exception> errors = new ArrayList<Exception>(1);
             private List<Object> nowProcessing;
-            private GroupMatchContainer groupMatchContainer = new GroupMatchContainer(null, null);
+            private final GroupMatchContainer groupMatchContainer = new GroupMatchContainer(null, null);
 
             private Builder(CommandSpec spec) { commandSpec = Assert.notNull(spec, "commandSpec"); }
             /** Creates and returns a new {@code ParseResult} instance for this builder's configuration. */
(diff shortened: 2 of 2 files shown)


NewbieOrange:
Also there are a lot of `@SuppressWarnings("deprecation")` removed since we're at language level 1.5 and should not trigger any warnings.

src/main/java/picocli/CommandLine.java
```diff
@@ -7732,7 +7724,6 @@ void initFromMixin(UsageMessageSpec mixin, CommandSpec commandSpec) {
                 if (initializable(sortOptions, mixin.sortOptions(), DEFAULT_SORT_OPTIONS))                             {sortOptions = mixin.sortOptions();}
                 if (initializable(synopsisHeading, mixin.synopsisHeading(), DEFAULT_SYNOPSIS_HEADING))                 {synopsisHeading = mixin.synopsisHeading();}
                 if (initializable(synopsisSubcommandLabel, mixin.synopsisSubcommandLabel(), DEFAULT_SYNOPSIS_SUBCOMMANDS)) {synopsisHeading = mixin.synopsisHeading();}
-                if (initializable(width, mixin.width(), DEFAULT_USAGE_WIDTH))                                          {width = mixin.width();}
```
remkop, line 7735: Why is this removed?

src/main/java/picocli/CommandLine.java
```diff
@@ -5138,9 +5133,8 @@ public Help create(CommandSpec commandSpec, Help.ColorScheme colorScheme) {
                 return cls.cast(new LinkedHashMap<Object, Object>());
             }
             try {
-                @SuppressWarnings("deprecation") // Class.newInstance is deprecated in Java 9
-                T result = cls.newInstance();
-                return result;
```
remkop, line 5143: This `@SuppressWarnings("deprecation")` annotation suppresses the warning one gets when building the project with Java 9. I think it is better to keep it.

src/main/java/picocli/AutoComplete.java
```diff
@@ -235,7 +235,7 @@ public void run() {
         }
     }
 
-    private static interface Function<T, V> {
```
remkop, line 238: No problem, I am okay with removing `static` for inner interfaces.

src/main/java/picocli/CommandLine.java
```diff
@@ -1470,7 +1470,7 @@ public void clearExecutionResults() {
      * @see RunAll
      * @deprecated Use {@link IExecutionStrategy} instead.
      * @since 2.0 */
-    @Deprecated public static interface IParseResultHandler {
+    @Deprecated public interface IParseResultHandler {
```
remkop, line 1473: Again, no problem removing `static` for inner interfaces.

src/main/java/picocli/CommandLine.java
```diff
@@ -1758,7 +1758,7 @@ protected R throwOrExit(ExecutionException ex) {
      * @since 2.0 */
     @Deprecated public static class DefaultExceptionHandler<R> extends AbstractHandler<R, DefaultExceptionHandler<R>> implements IExceptionHandler, IExceptionHandler2<R> {
         public List<Object> handleException(ParameterException ex, PrintStream out, Help.Ansi ansi, String... args) {
-            internalHandleParseException(ex, new PrintWriter(out, true), Help.defaultColorScheme(ansi)); return Collections.<Object>emptyList(); }
```
remkop, line 1761: I think I introduced this to get rid of a "raw collections" warning when building with Java 8 or higher... Probably better to keep the generified form with `<Object>`.

src/main/java/picocli/CommandLine.java
```diff
@@ -2036,7 +2035,6 @@ public int execute(String... args) {
             }
         } catch (ExecutionException ex) {
             try {
-                @SuppressWarnings("unchecked")
                 Exception cause = ex.getCause() instanceof Exception ? (Exception) ex.getCause() : ex;
```
remkop, line 2038: `Exception.getCause` returns a `Throwable`. We never catch Throwable so we can safely downcast it to Exception, but I believe we do need to keep this annotation to suppress the compiler warning.

src/main/java/picocli/CommandLine.java
```diff
@@ -2140,7 +2138,7 @@ private int resolveExitCode(int exitCodeOnSuccess, R executionResult, List<IExit
             for (IExitCodeGenerator generator : exitCodeGenerators) {
                 try {
                     int exitCode = generator.getExitCode();
-                    if ((exitCode > 0 && exitCode > result) || (exitCode < 0 && result <= 0 && exitCode < result)) {
+                    if ((exitCode > 0 && exitCode > result) || (exitCode < result && result <= 0)) {
```
remkop, line 2141: Nice simplification!

src/main/java/picocli/CommandLine.java
```diff
@@ -2217,7 +2214,7 @@ private int resolveExitCode(int exitCodeOnSuccess, R executionResult, List<IExit
 
         protected List<IExitCodeGenerator> extractExitCodeGenerators(ParseResult parseResult) {
             if (parseResult.commandSpec().userObject() instanceof IExitCodeGenerator) {
-                return Arrays.asList((IExitCodeGenerator) parseResult.commandSpec().userObject());
+                return Collections.singletonList((IExitCodeGenerator) parseResult.commandSpec().userObject());
```
remkop, line 2217: Nice. I overlooked that this API exists.

src/main/java/picocli/CommandLine.java
```diff
@@ -8431,7 +8422,7 @@ private String scopeString() {
                 StringBuilder temp = new StringBuilder();
                 StringBuilder current = splittable;
                 boolean escaping = false, inQuote = false;
-                for (int ch = 0, i = 0; i < value.length(); i += Character.charCount(ch)) {
+                for (int ch, i = 0; i < value.length(); i += Character.charCount(ch)) {
```
remkop, line 8425: personally I prefer to explicitly initialize all variables, but this is a minor quibble.

src/main/java/picocli/CommandLine.java
```diff
@@ -8473,7 +8464,7 @@ private String scopeString() {
             private static String restoreQuotedValues(String part, Queue<String> quotedValues, ParserSpec parser) {
                 StringBuilder result = new StringBuilder();
                 boolean escaping = false, inQuote = false, skip = false;
-                for (int ch = 0, i = 0; i < part.length(); i += Character.charCount(ch)) {
+                for (int ch, i = 0; i < part.length(); i += Character.charCount(ch)) {
```
remkop, line 8467: Same as above.

src/main/java/picocli/CommandLine.java
```diff
@@ -10038,7 +10025,6 @@ public MethodParam(Method method, int paramIndex) {
             }
             @Override public Annotation[] getDeclaredAnnotations() { return method.getParameterAnnotations()[paramIndex]; }
             @Override public void setAccessible(boolean flag) throws SecurityException { method.setAccessible(flag); }
-            @SuppressWarnings("deprecation")
```
remkop, line 10041: `Method.isAccessible` will give deprecation warnings when building on Java 9+.
I believe we want to suppress this.
Try this:

```
%JAVA_11_HOME%\bin\javac -Xlint:all -d c:\Temp picocli\CommandLine.java
```

src/main/java/picocli/CommandLine.java
```diff
@@ -14943,7 +14921,7 @@ public IParamLabelRenderer createDefaultParamLabelRenderer() {
          *   OptionSpec#description()} array, and these rows look like {@code {"", "", "", option.description()[i]}}.</p>
          */
         static class DefaultOptionRenderer implements IOptionRenderer {
-            private String requiredMarker = " ";
+            private String requiredMarker;
             private boolean showDefaultValues;
```
remkop, line 14925: Can we make `requiredMarker` and `showDefaultValues` `final` fields?

src/main/java/picocli/CommandLine.java
```diff
@@ -15063,7 +15041,7 @@ private Text createLongOptionText(OptionSpec option, IParamLabelRenderer rendere
          *   PositionalParamSpec#description()} array, and these rows look like {@code {"", "", "", param.description()[i]}}.</p>
          */
         static class DefaultParameterRenderer implements IParameterRenderer {
-            private String requiredMarker = " ";
+            private String requiredMarker;
             private boolean showDefaultValues;
```
remkop, line 15045: Can we make `requiredMarker` and `showDefaultValues` `final` fields?

src/main/java/picocli/CommandLine.java
NewbieOrange: This is never reached. initializable replaced the null varaible with default value, but an int is never null.

NewbieOrange: If we want to keep this, we will need to introduce `initializable`s with each primitive type.

NewbieOrange:
I have checked the other removed `@SuppressWarnings` does not introduce any new warning when compiled with OpenJDK 11.0.7 and -Xlint:all.

src/main/java/picocli/CommandLine.java
NewbieOrange: Yes, I added a commit with some other private fields being made final.

NewbieOrange:
The added final modifiers seem to break the codegen graalvm-aot tests, since the generated fields have an entra `"allowWrite" : true`.

NewbieOrange:
On a side note, I think we should also remove the broken kloudtek logo (HTTPS cert expired) and machine-translated article (从Commons CLI迁移到Picocli) from README.

remkop:
> I have checked the other removed @SuppressWarnings does not introduce any new warning when compiled with OpenJDK 11.0.7 and -Xlint:all.

Thank you!

> The added final modifiers seem to break the codegen graalvm-aot tests, since the generated fields have an entra `"allowWrite" : true`.
> 
> Edit: The field should not be made final, reverted.

Interesting! I’ll take a look tomorrow. 

> On a side note, I think we should also remove the broken kloudtek logo (HTTPS cert expired) and machine-translated article (从Commons CLI迁移到Picocli) from README.

Is the translation that bad? :-)
Even if it is, does it reflect badly on picocli? 
I was thinking it’s better to have more links than fewer, to spread awareness of the library.

NewbieOrange:
> > The added final modifiers seem to break the codegen graalvm-aot tests, since the generated fields have an entra `"allowWrite" : true`.
> > Edit: The field should not be made final, reverted.
> 
> Interesting! I’ll take a look tomorrow.
> 
This is due to a private field being made final, but it is accessed by reflection from annotations. I have removed the final modifier that caused the issue.

> > On a side note, I think we should also remove the broken kloudtek logo (HTTPS cert expired) and machine-translated article (从Commons CLI迁移到Picocli) from README.
> 
> Is the translation that bad? :-)
> Even if it is, does it reflect badly on picocli?
> I was thinking it’s better to have more links than fewer, to spread awareness of the library.

The translation is simply google translate I think. It is not unreadable tho, but some code snippets are being 'translated', and it is also not the original post. The original post is here (the content is the same): https://blog.csdn.net/Tybyqi/article/details/85787550

src/main/java/picocli/CommandLine.java
remkop: Please revert this one; the warning that this condition is always `false` should go away with the latest master.

NewbieOrange:
>Can these be final too?

Made the fields final.

> When I tried compiling on java 11 with -Xlint:all, I saw deprecation warnings for method.isAccessible.
You did not get this?

No.. probably due to difference in build options (different distro, etc), I have added the @SupressWarnings annotation anyway.

> Please revert this one; the warning that this condition is always false should go away with the latest master.

Done.

remkop:
Merged into master.
Thank you for the contribution! Much appreciated!

Issue #1074, opened 2020-05-26
Get expanded @-file arguments?


mashlol:
Is it possible to get the parsed, expanded @-file arguments after parsing is complete? In `ParseResult`, I see there is a `tentativeMatch` which does have the correct arguments parsed out, but it doesn't seem to be exposed externally. The `originalArgs` object still has the un-expanded @-file arguments like `@/path/to/file.args`

Basically given `binary --arg @/path/to/file --other_arg @/path/to/second/file`, is it possible to get `binary --arg expanded_file --other_arg expanded_file`? Either as a raw string, or a list of parsed tokens (preferred).

The use-case is to be able to log the actual arguments even if you invoke the binary with a file, and support replayability with a copy/pastable repro command.

remkop:
Hi, with the current version of picocli this feature does not exist.
We could add a method `ParseResult.expandedArgs()` that returns the argument list that was actually processed by the parser (so after `@file` expansion).

Will you be able to provide a pull request for this?

mashlol:
Sure, I can submit a pull request for this.

remkop:
This is a nice addition!
Thank you for the PR! Merged and will be included in the next release.

PR #1075, opened 2020-05-26
Expose the expanded argument list from ParseResult

mashlol:
Written to solve #1074 , this exposes the expanded argument list that picocli keeps internally. I had to reverse the list since it was stored as a stack, which I believe means we're copying the full argument list twice into memory. Ideally we'd use some sort of reversed view and not duplicate it, but since we don't have guava or anything like that it's slightly ugly imo. Happy to update with that though.


remkop:
Merged into master.

PR #1079, opened 2020-05-30
fix#1077

Linyer-qwq:
Follow the way `writeOption` do, add a function `writeEndOfOptions`.

### DIFF
--- a/picocli-codegen/src/main/java/picocli/codegen/docgen/manpage/ManPageGenerator.java
+++ b/picocli-codegen/src/main/java/picocli/codegen/docgen/manpage/ManPageGenerator.java
@@ -387,7 +387,7 @@ public class ManPageGenerator {
         List<ArgGroupSpec> groups = optionListGroups(spec);
         for (ArgGroupSpec group : groups) { options.removeAll(group.options()); }
 
-        if (options.isEmpty()) {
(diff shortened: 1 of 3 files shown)


picocli-codegen/src/main/java/picocli/codegen/docgen/manpage/ManPageGenerator.java
```diff
@@ -403,6 +403,12 @@ static void genOptions(PrintWriter pw, CommandSpec spec) {
             writeOption(pw, optionRenderer, paramLabelRenderer, option);
         }
 
+        if (spec.usageMessage().showEndOfOptionsDelimiterInUsageHelp()) {
```
remkop, line 406: We want to show the Options section with the end-of-options delimiter in the man page if the command has no options, but `@Command(showEndOfOptionsDelimiterInUsageHelp = true)`.
 
Higher up in this method, there is this check:

```java
        if (options.isEmpty()) {
            return;
        }
```

We should change that to:

```java
        if (options.isEmpty() && !spec.usageMessage().showEndOfOptionsDelimiterInUsageHelp()) {
            return;
        }
```

It would be great if you can also add a test for this case.

Linyer-qwq: Thanks for your guidance, I have submitted the changes. 😀

remkop:
Merged into master.
Thank you for your contribution!

PR #1100, opened 2020-06-05
fix factory used for repeatable subcommands

protogenes:
the factory should be passed when creating a CommandSpec copy and not degrade to the DefaultFactory


remkop:
Merged into master.

Issue #990, opened 2020-04-13
Options in subcommands are not reset to their initial value


remkop:
When a single `CommandLine` instance is reused multiple times, picocli is supposed to reset the values of all options and positional parameters to their initial value, and then to their default value (if a default value was defined in the annotations).

# Scope

With picocli 4.2.0, subcommand options are not cleared between invocations. This impacts applications that **reuse a `CommandLine` instance in multiple invocations**. If an option was specified in the first invocation and omitted in the second invocation, it will incorrectly retain the value from the first invocation.

The problem does not manifest when a subcommand is invoked only once in a JVM process. Also, applications that do not reuse a `CommandLine` instance, and applications without subcommands, are not impacted.

The problem was introduced in version 4.2.0 and does not exist in picocli 4.1.4 and earlier.

# Workaround

This regression does not apply to options that have a `defaultValue` defined in the annotation: options defined as `@Option(... defaultValue="...")` will correctly be reset to the default value in between multiple invocations. Use the defaultValue annotation where possible.

# Cause

The cause is related to recent changes for [lazy subcommand instantiation](https://github.com/remkop/picocli/issues/690) and [repeatable subcommands](https://github.com/remkop/picocli/issues/454).

```java
private TypedMember(Field field, IScope scope) {
    this(field);
    if (ObjectScope.isProxyClass(scope)) {
        throw new InitializationException("Invalid picocli annotation on interface field");
    }
    FieldBinding binding = new FieldBinding(scope, field);
    getter = binding; setter = binding;
    this.scope = scope;
    hasInitialValue &= ObjectScope.hasInstance(scope); // <--- used to be obj != null
}
```

I may be possible to fix this by implementing this idea: 
https://github.com/remkop/picocli/issues/961 (similar to https://github.com/remkop/picocli/issues/962)

remkop:
Fixed in master.

Issue #962, opened 2020-02-20
behavior of defaultValue


sjyMystery:
Hi, guys:
  There's an unexpected behavior of the `defaultValue` param in `@Option` Annotation.
   When use default values from environment variables of invalid types, it always throws ParameterException even though I passed the value of the argument.

# How to solve it
  Maybe we can disable the type-checking of default value when the argument value has been already passed.

  And I will be honored to submit the example code of this situation and create a  pull request if necessary.

remkop:
Can you please show how to reproduce this problem?

sjyMystery:
Of course, the code is below:

```java
import picocli.CommandLine;

@CommandLine.Command(name = "run")
public class App implements Runnable {

    @CommandLine.Option(names = "--port", defaultValue = "${env:PORT}", required = true)
    Integer PORT;


    static public void main(String[] args) {
        App app = new App();
        args = new String[]{"--port", "5672"};
        CommandLine cmd = new CommandLine(app);
        cmd.execute(args);
    }

    public void run() {
        System.out.println(PORT);

    }
}
```

before run :
```shell script
export PORT="xxx"
```

sjyMystery:
then may got:

```
Invalid value for option '--port': 'xxx' is not an int
Usage: run [--port=PORT>]
      --port=<PORT>
```

remkop:
@sjyMystery Thank you for the clarification. I understand the issue now.
It looks like this is the same cause as this ticket: https://github.com/remkop/picocli/issues/961 (I just changed the title of that ticket.)

Please read my [comment on that ticket](https://github.com/remkop/picocli/issues/961#issuecomment-588217786) for more details.

Are you interested in providing a pull request for this? That would be great!

sjyMystery:
I've get your point.By the way, is there anything need or suggested to read before I make a pr?Thanks.

remkop:
@sjyMystery There is no documentation on the implementation, just the source code. :-)

It may be as simple as reordering these 2 lines in `CommandLine$Interpreter.parse(List, Stack, String[], List)` on [line 11486](https://github.com/remkop/picocli/blob/master/src/main/java/picocli/CommandLine.java#L11486):

Before: 

```java
private void parse(List<CommandLine> parsedCommands, Stack<String> argumentStack, String[] originalArgs, List<Object> nowProcessing) {
...
  try {
    applyDefaultValues(required); // currently defaults are applied first
[...]

Issue #961, opened 2020-02-19
Default value should only be applied if value is missing (was: ITypeConverter called for default value only if value is missing)


chirlo:
As of 4.2.0, if:

- a parameter has a default value and a  `ITypeConverter` defined,

- **and** a value is provided for this parameter

then the converter gets called twice, once with the default value and a second time with the provided value. 
For example, for the following parameter

    @Parameters(index = "0", description = "name", defaultValue = "alice", converter = NameTypeConverter.class)
    Name name;`

The call `cmd bob` will call the `NameTypeConverter#convert(String)` once with the parameter `alice` and once with `bob`.

I have a use-case where a given json file is parsed to a POJO , with a default value of `config.json`. If the parameters is specifically provided as `my_config.json` , the converter will fail trying to parse the default value of `config.json`, since it doesnt' exist.

Ideally, since converters can fail or do some  extra checks, I'd suggest that the default value should only be evaluated if not specifically specified.

remkop:
Hi @chirlo,

I see what you are saying. 

For background, picocli must behave well when a single `CommandLine` instance is used multiple times to parse different sets of command line args.

For that reason, all options and positional parameters are first "reset" to their initial value (if they were declared with an initial value), next they are set to their default value (if defined), and finally a value may be assigned (overwritten again) when the option is matched on the command line.

I can look into doing this the other way around: parse the input, remember which options and positional parameters were matched, and only set the default/initial value on options and positional parameters that were not matched. This will require a bit more book-keeping, and the timing of when to apply defaults becomes a bit more tricky, but it should be possible. It may take some time though.

Are you interested in providing a pull request for this?

In your application, were you able to work around this?

remkop:
Fixed in master.

To verify this, check out picocli master locally, and build it with `publishToMavenLocal`. That should publish `picocli-4.2.1-SNAPSHOT` to your local .m2 Maven cache.

```
git clone https://github.com/remkop/picocli.git
cd picocli
gradlew publishToMavenLocal
```

You can then try this in a project that uses the `info.picocli:picocli:4.2.1-SNAPSHOT` dependency. Feedback very welcome!

Issue #690, opened 2019-05-10
When CommandLine is initialized with IFactory all subcommands get created instead of only the one called


danielBreitlauch:
I create a CommandLine with an IFactory containing a Guice injector.
The cli consists of a lot of subcommands that are heavy to initialize.
All of them injection quite some heavy client instances. So I was expecting that picocli is only creating objects of subcommands that are needed. Initializing all the unused, not called subcommands generates quite some start up overhead.

remkop:
Sorry for my late response. I was running a marathon. :-)

I need to think about this a bit. I'm open to this in principle, as long as it does not introduce too many complexities.

danielBreitlauch:
No worries (besides that you will have to work on your marathon time ;)

I dont know enough about the internals but it could be quite simple.
Do you need (sub)command class instances in order to parse the input? 
The annotations should be readable from the class not the instance right?

Then the command line can be parsed without instantiating any objects.
And then only the command objects need to be instantiated that are actually used.

If I get some time I will have a look but I guess you can do that way faster ;D
Thx for considering it. I think it wil also give a nice speed boost in some instances.

remkop:
Not sure when I will be able to get to it, given the todo list for [4.0-beta](https://github.com/remkop/picocli/milestone/53) and [4.0](https://github.com/remkop/picocli/milestone/40).

If you have time to look at it that would be great.
The place to start is probably `CommandLine.Model.CommandReflection#extractCommandSpec`; it currently instantiates the user object for each command and uses it to create a `IScope` object for the annotated fields and methods. One idea would be to create a lazy `IScope` implementation that does not create the instance until the `get` method is called.

remkop:
FYI: I have started to reduce the scope for picocli 4.0 in order to get something out before the end of the summer. I won't be able to get to this item until after that.
If you need something sooner, please consider contributing a pull request. :-)

remkop:
@danielBreitlauch a first cut of lazily instantiated user objects has landed in master. Please try it out and let me know what you think.

danielBreitlauch:
Thanks I will

remkop:
@danielBreitlauch Sorry, it looks like there is still more work to be done.
Commands objects are still initialized to capture the initial value of `@Option` and `@Parameter`-annotated fields; this initial value is used to clear field values before parsing in case a CommandLine instance is being reused. I need to look at this more.

remkop:
@danielBreitlauch I believe it is all fixed now. 
[...]

PR #1107, opened 2020-06-12
add ability to pass options via command's constructor

eitansuez:
hi remko,
  i'd like you to review this commit.
  the intent here is to support passing options via the constructor.

  i don't think it should be merged as is, without your review and perhaps some tweaks, as i don't fully understand all the magic that picocli does.

  this code change causes one of the subcommandtests to fail, but i hope that's something that's easily rectified.

  i definitely need to spend more time understanding the code.

  my main reflection at the moment is that i suspect spring does similar work when it builds application contexts.  i wonder if this project can leverage the same utilities that spring uses to build the command object, leveraging autowiring, constructor and setter injection so we don't have to.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -5723,6 +5723,7 @@ public class CommandLine {
             private final Interpolator interpolator = new Interpolator(this);
             private final UsageMessageSpec usageMessage = new UsageMessageSpec(interpolator);
             private TypedMember[] methodParams;
+            private TypedMember[] constructorParams;
 
             private final CommandUserObject userObject;
             private CommandLine commandLine;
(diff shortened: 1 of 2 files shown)


remkop:
> i definitely need to spend more time understanding the code.

Yes, I worked on this area of user object instantiation recently when implementing [#690], and when refactoring how default values are applied ([#962][#961][#990]). I am keenly aware that user object creation is one of the most complex areas of the library. I am not sure if this is because of the implementation or whether it is because of the nature of the problem.

This is why I mentioned the trade-off between the benefit that this feature would bring vs the increased maintenance burden of the additional complexity.

> my main reflection at the moment is that i suspect spring does similar work when it builds application contexts. i wonder if this project can leverage the same utilities that spring uses to build the command object, leveraging autowiring, constructor and setter injection so we don't have to.

This may be interesting to explore.  picocli itself should not require Spring to work, but there is a [picocli-spring-boot-starter](https://github.com/remkop/picocli/tree/master/picocli-spring-boot-starter) module, and one idea may be to support constructor injection for applications that use that module.

Speaking of Spring and dependency injection, picocli currently supports [integration with dependency injection containers](https://picocli.info/#_dependency_injection) via the `IFactory` interface. This allows DI containers to inject services and beans into subcommands, type converters, and other components. 

If our constructor injection implementation instantiates subcommands via reflection, without using the factory, then such subcommands that pass options to the constructor may not be able to get other services and beans injected by the dependency injection container... That would be a big drawback. Supporting constructor injection via the [PicocliSpringFactory](https://github.com/remkop/picocli/blob/master/picocli-spring-boot-starter/src/main/java/picocli/spring/PicocliSpringFactory.java) may avoid that problem.

eitansuez:
you are correct.  the IFactory.create(cls) method needs to be revised to include one more parameter: the commanSpec's constructorParams field, which becomes necessary to call the constructor with parameters.  i dodged making this change to avoid the cascade of changes to every client of IFactory, but that's clearly not right.  so this PR needs more work.

remkop:
We cannot break existing factories.

To be honest, I am having second thoughts about this. Is it going to be worth it? How valuable is constructor injection in our use case? 
I am less and less convinced that this is a good trade-off.

eitansuez:
i suppose i can create a type of factory where the initialization parameters are passed in when the factory is built, and then the create() method would not need to be modified.  in that way the interface would not change.  but i agree with you, i don't want to introduce a chance to the codebase that makes it more brittle.  let's abandon this PR altogether.  thanks for entertaining it.

remkop:
That makes sense. 
Thank you for exploring this!

Issue #1109, opened 2020-06-14
manpage generator: encoding issue


deining:
Using the new man page generation feature, I tried to generate a man page for the [I18NDemo](https://github.com/remkop/picocli/blob/master/picocli-examples/src/main/java/picocli/examples/i18n/I18NDemo.java) given in the examples section of the repo.

When invoking `ascciidoctor` on the produced `.adoc` file, I'm getting:

`Failed to load AsciiDoc document - invalid byte sequence in UTF-8`

I can easily cure this by adding the `encoding` attribute to the produced .adoc file:

```
// tag::picocli-generated-full-manpage[]
// tag::picocli-generated-man-section-header[]
:doctype: manpage
:encoding: ISO-8859-1
```

IMHO, encoding is something that should be considered by the man page generator.
AFAIK, since Java 9, to can use UTF-8 encoding for your resource bundles, so simply adding the `encoding` attribute to the generated file seems to be no valid option for me. Maybe we can overcome this by adding an additional encoding option to the `gen-manpage` command?

remkop:
I see, yes that makes sense. 
The default value would be the platform encoding, I guess. 
Sounds like a good idea. 
Will you be able to provide a pull request for this?

PR #1112, opened 2020-06-15
PR: manpage generator: overcome encoding issues (#1109)

deining:
I found a fairly easy way to overcome the encoding issues addressed in #1109. I guess the proposed changes should be sufficient here. Generating man pages for the [I18NDemo](https://github.com/remkop/picocli/blob/master/picocli-examples/src/main/java/picocli/examples/i18n/I18NDemo.java) given in the examples section of the repo works perfect now.
Unfortunately, the patch breaks two exiting tests and I didn't find an easy way to make them work again. Can you have a look here?
Also while not strictly necessary, I would suggest to explicitly add the encoding attribute to the generated .adoc files (second commit).
Do we need test(s) for the patch? If so, can you take over or can you bring me on the right track here? Thanks.

### DIFF
--- a/picocli-codegen/src/main/java/picocli/codegen/docgen/manpage/ManPageGenerator.java
+++ b/picocli-codegen/src/main/java/picocli/codegen/docgen/manpage/ManPageGenerator.java
@@ -22,8 +22,10 @@ import picocli.codegen.util.Assert;
 import picocli.codegen.util.Util;
 
 import java.io.File;
+import java.io.FileOutputStream;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.lang.reflect.Method;
 import java.util.*;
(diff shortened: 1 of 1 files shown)


remkop:
Hi @deining, thanks for the PR.
I am not sure which tests broke, if I apply the patch all tests in picocli-codegen seem to pass.

Overall I like what you are doing here, and I would go one step further and make this an option, so that users can control the encoding. Making UTF8 the default seems a good idea.
What do you think?

remkop:
About additional tests, once we add an option where users can control the encoding, I can add a test with some multi-byte character text.

deining:
> Overall I like what you are doing here, and I would go one step further and make this an option, so that users can control the encoding. Making UTF8 the default seems a good idea.
> What do you think?

According to this [comment](https://github.com/asciidoctor/asciidoctor.org/issues/160#issuecomment-174753569), asciidoctor discarded the :encoding: attribute and solely relies on UTF-8 as encoding for `.adoc` files. Having this learnt, letting the user control the encoding of the generated file does not seem an option to me any more. Also my second commit needs to be discarded.

remkop:
Ah interesting. Like you said, that rules out the idea of user-specified encoding.
That simplifies things, that is good.
Thanks for investigating!

You mentioned some tests were breaking, but I did not see that. Which tests were that?

Other than that, I think this PR is ready to be merged.

remkop:
> Other than that, I think this PR is ready to be merged.

I typed a bit too fast... :-) Can you discard your [second commit](https://github.com/remkop/picocli/pull/1112/commits/da1956d47088982c1e9ef7dba85ee42c6b8ec636) so we no longer add the `:encoding:` attribute?

deining:
Just discarded second commit.
About the breaking tests: these were false alarms.
PR should be ready for merging.

remkop:
Thank you! Bedtime for me now. Will look further tomorrow!

remkop:
Merged into master! Thanks for the contribution!

PR #1116, opened 2020-06-20
DOC: Fix typo in picocli user manual

sakata1222:
I found a typo in the "Guice Example".

- The method variable name of `GuiceFactory#create` is `aClass`.
- I think the 1st argument of `new CommandLine` should have annotations of picocli, so we should specify a implemented class (i.e. `InjectionDemo.class`) not `Runnable.class`
- If we specify the implemented class, bind definition for `Runnable.class` might not be unnecessary.


sakata1222:
I found the other mistakes, I will update and reopen this pull request.

remkop:
Merged into master.

PR #1150, opened 2020-08-20
Fix terminal width detection on macOS

electrum:
`stty` on macOS outputs the value *before* the name, like this:
```
speed 9600 baud; 40 rows; 80 columns;
```
Rather than trying to fix the regex to match either one, we can use the simpler `tputs cols` command.


remkop:
@electrum Picocli 4.5.1 has been released, including your fix for this issue.
Thanks again for the contribution!

electrum:
Thanks @remkop! I tested that version and confirmed the wrapping works.

remkop:
Excellent!
Thanks again for the contribution!

PR #1155, opened 2020-08-26
Documentation: fix sample code in chapter "Validation"

deining:


remkop:
Thanks for the review and the fixes! 
I should obviously run the sample code before putting it in the docs! :flushed:

Merged and published.
Thanks again!

PR #1167, opened 2020-09-10
Quick guide: fix broken links

deining:


remkop:
Merged.

Nice catch! Thank you as always!!

PR #1169, opened 2020-09-11
Close org.jline.terminal.Terminal (fixes #1168)

### LINKED ISSUE #1168: jline3 example should close Terminal (using try-with-resource or try-finally)
Should this line ensure that the terminal is closed? https://github.com/remkop/picocli/blob/72e1d4f2d35f8c12a0b64d8fcd47c2df02269c9f/picocli-shell-jline3/src/test/java/picocli/shell/jline3/example/Example.java#L165

In the docs https://www.javadoc.io/doc/org.jline/jline/3.14.1/org/jline/terminal/Terminal.html it states that "Terminals should be closed by calling the Closeable.close() method in order to restore their original state" although I cannot tellw ithout looking at the code what resources it might actually free and what constitutes the "original state".

dwalluck:


remkop:

PR #1170, opened 2020-09-11
Ensure ANSI is disabled in ManPageGeneratorTest regardless of environ…

dwalluck:
…ment (issue #1103)


dwalluck:
Sorry that I didn't find this one at the same time as #1166. I think that I hadn't updated my master branch maybe.

remkop:
Merged.

Thanks for catching this!

PR #1172, opened 2020-09-11
Fix #1156, make HelpCommand respect subcommands case-sensitivity and abbreviation

### LINKED ISSUE #1156: Built-in Help being used on subcommands, still case-sensitive even if case-sensitivity disabled
Hey, first of all I'll start with saying that I'm just about done using Pico for a work project and it's been absolutely amazing.

One small inconsistency that I still have left (and just put down in a "common issues" wiki section) is that even with case-sensitivity disabled, for example:

`[Tools] help SUBCOMMAND`

Gives an "Unknown subcommand" error, despite "subcommand" existing with different capitalization. This is inconsistent with everything else, and the only thing that the user needs to be case-sensitive for, as:

`[Tools] HELP subcommand`
`[Tools] SUBCOMMAND`
`[Tools] subcommand -OpTiOn`

All work perfectly fine. It's *just* when trying to run help on a subcommand, the subcommand's name must be perfect.

I have a hunch that the issue lies in the implementation of the Help command, but am not sure. I'm open to working on this myself if I have some free time in a few weeks, but figured I'd put this issue out there now.

NewbieOrange:
This pull request will fix #1156.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -13942,12 +13942,13 @@ public class CommandLine {
             if (parent == null) { return; }
             Help.ColorScheme colors = colorScheme != null ? colorScheme : Help.defaultColorScheme(ansi);
             if (commands.length > 0) {
+                Map<String, CommandLine> parentSubcommands = parent.getCommandSpec().subcommands();
                 String fullName = commands[0];
                 if (parent.isAbbreviatedSubcommandsAllowed()) {
-                    fullName = AbbreviationMatcher.match(parent.getCommandSpec().subcommands().keySet(), commands[0],
-                            parent.getCommandSpec().subcommandsCaseInsensitive(), self);
+                    fullName = AbbreviationMatcher.match(parentSubcommands.keySet(), fullName,
+                            parent.isSubcommandsCaseInsensitive(), self);
                 }
-                CommandLine subcommand = parent.getCommandSpec().subcommands().get(fullName);
+                CommandLine subcommand = parentSubcommands.get(fullName);
                 if (subcommand != null) {
                     if (outWriter != null) {
                         subcommand.usage(outWriter, colors);
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -365,10 +365,7 @@ public class CommandLine {
      * @since 0.9.7
      */
     public Map<String, CommandLine> getSubcommands() {
-        CaseAwareLinkedMap<String, CommandLine> subcommands = new CaseAwareLinkedMap<String, CommandLine>();
-        subcommands.setCaseInsensitive(isSubcommandsCaseInsensitive());
-        subcommands.putAll(getCommandSpec().subcommands());
-        return subcommands;
+        return new CaseAwareLinkedMap<String, CommandLine>(getCommandSpec().commands);
     }
     /**
      * Returns the command that this is a subcommand of, or {@code null} if this is a top-level command.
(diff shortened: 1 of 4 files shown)


src/main/java/picocli/CommandLine.java
```diff
@@ -13924,7 +13934,7 @@ public void run() {
             if (parent == null) { return; }
             Help.ColorScheme colors = colorScheme != null ? colorScheme : Help.defaultColorScheme(ansi);
             if (commands.length > 0) {
-                CommandLine subcommand = parent.getSubcommands().get(commands[0]);
+                CommandLine subcommand = parent.getCommandSpec().subcommands().get(commands[0]);
```
NewbieOrange, line 13937: To avoid unnecessary copies here

remkop: What did you think of my [suggestion](https://github.com/remkop/picocli/issues/1156#issuecomment-690960531) (which may have been a bit hidden in the code) to use 

```
AbbreviationMatcher.match(parent.getCommandSpec().subcommands().keySet(),
                        commands[0], parent.getCommandSpec().subcommandsCaseInsensitive(), self);
```

to also add support for abbreviations to the `HelpCommand`?

remkop: Oops, nevermind, looks like you already did this.

src/main/java/picocli/CommandLine.java
```diff
@@ -365,7 +365,10 @@ public CommandLine addSubcommand(String name, Object command, String... aliases)
      * @since 0.9.7
      */
     public Map<String, CommandLine> getSubcommands() {
-        return new LinkedHashMap<String, CommandLine>(getCommandSpec().subcommands());
+        CaseAwareLinkedMap<String, CommandLine> subcommands = new CaseAwareLinkedMap<String, CommandLine>();
+        subcommands.setCaseInsensitive(isSubcommandsCaseInsensitive());
+        subcommands.putAll(getCommandSpec().subcommands());
```
remkop, line 370: If we have a copy constructor that accepts a `CaseAwareLinkedMap` instead of a plain `Map`, then we can call `setCaseInsensitive` inside the constructor, based on whether the passed-in  `CaseAwareLinkedMap` parameter is case-sensitive.

That would reduce the above  code to simply:

```
return new CaseAwareLinkedMap<String, CommandLine>(getCommandSpec().commands);
```

src/main/java/picocli/CommandLine.java
```diff
@@ -5557,24 +5560,34 @@ public int size() {
...
+            }
+
+            /**
+             * Constructs a {@code CaseAwareLinkedMap} instance with the same mappings as the specified map.
+             * The {@code CaseAwareLinkedMap} instance is created with a default locale {@link java.util.Locale#ENGLISH}.
+             * @param map the map whose mappings are to be placed in this map
+             * @throws NullPointerException if the specified map is null
+             */
+            public CaseAwareLinkedMap(Map<? extends K, ? extends V> map) {
+                this(ENGLISH);
```
remkop, line 5589: I propose we change this constructor to accept a `CaseAwareLinkedMap` instead of a plain `Map`.
In the implementation, we can then copy the `locale`, `caseInsensitive`, and other attributes from the passed-in  `CaseAwareLinkedMap` parameter. That makes the main use case easiest to use.

src/main/java/picocli/CommandLine.java
```diff
@@ -13924,7 +13942,13 @@ public void run() {
...
             Help.ColorScheme colors = colorScheme != null ? colorScheme : Help.defaultColorScheme(ansi);
             if (commands.length > 0) {
-                CommandLine subcommand = parent.getSubcommands().get(commands[0]);
+                Map<String, CommandLine> parentSubcommands = parent.getCommandSpec().subcommands();
+                String fullName = commands[0];
+                if (parent.isAbbreviatedSubcommandsAllowed()) {
+                    fullName = AbbreviationMatcher.match(parentSubcommands.keySet(), fullName,
+                            parent.isSubcommandsCaseInsensitive(), self);
+                }
+                CommandLine subcommand = parentSubcommands.get(fullName);
```
remkop, line 13951: Nice!

remkop (changes requested):
Very nice, many thanks!
I think we are almost there.

src/main/java/picocli/CommandLine.java
NewbieOrange: The issue here is that subcomands() actually returns an unmodifiable-map, which cannot be cast to CaseAwareLinkedMap without some nesty hacks. Not sure how to workaround that without breaking encapsulation.

remkop: You can reference the `CommandSpec.commands` field directly, that has the right type.

remkop:
Merged.

@NewbieOrange Thank you very much for the PR and for your patience during the reviews!
Your contributions are much appreciated!

PR #1173, opened 2020-09-12
Different sample applications for manual and quick guide (#1171)

deining:


remkop:
Merged. 
Very nice, thank you!
I’m still polishing it a bit and will update later.

remkop:
Thanks again for the improvements.
I made some more changes and published the user manual and Quick Guide now.
Let me know what you think.

deining:
> Thanks again for the improvements.

You are welcome.

> I made some more changes and published the user manual and Quick Guide now.

I'm glad you made some polishment here. Your corrections on the quickguide are real improvements, I really appreciate your input here.

> Let me know what you think.

Having worked on the documentation a bit, I'm on the impression that the current single page approach is reaching its limits.
Have you every thought about a multi page approach for the documentation?

remkop:
I like the single page: Having everything on a single page means that users can use browser search (Ctl-F) to easily find whatever they’re looking for. Multiple pages makes this harder for users while giving no advantage that I can see...

PR #1179, opened 2020-09-17
Use .invalid domain name for InetAddress test

electrum:
Some DNS setups such as ISP "DNS hijacking" will return an IP address for garbage names. Using the ".invalid" domain specified in RFC 6761 makes the test more reliable.


remkop:
Merged. Nice! I was not aware of RFC 6761, very good to know!

Thanks for the contribution!

PR #1187, opened 2020-09-25
Extend and improve documentation on subcommands (#890)

### LINKED ISSUE #890: Manual/Quick guide: better support for first time users
While answering #878, I realized, that both user manual and quick guide could be improved. Goal should be to get first time users started as quickly as possible.
Thinking back to my first time contact with picocli I remembered it wasn't straightforward for me to get things up and running, either. Don't get me wrong: The picocli manual is very complete, concise and was/is of great help for me in many cases. However, it has -well- manual-style, and this might a hindrance for first time users. Luckily I see good chances to iron this out quite easily. Here are my thoughts/comments on this: 

I assume first time users start reading at the beginning of the manual. Currently, the outline is:

* Introduction
* Getting started
* Options and parameters

What I'm missing here: nowhere in the manual, first time users are introduced to the concepts of commands and how to annotate them. This is a shortcoming IMHO that may lead to questions as exemplified in #878.

The introduction features the `Checksum` example, but it does not explain how to annotate command classes at all. All it states is:

```
How it works: annotate your class and picocli initializes it
from the command line arguments, converting ...
```

I tried to address this issue already while updating the quick guide to the execute API, that's what I came up with:

```
Create a class that represents your command and annotate
this class with @Command. Then, annotate the fields or
methods of the command class with @Option or @Parameters
to ...
```

IMHO, this should be reflected in the user manual, too. I doubt whether this is sufficient however.
This is what I would like to suggest/ see implemented:

a) Extend the 'Getting started' chapter like this:

* Getting started
  * Setting up your enviroment
    * Add as external dependency
    * Add as source
    * Annotation processor
  * Real word example: `Checksum`command

In short I would propose to move the `Checksum` example from the chapter `Introduction` to the chapter `Getting started`. The example should be explained in more detail there (going through the releveant lines of the source code step by step).

b) Add an additional chapter to the outline (optional): 

* Introduction
* Getting started
* **Commands**
* Options and parameters

The `Commands` chapter should cover subcommands, too, since this is a common requirement for many command line apps (at least for me, see also #878).

Just my two cents, what do you think?

deining:
In an order to eventually close #890, I authored this PR, I hope you like it.
Feel free to perform some polishment like you did before.

### DIFF
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -114,6 +114,11 @@ The example `main` method calls `CommandLine.execute` to parse the command line,
 
 The <<Mixin Standard Help Options,mixinStandardHelpOptions>> attribute adds `--help` and `--version` options to your application.
 
+[TIP]
+====
+You may find more https://picocli.info/quick-guide.html#_example_applications[sample applications] in the picocli quick guide, they are explained in detail there.
+====
+
 == Getting Started
 You can add picocli as an external dependency to your project, or you can include it as source.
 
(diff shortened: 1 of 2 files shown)


remkop:
Thank you for the contribution again!
I plan to merge this soon, I have some other commitments I need to take care of first.

remkop:
I am also rewriting and rearranging some of the other subsections of the Subcommands section.

About the new Quick Guide example:

> The `ISOCodeResolver` application has two subcommands, `language` and `country`, that may be used to either resolve either languages or country codes following the ISO standard.

Can you clarify (ideally with links) which ISO standard(s) this refers to?

remkop:
Ah sorry, I found that the code example references the exact ISOs. Got it.

PR #1198, opened 2020-10-01
Documentation: add copy button to code blocks

deining:
Using the button, you can copy textual content of code block to clipboard.


remkop:
Merged, and pushed the rendered HTML.

This is very nice! It is subtle but a very nice enhancement for users.
I like this a lot! Thanks for the contribution again!

PR #1200, opened 2020-10-02
User manual: online execution of checksum sample app

deining:
See discussion [here](https://github.com/remkop/picocli/pull/1196#issuecomment-702616049).

### DIFF
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -64,6 +64,15 @@ TIP: Picocli also provides a <<Programmatic API,programmatic API>>, separately f
 === Example application
 The example below shows a short but fully functional picocli-based `checksum` command line application:
 
+[TIP]
+====
+You might want to run the `checksum` application online first.
+To do so, https://www.jdoodle.com/embed/v0/2mQM?stdin=1&arg=1[open^] the source code in a separate browser tab and press the blue `Execute` button below.
+This prints the checksum of the file `/etc/hosts` to the console.
+Then type a file name (e.g. `/usr/bin/java` or `/proc/cpuinfo`) into the text field `Commandline Arguments` and execute the program again.
+Fill in `--help` to get usage instructions.
+====
+
 .A full working example picocli-based command line application
 [[CheckSum-application]]
 [source,java]


docs/index.adoc
```diff
@@ -61,8 +61,13 @@ How it works: annotate your class and picocli initializes it from the command li
 
 TIP: Picocli also provides a <<Programmatic API,programmatic API>>, separately from the annotations API.
 
-=== Example application
-The example below shows a short but fully functional picocli-based `checksum` command line application:
+=== Example application: Checksum
```
remkop, line 64: I am not in favor of adding the word ": Checksum"
* it will cause the title to wrap in the TOC
* it distracts from the purpose: the purpose is to provide an example that gives a taste of picocli's power and ease of use. The fact that it is also a useful program by itself is secondary.

docs/index.adoc
```diff
@@ -61,8 +61,13 @@ How it works: annotate your class and picocli initializes it from the command li
...
+`Checksum` is a short but fully functional picocli-based command line application which prints the checksum (MD5 by default) of a file to STDOUT.
```
remkop, line 65: For the same reason, I think the original description was better: I don't want to emphasize too much what the example does.

docs/index.adoc
```diff
@@ -61,8 +61,13 @@ How it works: annotate your class and picocli initializes it from the command li
...
+
+[TIP]
+====
+You might want to run the `checksum` application online first. To do so, open the source code in a separate https://www.jdoodle.com/embed/v0/2mQM?stdin=1&arg=1[browser tab^] and press the blue `Execute` button below. This prints the checksum of the file `/etc/hosts` to the console. Then type a file name (e.g. `/usr/bin/java` or `/proc/cpuinfo`) into the text field `Commandline Arguments` and execute the program again. Fill in `--help` to get usage instructions.
+====
```
remkop, line 70: Great! I like it.

docs/index.adoc
```diff
@@ -100,7 +105,7 @@ class CheckSum implements Callable<Integer> {
     public Integer call() throws Exception { // your business logic goes here...
         byte[] fileContents = Files.readAllBytes(file.toPath());
         byte[] digest = MessageDigest.getInstance(algorithm).digest(fileContents);
-        System.out.printf("%0" + (digest.length*2) + "x%n", new BigInteger(1, digest));
+        System.out.printf(algorithm + " hash of " + file.getPath() + ": %0" + (digest.length*2) + "x%n", new BigInteger(1, digest));
```
remkop, line 108: I agree that this makes the program a better `checksum` application. But I prefer the simplicity of the current version.

There are other ways in which we could improve the example, like allowing it to deal with multiple files, etc. But this would distract from the purpose, which is to give a glimpse of the power and ease of use of the library.

docs/index.adoc
remkop: One thing: I like the convention of one sentence per line in AsciiDoc.
Can you change that?

remkop (changes requested):
Sorry, I am very picky about especially the beginning of the user manual.
Please don't take this as criticism, but as positive feedback!

deining:
> Sorry, I am very picky about especially the beginning of the user manual.
> Please don't take this as criticism, but as positive feedback!

I like the way we interact and cooperate on this project.
Don't merge anything that you are not convinced of.
In the past, your comments always lead to an inprovement in some or the other way, let's go ahead with this!

docs/index.adoc
deining: I saw the title wrap in the TOC already and didn't like it, too.
Reverted to original version.

docs/index.adoc
deining: Reverted to original version.

docs/index.adoc
deining: AsciiDoc now has on line per sentence.
As another minor change, I changed the link text.

docs/index.adoc
deining: Reverted back to original version.
I left the extended `printf` statement in the Jdoodle version since I think it makes sense here. This comes at the price that code sample in the manual and jdoodle are slight different. I think this is mor than outweighed by the fact that the user can immediately get a grasp what the application is doing when executing the jdoodle app.

remkop:
Merged. Thanks a lot, this is pretty cool!
I made the text as terse as possible, I hope you like it.

PR #1202, opened 2020-10-04
Update to latest Asciidoctor gradle plugin

deining:
This PR updates gradle Asciidoctor plugin to latest version.
After applying this PR, `gradle ascii` works again on master.

Unfortunately, now tabbed panes do not show up any more. :-(
Seems like the used extension is not picked up any more by the new version. I will look into that soon.
For now, the produced `index.html` is **not** ready for being published to the web site.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ buildscript {
     }
 
     dependencies {
-        classpath "org.asciidoctor:asciidoctor-gradle-plugin:$asciidoctorGradlePluginVersion"
+        classpath "org.asciidoctor:asciidoctor-gradle-jvm:$asciidoctorGradlePluginVersion"
         classpath 'org.asciidoctor:asciidoctorj-pdf:1.5.0-beta.8'
         classpath "com.jfrog.bintray.gradle:gradle-bintray-plugin:$gradleBintrayPluginVersion"
         classpath "gradle.plugin.org.beryx:badass-jar:1.1.3"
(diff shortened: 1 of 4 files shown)


remkop:
Would it not be easier to revert the last commit (2453cb49)?

deining:
> Would it not be easier to revert the last commit ([2453cb4](https://github.com/remkop/picocli/commit/2453cb49bcf863837a10b41d59dca3fedccecc45))?

I force-pushed some changes for this PR, everything is working now:

- `gradle ascii` runs successfully (using the latest version of the asciidoctor gradle plugin)
- the produced manual file `index.html` now shows a tabbed source code pane, showing both the Java and the Kotlin version of the checksum app.

Please note that with the new version of the gradle plugin, html output is now written to `build/docs` (was: `build/docs/html5`). I don't know if this breaks your workflow when publishing documents. If so it could be easily corrected by adjusting the `outputDir` attribute of the `asciidoctor` task.

remkop:
Merged. Looks good!
One thing I noticed is that the Copy source function no longer works, but if we have to chose, they are both nice, but I think I like the tabbed examples feature better. 

I should look into Groovy and Scala versions of the `checksum` example too, but at the moment too busy with other commitments...

deining:
> Merged. Looks good!
> One thing I noticed is that the Copy source function no longer works, but if we have to chose, they are both nice, but I think I like the tabbed examples feature better.

Me too!

About the `copy` function:  I realized already that the copy button is missing, but haven't found time to look into this. Using gradle 5.1, both the `copy`button and the tabbed pane are working, so it can be done :-) 

> I should look into Groovy and Scala versions of the `checksum` example too, but at the moment too busy with other commitments...

If I remember correctly, there are files for that in the examples section already. I can take that over and will have a look soon.

PR #1203, opened 2020-10-05
Add source code highlighting for Kotlin, Scala and Groovy

deining:
This PR changes changes source highlighter to `rouge` since the current highlighter `coderay` does not support Kotlin and Scala.
This gives a better result for the listings in chapter 30 `Picocli in Other Languages`.


remkop:

PR #1205, opened 2020-10-06
Manual: add checksum sample app coded in Groovy and Scala

deining:
This PR adds checksum sample app coded in Groovy and Scala.

For Groovy language, I decided to add a Groovy scipt.
Should we add Groovy example with class, too? What do you think?

### DIFF
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -82,7 +82,7 @@ import java.nio.file.Files;
 import java.security.MessageDigest;
 import java.util.concurrent.Callable;
 
-@Command(name = "checksum", mixinStandardHelpOptions = true, version = "checksum 4.0",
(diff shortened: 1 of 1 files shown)


remkop:
Thank you!
I went ahead and added a Groovy class example, it is worth doing so because it has some significant differences with Groovy script.

Some small details I changed:

* I used `version = "checksum 4.0"` throughout; there is nothing in the examples that is tied it to the latest picocli version, but since it uses the `execute` API, this example requires picocli 4.0 at a minimum
* I replaced the single unicode character `…` with `...` (three dots) to be sure it can be rendered in any terminal with any character encoding and code page

Overall this looks very nice now!
It showcases how picocli can be used by all these JVM languages in a very natural way.
Thank you again!

PR #1209, opened 2020-10-12
Add Maven coordinates to JLine2/3 modules

Holmistr:
Minor documentation improvement. 

*Story behind it*
I wanted to try out the JLine integration. I copy pasted the example and then simply started to add imports. However, my IDE couldn't find the imports, so I was obviously missing some dependency. But the dependency to be used for this integration is not mentioned anywhere in the docs.


remkop:

PR #1210, opened 2020-10-12
Manual: add Kotlin version to subcommands example app

deining:


remkop:
Merged. Thank you for the contribution.
Not sure when I can push the rendered HTML, bit hectic now...

PR #1211, opened 2020-10-13
Manual, section subcommands: add several tabs with Kotlin source code

deining:
Time permitting, I will add more tabs with source code. So don't worry if you don't find time to recreate the html page immediately.


remkop:

remkop:
The rendered HTML is also published now.


By the way, I noticed that the Kotlin highlighter has some trouble with things like 

```
@Command(subcommands = [ A::class ]) // <--- notice the space between 'class' and `]` 
class X : Runnable
```

The space between 'class' and `]` causes everything until 'X' to have red background.
Removing the space between `::class` and the closing `]` seems to fix it.
(Just FYI.)

remkop:
And thanks again for the contributions! Shaping up very nicely!

Issue #566, opened 2018-12-10
usage help for nested sub listing full command tree


lgawron:
Currently invoking usageHelp for cli with nested subcommands shows only one level of commands.

Would it be possible to generate a full command listing like:

Commands:

- help     Displays help information about the specified command
- mobile   Manage device
  - dump-config 
  - disable
- user     Manage user
  - nuke 
- jobs     Manage jobs
  - list
  - trigger
  - disable
- version  Get server version

and so on?

remkop:
Update: there's a very promising PR in progress (#567) that will facilitate the above customization. If things go well I can do a 3.9 release with this.

remkop:
#530 (custom usage help message section renderers) has been merged to master.
Please try and see if this API allows you to achieve the layout you have in mind.

bobtiernay-okta:
This would be a very appreciated addition.

remkop:
I've added an example that shows how to accomplish this customization with the new API introduced in picocli 3.9:

```java
package picocli.examples.customhelp;

import picocli.CommandLine;
import picocli.CommandLine.Command;
import picocli.CommandLine.Help;
import picocli.CommandLine.Help.Column;
import picocli.CommandLine.Help.Column.Overflow;
import picocli.CommandLine.Help.TextTable;
import picocli.CommandLine.IHelpSectionRenderer;
import picocli.CommandLine.Model.CommandSpec;
import picocli.CommandLine.Model.UsageMessageSpec;

import static picocli.CommandLine.Model.UsageMessageSpec.SECTION_KEY_COMMAND_LIST;

/**
 * This example demonstrates how to customize a section of the usage help message.
 * It replaces the standard command list with a custom list that displays
 * not just the immediate subcommands but the full hierarchy of subcommands:
 * <pre>
 * Usage: showall [-hV] [COMMAND]
 * Demonstrates a usage help message that shows not just the subcommands of this
 * command, but also the nested sub-subcommands.
 *   -h, --help      Show this help message and exit.
 *   -V, --version   Print version information and exit.
 * Commands:
 *   sub1           subcommand1 of showall
 *     sub1sub1     subcommand1 of subcommand1 of showall
 *     sub1sub2     subcommand2 of subcommand1 of showall
 *   sub2           subcommand2 of showall
 *     sub2sub1     subcommand1 of subcommand2 of showall
 * </pre>
 *
 * As requested in https://github.com/remkop/picocli/issues/566
 */
[...]

PR #1215, opened 2020-10-14
User manual: documentation improvements

deining:
This PR improves user manual by

- adding missing semicola to Java code sample
- adding further tabs with Kotlin source to user manual
- adding instructions on usage help for nested subcommands

### DIFF
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -5392,6 +5392,17 @@ By default, the usage help message only shows the subcommands of the specified c
 and not the nested sub-subcommands. This can be customized by specifying your own <<Reordering Sections,`IHelpSectionRenderer`>> for the command list section.
 The `picocli-examples` module has an https://github.com/remkop/picocli/blob/master/picocli-examples/src/main/java/picocli/examples/customhelp/ShowCommandHierarchy.java[example] that shows how to accomplish this.
 
+To get usage help for nested subcommands, either set `mixinStandardHelpOptions = true` in all commands in the hierarchy, and/or add `picocli.CommandLine.HelpCommand` to the subcommands of all commands. Then you can get help message for a nested (sub-)subcommand like this:
+
+[source,bash]
+----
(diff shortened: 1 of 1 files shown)


docs/index.adoc
```diff
@@ -5284,6 +5392,12 @@ By default, the usage help message only shows the subcommands of the specified c
 and not the nested sub-subcommands. This can be customized by specifying your own <<Reordering Sections,`IHelpSectionRenderer`>> for the command list section.
 The `picocli-examples` module has an https://github.com/remkop/picocli/blob/master/picocli-examples/src/main/java/picocli/examples/customhelp/ShowCommandHierarchy.java[example] that shows how to accomplish this.
 
+If you want to get usage help for nested subcommands, make sure that all commands in the hierarchy have `picocli.CommandLine.HelpCommand` as (one or their) subcommands. Then you can get help message for a nested subcommand like this:
```
remkop, line 5395: That is not the only way though. How about this:

> To get usage help for nested subcommands, either set `mixinStandardHelpOptions = true` in all commands in the hierarchy, or add `picocli.CommandLine.HelpCommand` to the subcommands of all commands.

deining: You are right. There are two ways to get use help, so we should document both of them. Your proposal is perfect, I changed the PR accordingly.

Concerning the `bash`-block showing how to invoke usage help:
There was a previous [discussion](https://github.com/remkop/picocli/issues/566#issuecomment-453351720) that teached me that the way kow to invoke usage help is not necessarily self-explaining. That's why added the `bash`-block and IMHO, we should keep it. I modified the code block to take both ways of getting usage help into account. I agree it's a bit lengthy now, but I would like to keep that block, nevertheless.

Thanks for reviewing my modified PR!

remkop (approved):
That is a lot of work!
Looks great, thank you!

remkop:
Merged, and published the rendered HTML.

PR #1223, opened 2020-10-22
DOC: add tabs with Kotlin code for samples (manual chapter 10: Validation)

deining:


docs/index.adoc
```diff
@@ -3003,9 +3003,10 @@ Methods annotated with `@Option` and `@Parameters` can do simple input validatio
 
 The following example validates that the value specified for the `--prime` option is a prime number:
 
-[source,java]
+.Java
+[source,java,role="primary"]
 ----
-class SingleOptionValidationExample {
+class SingleOptionValidationExample implements Runnable {
```
remkop, line 3009: I was thinking to leave out the `implements Runnable`.  The main point of this example is that apps can use an `@Option`-annotated method to do validation. 

I think it is okay to for some examples to be fragments (as opposed to a fully executable program). Sometimes that helps to keep the example focused and communicates the point better.

docs/index.adoc
```diff
@@ -3030,13 +3031,42 @@ class SingleOptionValidationExample {
 }
 ----
 
+.Kotlin
+[source,kotlin,role="secondary"]
+----
+class SingleOptionValidationExample : Runnable {
```
remkop, line 3037: (Similar to the Java example, I suggest leaving out the `: Runnable` bit, to keep it as small as possible.)

remkop (changes requested):
Thanks again for the PR!

Let me know what you think about the idea to leave out the `implements Runnable` from `SingleOptionValidationExample` to keep it as small as possible.

docs/index.adoc
deining: I agree with your proposal. PR was modified as suggested.

remkop (approved):
Great, thank you!

PR #1228, opened 2020-10-25
Manual: add tabs with Kotlin code for samples (chapters 8, 16, 20)

deining:
BTW: what does this mean (index.adoc, line 6008): 

`this command adds the frup to the frooble`


remkop:
> BTW: what does this mean (index.adoc, line 6008):
> 
> `this command adds the frup to the frooble`

That sentence is just a random example description for the example `add` subcommand. Perhaps it should be replaced with something more obvious. :-)

remkop:

Issue #1184, opened 2020-09-23
Make some Help API public


remkop:
Some things in the usage help message are currently difficult to customize using the Help API, because methods are not public.
See https://github.com/remkop/picocli/issues/1181

Proposed changes:

* provide public getter methods for the `Help.Layout` attributes (colorScheme, textTable, optionRenderer , parameterRenderer )
* make `Help.calcLongOptionsColumnWidth` public

PR #1229, opened 2020-10-26
Fix compilation error introduced with fc5ef6de6 (#1184)

deining:


remkop:
Last-minute rushed changes, never good... :-(
Thanks for the fix!

PR #1232, opened 2020-10-26
Micronaut example: add Kotlin version, extended description of Micronaut use

deining:
This PR adds adds the Kotlin version for the Micronaut example (chapter 22).

When working on this, I extended the description a bit.
Also, the example was sligthly altered, it now closely resembles the auto-generated sample source file.

Also I removed these lines

```
@Client("https://api.github.com")
@Inject
var client: RxHttpClient? = null
```

since they weren't used inside the sample. As a side effect, number of imports could be reduced after removal of these lines.

Feel free to polish my description!

### DIFF
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -9373,35 +9373,133 @@ image:https://micronaut.io/images/micronaut_mini_copy_tm.svg[Micronaut Logo,widt
 
 The Micronaut microservices framework provides https://docs.micronaut.io/latest/guide/index.html#picocli[built-in support] for picocli with its `PicocliRunner` class.
 
-[source,java]
+
+Our micronaut example app provides a command line interface to a HTTP client that can be used to retrieve the number of stars for a GitHub project, leveraging the GitHub API. Owner and name (called `slug`) of the project to be queried must be given as command line parameter.
+
+To get started, you have to https://micronaut-projects.github.io/micronaut-starter/latest/guide/#installation[install] Micronaut framework first.
+Afterwards you can use the `mn`  command line interface to generate your first project `git-stars`. Don't forget the `--features http-client` option which will add the `io.micronaut:micronaut-http-client` dependency to the project.
+
+[source,bash]
 ----
-import io.micronaut.configuration.picocli.PicocliRunner;
-import io.micronaut.http.annotation.*;
-import io.micronaut.http.client.*;
+mn create-cli-app git-stars --features http-client --lang=java
+| Application created at /home/user/projects/micronaut/git-stars
+----
+
+This command will create a self contained micronaut project with picocli-support, together with its dependencies, a gradle build script and test cases.
+Open the auto-generated source file `GitStarsCommand.java` and alter and extend it so that it looks like this:
+
+.Java
+[source,java,role="primary"]
+----
+package git.stars;
+
+import java.util.Map;
 import javax.inject.Inject;
 
 import picocli.CommandLine.Command;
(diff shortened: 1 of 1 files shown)


remkop:
I would prefer to keep the example injected service. They illustrate that services can be injected thanks to the factory that enables dependency injection. The business logic is omitted, so it is not clear how they are used, that is true.

Perhaps we can clarify a little by adding a comment?

```java
    @Client("https://api.github.com")
    @Inject RxHttpClient client; // example injected service

...

    public void run() {
        // business logic here
        // injected services can be used here
    }
```

deining:
> I would prefer to keep the example injected service.

I do understand that, and in general I'm not against including these lines at all.

> They illustrate that services can be injected
> thanks to the factory that enables dependency injection.

That's true. Without having these lines use of Micronaut is overhead without real value.

My point is: without these lines, there is only single dependency for that sample: `io.micronaut.configuration:micronaut-picocli`. By adding these lines, you need a lot more dependencies which make the example more difficult to set up. And, based von my experience, improper setup of these dependiencies may cause the program to fail. And I simply want to save users from that potentially frustrating experience(s).

Proposal 1: we keep the lines as comments:

> ```java
>     // in a real life app, comment out the lines below
>     // @Client("https://api.github.com")
>     // @Inject RxHttpClient client; // example injected service
> 
> ...
> 
>     public void run() {
>         // business logic here
>         // injected services can be used here
>     }
> ```

Proposal 2: we show and explain the [example](https://micronaut-projects.github.io/micronaut-picocli/latest/guide/#_an_example_http_client) from the micronaut quick start guide. Currently, the example is quite long, we might be able to shorten it a bit. This would make obvious to the user how the injected services are used.

Proposal 3: ?? (any comments or ideas for a completely different sample are welcome!).

Looking forward to your comment!

remkop:
How about something like this:

```java
    @Client("https://api.github.com")
    @Inject RxHttpClient client; 

    @Parameters(description = "GitHub slug", paramLabel = "<owner/repo>") 
    String slug = "remkop/picocli";

    public void run() { 
        Map m = client.retrieve(GET("/repos/" + slug).header(
                "User-Agent", "remkop-picocli"), Map.class).blockingFirst();
        System.out.printf("%s has %s stars%n", slug, m.get("watchers"));
    }
```

Unsure if we need to add details about what dependencies are needed to run the example.

What do you think?

deining:
> How about something like this:
> 
> ```java
>     @Client("https://api.github.com")
>     @Inject RxHttpClient client; 
> 
>     @Parameters(description = "GitHub slug", paramLabel = "<owner/repo>") 
>     String slug = "remkop/picocli";
> 
>     public void run() { 
>         Map m = client.retrieve(GET("/repos/" + slug).header(
>                 "User-Agent", "remkop-picocli"), Map.class).blockingFirst();
>         System.out.printf("%s has %s stars%n", slug, m.get("watchers"));
>     }
> ```

That's a great idea. I took that over into the PR 1:1.

> Unsure if we need to add details about what dependencies are needed to run the example.

You can specify --feature `x` as option the `mn`'s create-cli-app subcommand. This is straightforward an worked for me. I added this to the PR.

> What do you think?

I think, together we could improve the sample. Thanks for your input and thanks for reviewing my PR.

remkop:
Merged, and published the rendered HTML.

Very nice. That section is certainly a lot more comprehensive now.
That reminds me that I should update the section for Quarkus: they have much better picocli support now than when I wrote that section.

PR #1236, opened 2020-10-28
Update patterns that our compiler will complain about.

nveeser-google:
Calling getClass() is likely to return an internal Proxy class (ie
com.sun.proxy.$Proxy1) so use annotationType() instead.

Assignment in an if statement is ambiguous so wrap in parens to make it
obvious that it's intended.

This are things that our compiler will fail on.


remkop:
Merged.

PR #1239, opened 2020-10-29
Implement foldable TOC for user manual and quick guide

deining:
Implemented code folding following the discussion/instructions found [here](https://github.com/asciidoctor/asciidoctor/issues/699).

### DIFF
--- a/docs/index-docinfo-footer.html
+++ b/docs/index-docinfo-footer.html
@@ -1,0 +1,32 @@
+<script src="https://cdnjs.cloudflare.com/ajax/libs/tocbot/3.0.7/tocbot.min.js"></script>
+<script>
+    /* Tocbot dynamic TOC, works with tocbot 3.0.2 */
+    var oldtoc = document.getElementById('toctitle').nextElementSibling;
+    var newtoc = document.createElement('div');
+    newtoc.setAttribute('id', 'tocbot');
+    newtoc.setAttribute('class', 'js-toc');
+    oldtoc.parentNode.replaceChild(newtoc, oldtoc);
+    tocbot.init({ contentSelector: '#content',
+        headingSelector: 'h1, h2, h3, h4',
+        smoothScroll: false });
+    var handleTocOnResize = function() {
+        var width = window.innerWidth
+                    || document.documentElement.clientWidth
+                    || document.body.clientWidth;
+        if (width < 768) {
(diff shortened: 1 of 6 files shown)


remkop:
I just saw the rendered HTML and now I understand what the foldable TOC does.
I am sorry, but I don't like the result.

I think it makes it harder for people to find things: it is not always easy to guess from the first-level heading what information will be available in the nested sub-sections. So people would need to click on each heading to see what is inside of it, and they need to remember the contents when they click on another header, because expanding another section collapses the first section.

Another reason is that picocli is competing with about [30 other libraries](https://marxsoftware.blogspot.com/2017/11/java-cmd-line-observations.html), and feedback from users shows that the reasons people like picocli is that it is full-featured and how great the manual is. The TOC plays a major part in conveying that message. By collapsing the 2nd level headers, people won't be able to see at a glance how full-featured picocli is, and how in-depth the manual is.

I appreciate the effort you put in, but I prefer not to take this change.

deining:
Thanks for your review.

> I appreciate the effort you put in, but I prefer not to take this change.

Don't worry, the effort was moderate, and I can reuse the knowledge I gained in my own documentation projects.

> By collapsing the 2nd level headers, people won't be able
> to see at a glance how full-featured picocli is, and how in-depth the manual is.

I see the point. Fortunately the used `tocbot` script offers the `collapseDepth` attribute. By setting `collapseDepth: 3`, 2nd level headers remain visibile, while subheaders will show once the user clicked the the 2nd level header link. There aren't two many 3rd level headers in the manual, so the improvement is moderate/subtle, but still present IMHO.

In your reply, you are only referring to the index only. Did you have the chance to look at the quick guide, too? The patch in the current form leaves the TOC of the quick guide as is, while adding the functionality of showing 2nd level headers once the user clicks on the first level header links. I consider this an improvement.

What do you think?

remkop:
I see, interesting. OK, let me take another look.

I had not looked at the Quick Guide, and I can see how this would be an improvement for that in that guide since we only show 1st level headers. Also interesting about the `collapseDepth` so we can show 2 levels by default in the main user guide.

I have some other work in progress that I want to finish up first, but I will take another more detailed look.
Thanks for the clarification!

deining:
I just authored and submitted an improved PR on this topic.

PR #1242, opened 2020-10-31
Foldable toc, improved

deining:


remkop:
Looks good, thanks for the improvement!
I only had time for a quick look now, will publish the rendered HTML tomorrow. 

Thank you again!

remkop:

It still takes a little getting used to 😅 but I am a lot more comfortable with having 2 levels visible in the TOC for the user manual; this is a big improvement over #1239 to me. 
For the Quick Guide it seems quite a good fit: I like the initial impression of simplicity by not showing too many subsection and sub-subsection titles at once, but that means the TOC is very sparse. Having only the focused section expand seems a good compromise.

PR #1247, opened 2020-11-02
Manual: extended Spring Boot example

deining:
This PR brings an extended Spring Boot example
- details on how to set up a skeleton Spring Boot project
- a fully working, meaningful sample app.

This comes with a price, the section is much longer now. I didn't know where to shorten without putting the Sping Boot style of the example at risk.


remkop:

True that this is a bit longer, but that is now consistent across Micronaut and Quarkus as well. 
We can reconsider this later. It is nice to have a realistic example!

PR #1248, opened 2020-11-02
Fix gradle warnings

deining:
`gradlew --warning-mode all` now does not emit any warnings any more.


remkop:
Merged.

Issue #1250, opened 2020-11-03
Inherited positional parameter gets overridden by default value if placed after subcommand


danielthegray:
If a positional parameter is defined on a root command, with an `INHERIT` scope, and with a default value, the positional parameter value is only preserved correctly if it is written *before* the subcommand. If written after the subcommand, the default value gets overwritten. This should not happen, as it has already been initialized in the call.

I have prepared a pull request with a test which showcases this failure.

danielthegray:
Another problem is that the help message itself puts the inherited positional parameter at the end, even though this is the only way that it does **NOT** work. So another option would be to move it back to before the subcommand in the usage/help message.

remkop:
Thanks for raising this!
Away from my PC now, I will look into this in more detail when I get a chance.

danielthegray:
If it's at all helpful, I'll share the findings from my debugging process:

In the case where the positional parameter is set before the subcommand, the `processArguments()` call in CommandLine.java:12547 processes it and correctly places the ArgSpec into the `initialized` set. Therefore, in the `applyDefaultValues()` call in CommandLine.java:12548 (and the condition check in CommandLIne.java:12623) the default value of the parameter is not set, since it is found in the initialized.

In the second case, the subcommand creates its own `initialized` set (in the `.parse()` call inside `processSubcommand` on CommandLine.java:12811) and thus processes the parameter value correctly, but then gets overwritten in the `applyDefaultValues()` call of the root command `parse()` call.

I initially tried to fix it by sharing one single `initialized` set between subcommand calls (by adding it as a parameter), but unfortunately, the ArgSpec object was different for inside the subcommand, so the `applyDefaultValues()` check failed and the value was overwritten anyway.

remkop:
@danielthegray Yes, that is helpful and your analysis makes a lot of sense.
I just need to find the time to sit down and work through it to get to a solution, but I have not been able to do so yet.
Maybe (hopefully) this weekend. 

Thank you again for the test case, that is very helpful. (I've marked it as `@ignored` for now to allow other people to build the project.)

remkop:
One idea is to, in addition to sharing a single `initialized` set between subcommands (by adding it as a parameter to `parse`), to also ensure that this `initialized` set contains both the actual ArgSpec object, but also, if `ArgSpec.inherited() == true`, then also to add the corresponding ArgSpec in the parent command that this ArgSpec was inherited from.

[...]

PR #1251, opened 2020-11-03
Add showcase example for issue #1250

danielthegray:
The shows two calls that should normally be equivalent, but that have different end results, due to the incorrect handling of the "initialized" parameter state.

I tried to make them share the "initialized" object, as an initial attempt to make the test pass, but unfortunately, the PositionParameterSpec object is different than the one contained by the root CommandSpec, so the initialized check fails anyway.

In any case, this test failure is useful for showing the problem.


danielthegray:
Hello @remkop, I just wanted to make sure that I communicated clearly that this is just a unit test and does not have a fix for the bug yet, since my fix didn't work.

remkop:
Understood. :+1:
Thanks for the unit test!

I still haven’t had time to look at this in a debugger yet, but I hope that the test you provided will help me quickly zoom in to the cause.

PR #1253, opened 2020-11-05
Fix line endings to LF

danielthegray:
These files had mixed line endings it seems, and were always appearing
as uncommitted changes in my environment, which makes it harder to see
the changed state of things (I'm on Linux), and also makes it impossible to do some switches between branches.


deining:
Great to see a PR that addresses this issue, since I was bothered from/struggling with this issue, too (I'm on Windows most of the time).
I hope this won't be reintroduced when regenerating the docs. I saw the issue come first two weeks ago (or so), shortly after we upgraded the asciidcotor-gradle-plugin. IIRC, it disappeared for some time and then came back again. I couldn't find any pattern and didn't have the time to investigate this, though.

danielthegray:
It might still be there, after regenerating the docs. I notice that it's always the same snippet. The asciidocs thing must be pulling this snippet from one file. I will try to make all the files in the docs folder just have Unix line endings.

danielthegray:
I've changed the entry for .html files in the `.gitattributes` file to ensure that the file is always checked in with LF line endings. I hope this fixes it. It seems that it introduced by the Asciidoctor documentation generation, it must have that faulty CSS/JS snippet inside its code somewhere, but this should hopefully keep it fixed more permanently.

remkop:
Merged.
Thanks for the contribution!

PR #1254, opened 2020-11-05
Shared initialized set

### LINKED ISSUE #1250: Inherited positional parameter gets overridden by default value if placed after subcommand

danielthegray:
This was an attempt to fix #1250 following the proposal there, but this causes other test failures. I am leaving this on a branch/PR for further examination if you think it is salvageable.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -6132,8 +6132,8 @@ public class CommandLine {
                 for (ArgSpec arg : args()) {
                     if (arg.scopeType() == ScopeType.INHERIT) {
                         subSpec.add(arg.isOption()
-                                ? OptionSpec.builder((OptionSpec) arg).inherited(true).build()
-                                : PositionalParamSpec.builder((PositionalParamSpec) arg).inherited(true).build());
+                                ? OptionSpec.builder((OptionSpec) arg).inherited(true).root(arg.root()).build()
+                                : PositionalParamSpec.builder((PositionalParamSpec) arg).inherited(true).root(arg.root()).build());
                     }
                 }
                 return this;
@@ -6260,7 +6260,7 @@ public class CommandLine {
                     Set<CommandLine> done = new HashSet<CommandLine>();
                     for (CommandLine sub : subcommands().values()) {
                         if (!done.contains(sub)) {
-                            sub.getCommandSpec().addOption(OptionSpec.builder(option).inherited(true).build());
+                            sub.getCommandSpec().addOption(OptionSpec.builder(option).inherited(true).root(option.root()).build());
                             done.add(sub);
                         }
                     }
@@ -6320,7 +6320,7 @@ public class CommandLine {
                 if (positional.scopeType() == ScopeType.INHERIT) {
                     Set<CommandLine> subCmds = new HashSet<CommandLine>(subcommands().values());// subcommands may be registered multiple times with different aliases
                     for (CommandLine sub : subCmds) {
-                        sub.getCommandSpec().addPositional(PositionalParamSpec.builder(positional).inherited(true).build());
+                        sub.getCommandSpec().addPositional(PositionalParamSpec.builder(positional).inherited(true).root(positional.root()).build());
                     }
                 }
                 return this;
@@ -6931,37 +6931,6 @@ public class CommandLine {
                 boolean result = count > 0 && count * 10 >= optionsMap().size() * 9; // at least one prefix char in common with 9 out of 10 options
                 if (tracer != null && tracer.isDebug()) {tracer.debug("'%s' %s an option: %d matching prefix chars out of %d option names%n", arg, (result ? "resembles" : "doesn't resemble"), count, optionsMap().size());}
                 return result;
-            }
-
-            public ArgSpec findOriginal(ArgSpec argSpec) {
-                if (argSpec.isOption()) {
-                    OptionSpec optionSpec = (OptionSpec) argSpec;
-                    String longestOptionName = optionSpec.longestName();
-                    for (OptionSpec potentialOriginal: options()) {
-                        if (longestOptionName.equals(potentialOriginal.longestName())) {
-                            return potentialOriginal;
-                        }
-                    }
-                } else {
-                    PositionalParamSpec paramSpec = (PositionalParamSpec) argSpec;
-                    String stringRepresentation = paramSpec.toString();
-                    String label = paramSpec.paramLabel();
-                    String descriptionKey = paramSpec.descriptionKey();
-                    String[] description = paramSpec.description();
-                    Range index = paramSpec.index();
-                    Range arity = paramSpec.arity();
-                    for (PositionalParamSpec potentialOriginal: positionalParameters()) {
-                        if (Assert.equals(stringRepresentation, potentialOriginal.toString())//
-                                && Assert.equals(label, potentialOriginal.paramLabel())//
-                                && Assert.equals(descriptionKey, potentialOriginal.descriptionKey())//
-                                && Arrays.equals(description, potentialOriginal.description()) //
-                                && Assert.equals(index, potentialOriginal.index())//
-                                && Assert.equals(arity, potentialOriginal.arity())) {
-                            return potentialOriginal;
-                        }
-                    }
-                }
-                return null;
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/picocli/CommandLine.java
```diff
@@ -6932,7 +6932,39 @@ boolean resemblesOption(String arg, Tracer tracer) {
                 if (tracer != null && tracer.isDebug()) {tracer.debug("'%s' %s an option: %d matching prefix chars out of %d option names%n", arg, (result ? "resembles" : "doesn't resemble"), count, optionsMap().size());}
                 return result;
             }
+
+            public ArgSpec findOriginal(ArgSpec argSpec) {
```
remkop, line 6936: Can we make this method package protected instead of public?

Thinking about this some more, it may be better to have a method `ArgSpec::ancestor() : ArgSpec` that returns the original `ArgSpec` for inherited args, and `null` for non-inherited args.  We would also add a getter and setter to `ArgSpec.Builder`. This property would be set in `CommandSpec::addSubcommand` when the inherited subcommand is created:

```java
// CommandSpec addSubcommand(String, CommandLine) // line 6133
if (arg.scopeType() == ScopeType.INHERIT) {
    subSpec.add(arg.isOption()
            ? OptionSpec.builder((OptionSpec) arg).inherited(true).ancestor(arg).build()
            : PositionalParamSpec.builder((PositionalParamSpec) arg).inherited(true).ancestor(arg).build());
}
```

... and similarly to `CommandSpec.addOption` and `CommandSpec.addPositional`.
This can be done separately though.
Thoughts?

remkop (changes requested):
I finally had a chance to review the PR in detail. First of all, many thanks!

With a small change, all tests pass: in `processSubcommand`, instead of passing the `initialized` set as is to the subcommand, we pass an empty list and use that to collect the args (and associated original/ancestor args) that were initialized by the subcommand. If we use this set for validation and setting defaults in the parent command, it works as intended. See my comment on that change for details.

If you make that change the PR can be merged. It would be good if we can have additional tests for named options similar to `SubcommandTest::testInheritedParameter`.

I also have an idea to improve the implementation even further: instead of doing a best-effort lookup with our `findOriginal` method, we can add a `ancestor()` method to the `ArgSpec` class that returns the original for an inherited ArgSpec (and `null` for non-inherited args). See my comment on that change for details. We can either make that improvement part of this PR if you want, or do this separately. Please let me know if you are interested in working on that.

danielthegray:
First of all, @remkop thanks a lot for the work you put into reviewing this PR, and for doing it so quickly! You have been one of the most responsive maintainers that I have interacted with, and had several great ideas! I'm in a bit of a tight stretch work-wise, I'll try to have a look at your suggestions tomorrow or the day after, but I am interested in implementing the improvements you mention.
Thanks again!

remkop:
@danielthegray That is great, thanks for helping out with this!

Thinking about some details that may come up during the implementation:
* an alternative name to `ancestor()` may be `root()`: the meaning is very similar to [`CommandSpec::root`](https://picocli.info/apidocs/picocli/CommandLine.Model.CommandSpec.html#root--), so using the same name exposes less API surface to users and makes it easier to learn
* for the top-level `ArgSpec` (the one that has `scope = INHERIT`), there is no ancestor/root, and I suggest that it returns itself rather than `null` (also similar to how `CommandSpec::root` behaves)
* for `ArgSpec` objects that are not inherited, the ancestor()/root() method should probably return `null`

src/main/java/picocli/CommandLine.java
```diff
@@ -12808,7 +12843,7 @@ private void processSubcommand(CommandLine subcommand, ParseResult.Builder build
             if (!isAnyHelpRequested() && !required.isEmpty()) { // ensure current command portion is valid
                 throw MissingParameterException.create(CommandLine.this, required, separator);
             }
-            subcommand.interpreter.parse(parsedCommands, args, originalArgs, nowProcessing, inheritedRequired);
+            subcommand.interpreter.parse(parsedCommands, args, originalArgs, nowProcessing, inheritedRequired, initialized);
```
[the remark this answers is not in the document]
danielthegray: I've implemented this change and only one test was failing, where a position parameter was creating without being linked to any command. I've added an `if` statement to exit early if it is null... I'll add a comment in that section.

remkop:
I saw you pushed a change. Do you want me to merge this, and do any further changes in a separate PR?

src/main/java/picocli/CommandLine.java
danielthegray: I like this idea very much, especially with your added suggestion to use the name `root()` which matches the API for the CommandSpec. I am happy to add it here as an additional improvement now.

remkop: Understood. I’ll hold off on merging until you give the go-ahead. 
Thanks a lot for helping out with this!

danielthegray:
No, I essentially rebased the previous work off master (now that the annoying line endings problem was gone, making it easier to switch branches without being left in an un-workable state).

I think it would be best to incorporate your solution as I think it is the most robust way to inherit parameters without needing to worry about possibly not matching the search criteria, and really fixing once and for all this default value overwriting.

I have implemented your suggestion, with a small change which I will detail in the relevant review thread, to allow for infinitely nested subcommands. Let me know what your thoughts are!

src/main/java/picocli/CommandLine.java
danielthegray: I've added the code snipped that you suggested with a small change:
```
// CommandSpec addSubcommand(String, CommandLine) // line 6133
if (arg.scopeType() == ScopeType.INHERIT) {
    subSpec.add(arg.isOption()
            ? OptionSpec.builder((OptionSpec) arg).inherited(true).ancestor(arg.root()).build()
            : PositionalParamSpec.builder((PositionalParamSpec) arg).inherited(true).ancestor(arg.root()).build());
}
```
since if we just pass on the current one, we will face the same problem again if we are nesting more than one level deep (which is my case). There is only one positional parameter that is being inherited from the root, but it would still be a problem I think, and thanks to your spot-on definition of the `.root()` call pointing to the `ArgSpec` itself if it is at the root of the tree, it will always work (along with a small check in the initialization of the `ArgSpec` to set the `root` field to itself if it is coming from the `Builder<T>` as `null` and has the scope set to INHERIT. Let me know what you think!

danielthegray: I'll mark this as resolved, but I'm open to other comments to other parts of the code if something isn't 100% in line with the project's practices or styles.

danielthegray:
> If you make that change the PR can be merged. It would be good if we can have additional tests for named options similar to SubcommandTest::testInheritedParameter.

I missed this during my last check, I will add these as well.

remkop:
Merged with some minor changes.
Thank you very much for the contribution!

danielthegray:
@remkop Thank you very much! That was a great catch, that the `.root(original.root())` is not needed. I have removed a javadoc comment that I added to that effect, which would be good to merge in as well, I think.

remkop:
Ok, can you paste a diff/patch of your change here? Since this PR has already been merged I don't think it is possible to merge any subsequent changes.

danielthegray:
Sure!
```diff
diff --git a/src/main/java/picocli/CommandLine.java b/src/main/java/picocli/CommandLine.java
index 02dd44c2..ca21e508 100644
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -9060,7 +9060,6 @@ public class CommandLine {
                 public T hidden(boolean hidden)              { this.hidden = hidden; return self(); }
 
                 /** Sets whether this option is inherited from a parent command, and returns this builder.
-                 * <b>Do not forget to also set {@link #root()} when setting this, to ensure that it always can track back to the root element.</b>
                  * @since 4.3.0 */
                 public T inherited(boolean inherited)        { this.inherited = inherited; return self(); }

```

remkop:
Merged. Thank you for the attention to details!

danielthegray:
Thank you again for getting this merged so quickly!

PR #1255, opened 2020-11-05
Quick guide: add Kotlin + Scala versions

deining:
This PR adds Kotlin + Scala versions to the subcommands example of the quick guide.
It adds Groovy versions to sample code in chapter 17 'Subcommands' of the quick guide.

It contains some minor fixes to the gradle build scripts of the example projects:
- avoid deprecation warnings
- bump to latest version of the dependencies, like ascii-doctor-plugin, ... 

It also fixes deprecated syntax in gradle build scripts in a few README files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,7 +329,7 @@ See the [source code](https://github.com/remkop/picocli/blob/master/src/main/jav
 
 ### Gradle
 ```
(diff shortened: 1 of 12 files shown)


remkop:
Thank you for the contribution! It is getting better and better!!

PR #1265, opened 2020-11-23
Allow users of picocli-shell-jline3 to use `clear` command and a nicer `help` command

sualeh:
Allow users of picocli-shell-jline3 to use `clear` command and a nicer `help` command.

- The clear screen command is provided out of the box by jline2, and similarly should be built into `picocli.shell.jline3.PicocliCommands`. This change allows users of `PicocliCommands` to include a clear screen command with a single line: `picocliCommands.includeClearScreenCommand(reader);` - see the example.
- The help command is provided by jline3 by default, but picocli help is much nicer. This change allows users of `PicocliCommands` to override the jline3 help command with a single line: `systemRegistry.register("help", picocliCommands);` - see the example.

### DIFF
--- a/picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
+++ b/picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
@@ -22,6 +22,8 @@ import org.jline.reader.Completer;
 import org.jline.reader.LineReader;
 import org.jline.reader.ParsedLine;
 import org.jline.reader.impl.LineReaderImpl;
+import org.jline.reader.impl.completer.ArgumentCompleter;
+import org.jline.reader.impl.completer.NullCompleter;
 import org.jline.reader.impl.completer.SystemCompleter;
 import org.jline.utils.AttributedString;
 
@@ -110,9 +112,9 @@ public class PicocliCommands implements CommandRegistry {
         return out;
     }
 
-    private class PicocliCompleter implements Completer {
+    private class PicocliCompleter extends ArgumentCompleter implements Completer {
 
-        public PicocliCompleter() {}
+        public PicocliCompleter() { super(NullCompleter.INSTANCE); }
 
         @Override
         public void complete(LineReader reader, ParsedLine commandLine, List<Candidate> candidates) {
--- a/picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
+++ b/picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
@@ -1,7 +1,15 @@
 package picocli.shell.jline3;
 
+import java.io.IOException;
 import java.nio.file.Path;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.Callable;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
@@ -13,18 +21,15 @@ import org.jline.reader.Candidate;
 import org.jline.reader.Completer;
 import org.jline.reader.LineReader;
 import org.jline.reader.ParsedLine;
+import org.jline.reader.impl.LineReaderImpl;
 import org.jline.reader.impl.completer.SystemCompleter;
 import org.jline.utils.AttributedString;
+
 import picocli.CommandLine;
+import picocli.CommandLine.Command;
 import picocli.CommandLine.Help;
 import picocli.CommandLine.Model.CommandSpec;
 import picocli.CommandLine.Model.OptionSpec;
-
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
 
 /**
  * Compiles SystemCompleter for command completion and implements a method commandDescription() that provides command descriptions
@@ -35,6 +40,26 @@ import java.util.Map;
  * @since 4.1.2
  */
 public class PicocliCommands implements CommandRegistry {
+
+    /**
+     * Command that clears the screen.
+     */
+    @Command(name = "cls", aliases = "clear", mixinStandardHelpOptions = true,
+            description = "Clears the screen", version = "1.0")
+    static class ClearScreen implements Callable<Void> {
+
+    	private final LineReaderImpl reader;
+
+        ClearScreen(LineReaderImpl reader) {
+			this.reader = reader;
+		}
+
+		public Void call() throws IOException {
+            reader.clearScreen();
+            return null;
+        }
+    }
+
     private final Supplier<Path> workDir;
     private final CommandLine cmd;
     private final Set<String> commands;
@@ -47,12 +72,23 @@ public class PicocliCommands implements CommandRegistry {
     public PicocliCommands(Supplier<Path> workDir, CommandLine cmd) {
         this.workDir = workDir;
         this.cmd = cmd;
-        commands = cmd.getCommandSpec().subcommands().keySet();
+        commands = new HashSet<>(cmd.getCommandSpec().subcommands().keySet());
         for (String c: commands) {
             for (String a: cmd.getSubcommands().get(c).getCommandSpec().aliases()) {
                 aliasCommand.put(a, c);
             }
         }
+    }
+
+    public void includeClearScreenCommand(LineReader reader) {
+    	if (reader == null) return;
+    	ClearScreen clearScreen = new ClearScreen((LineReaderImpl) reader);
+    	cmd.addSubcommand(clearScreen);
+
+    	commands.add("clear");
+
+    	aliasCommand.put("clear", "clear");
+    	aliasCommand.put("cls", "clear");
     }
 
     /**
--- a/picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
+++ b/picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
@@ -1,7 +1,6 @@
 package picocli.shell.jline3;
 
 import java.io.IOException;
-import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -9,7 +8,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.Callable;
-import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import org.jline.builtins.Options.HelpException;
@@ -123,17 +121,11 @@ public class PicocliCommands implements CommandRegistry {
         }
     }
 
-    private final Supplier<Path> workDir;
     private final CommandLine cmd;
     private final Set<String> commands;
     private final Map<String,String> aliasCommand = new HashMap<>();
 
-    public PicocliCommands(Path workDir, CommandLine cmd) {
-        this(() -> workDir, cmd);
-    }
-
-    public PicocliCommands(Supplier<Path> workDir, CommandLine cmd) {
-        this.workDir = workDir;
+    public PicocliCommands(CommandLine cmd) {
         this.cmd = cmd;
         commands = cmd.getCommandSpec().subcommands().keySet();
         for (String c: commands) {
--- a/picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
+++ b/picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
@@ -5,7 +5,6 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


sualeh:
@remkop - any thoughts on this?

picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
```diff
@@ -47,14 +74,25 @@ public PicocliCommands(Path workDir, CommandLine cmd) {
...
+    	if (reader == null) return;
+    	ClearScreen clearScreen = new ClearScreen((LineReaderImpl) reader);
+    	cmd.addSubcommand(clearScreen);
+
+    	commands.add("clear");
+
+    	aliasCommand.put("clear", "clear");
+    	aliasCommand.put("cls", "clear");
+    }
+
```
remkop, line 95: It is a bit unfortunate that we cannot do this declaratively, like `@Command(subcommands = { (...), ClearScreen.class, (...)}` in current master.

Is this necessary because we need to set the `LineReaderImpl`, or are there other reasons as well?

See also my suggestion below for installing the ClearScreen command as an alternative to `picocliCommands.includeClearScreenCommand(reader);`

picocli-shell-jline3/src/test/java/picocli/shell/jline3/example/Example.java
```diff
@@ -172,7 +152,7 @@ public static void main(String[] args) {
                         .variable(LineReader.LIST_MAX, 50)   // max tab completion candidates
                         .build();
                 builtins.setLineReader(reader);
-                commands.setReader(reader);
+                picocliCommands.includeClearScreenCommand(reader);
```
remkop, line 155: If I understand correctly, the above line creates a `ClearCommand` with the specified `LineReader`, and adds it to the command hierarchy. 

I would prefer if we could add the command declaratively with `@Command(subcommands = ClearCommand.class)`. To do that, we need some way to pass the `LineReader` to the `ClearCommand` (without relying on `@ParentCommand`).

One idea is to do it with a custom factory like this:

```java
LineReader lineReader = ... // create line reader
IFactory factory = new IFactory() {
    public <K> K create(Class<K> clazz) throws Exception {
        if (ClearCommand.class == clazz) { return new ClearCommand(lineReader); }
        return CommandLine.defaultFactory().create(clazz); 
    }
}
CommandLine cmd = new CommandLine(commands, factory);
...
```

Thoughts?

remkop (changes requested):
@sualeh I finally had time to look at this in more detail.

I have an idea for keeping things more declarative, can you take a look and let me know your thoughts?

picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
```diff
@@ -35,6 +42,26 @@
...
+            description = "Clears the screen", version = "1.0")
+    static class ClearScreen implements Callable<Void> {
+
+    	private final LineReaderImpl reader;
+
+        ClearScreen(LineReaderImpl reader) {
+			this.reader = reader;
+		}
+
+		public Void call() throws IOException {
```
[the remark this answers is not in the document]
sualeh: Done

picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
```diff
@@ -47,14 +74,25 @@ public PicocliCommands(Path workDir, CommandLine cmd) {
     public PicocliCommands(Supplier<Path> workDir, CommandLine cmd) {
         this.workDir = workDir;
         this.cmd = cmd;
-        commands = cmd.getCommandSpec().subcommands().keySet();
+        commands = new HashSet<>(cmd.getCommandSpec().subcommands().keySet());
```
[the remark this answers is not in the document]
sualeh: We modify it to add the clear command - on https://github.com/sualeh/picocli/blob/master/picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java#L90

picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
```diff
@@ -74,9 +112,9 @@ public SystemCompleter compileCompleters() {
         return out;
     }
 
-    private class PicocliCompleter implements Completer {
+    private class PicocliCompleter extends ArgumentCompleter implements Completer {
```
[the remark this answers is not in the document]
sualeh: jline3 asks for a Completer, but when you do not supply an ArgumentCompleter, it throws a null pointer exception. Go figure.

picocli-shell-jline3/src/test/java/picocli/shell/jline3/example/Example.java
```diff
@@ -164,6 +143,7 @@ public static void main(String[] args) {
             try (Terminal terminal = TerminalBuilder.builder().build()) {
                 SystemRegistry systemRegistry = new SystemRegistryImpl(parser, terminal, Example::workDir, null);
                 systemRegistry.setCommandRegistries(builtins, picocliCommands);
+                systemRegistry.register("help", picocliCommands);
```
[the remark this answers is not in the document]
sualeh: This overrides the help command from jline3 system commands (which is pretty crappy) with picocli's help command, which is much better.

picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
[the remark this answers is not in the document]
remkop: Ah yes, I overlooked that. Thank you. But perhaps that can be avoided with my factory suggestion...

picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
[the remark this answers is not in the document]
remkop: We should probably report this to the JLine maintainers so they can fix it.

picocli-shell-jline3/src/main/java/picocli/shell/jline3/PicocliCommands.java
sualeh: Please see the comment below.

picocli-shell-jline3/src/test/java/picocli/shell/jline3/example/Example.java
sualeh: @remkop - this is a better way to use the picocli API, of course. However there is a circular dependency in how these are created:
line reader -> system registry completer -> picocli commands -> line reader
In order to break this cyclic relationship, we need to construct the clear command later on, and pass the line reader in. Can you think of another way to break the cycle?

remkop: Does this work?

```java
// set up picocli commands
class MyFactory implements CommandLine.IFactory {
    LineReader reader;
    public <K> K create(Class<K> clazz) throws Exception {
        if (ClearCommand.class == clazz) { return new ClearCommand(reader); }
        return CommandLine.defaultFactory().create(clazz);
    }
}
MyFactory myFactory = new MyFactory();
CliCommands commands = new CliCommands();
CommandLine cmd = new CommandLine(commands, myFactory);
PicocliCommands picocliCommands = new PicocliCommands(Example::workDir, cmd);

Parser parser = new DefaultParser();
try (Terminal terminal = TerminalBuilder.builder().build()) {
    SystemRegistry systemRegistry = new SystemRegistryImpl(parser, terminal, Example::workDir, null);
    systemRegistry.setCommandRegistries(builtins, picocliCommands);

    LineReader reader = LineReaderBuilder.builder()
            .terminal(terminal)
            .completer(systemRegistry.completer())
            .parser(parser)
            .variable(LineReader.LIST_MAX, 50)  
            .build();
    myFactory.reader = reader; // now pass the reader to the factory
```

sualeh: @remkop - I have moved the `ClearCommand` class to `PicocliCommands`, so the users of picocli do not have to deal with it. The idea is that a user can turn on the clear command with one method call on `PicocliCommands` (and similarly turn on picocli help with one method call).

remkop: ok, perhaps the factory should also live in PicocliCommands, so it is more in one place.

sualeh: Sure, but what if someone uses their own factory? (I use a custom factory to maintain state in the jline3 shell. The factory injects the same state object instance into each command.) If we are going this route, picocli would need to allow a factory chain. Maybe that is a feature request for picocli itself?

remkop: Okay, so it should be something like:

```java
class MyFactory implements CommandLine.IFactory {
    LineReader reader;
    CommandLine.IFactory fallbackFactory;

    public MyFactory() {
        this(CommandLine.defaultFactory());
    }

    public MyFactory(CommandLine.IFactory fallbackFactory) {
        this.fallbackFactory = Objects.requireNonNull(fallbackFactory);
    }

    public <K> K create(Class<K> clazz) throws Exception {
        if (ClearCommand.class == clazz) { return new ClearCommand(reader); }
        return fallbackFactory.create(clazz);
    }
}
```

sualeh: Thanks.

I will need to modify the `PicocliCommands` constructor (maybe add a boolean argument) , and add in the clear command to the passed in command-line object. If I am going to modify the passed in `CommandLine` object anyway, I can wrap in the factory with `MyFactory` internally in the `PicocliCommands` constructor. In order to do this, I would need a getter to get the original factory on the original `CommandLine`. (I could not find this getter on `CommandLine`.)

This is all probably making it more convoluted than my original solution. What do you think?

sualeh: PS: Please consider making the factory follow a chain of responsibility design pattern in a future release. Would you like me to create an issue for this?

remkop: > I will need to modify the PicocliCommands constructor (maybe add a boolean argument) and add in the clear command to the passed in command-line object

Why? I was thinking of just moving the factory to the `PicocliCommands` class. Something like this:

```java
PicocliCommands.Factory myFactory = new PicocliCommands.Factory();
CliCommands commands = new CliCommands();
CommandLine cmd = new CommandLine(commands, myFactory);
PicocliCommands picocliCommands = new PicocliCommands(Example::workDir, cmd);

Parser parser = new DefaultParser();
try (Terminal terminal = TerminalBuilder.builder().build()) {
    SystemRegistry systemRegistry = new SystemRegistryImpl(parser, terminal, Example::workDir, null);
    systemRegistry.setCommandRegistries(builtins, picocliCommands);

    LineReader reader = LineReaderBuilder.builder()
            .terminal(terminal)
            .completer(systemRegistry.completer())
            .parser(parser)
            .variable(LineReader.LIST_MAX, 50)  
            .build();
    myFactory.setReader(reader); // now pass the reader to the factory
```


> PS: Please consider making the factory follow a chain of responsibility design pattern in a future release. Would you like me to create an issue for this?

Factory implementors can easily implement their own chain of responsibility if they want. One simple way to do this is with a constructor that accepts another factory as in the code snippet I posted in my previous comment. I do not see what value a chain of responsibility would add, but perhaps I am missing something.

remkop: PS. I created https://github.com/remkop/picocli/issues/1267 for creating an accessor for the factory. We don't need it for our use case here, but it is nice to have.

remkop: @sualeh I saw your message on the picocli Google group, but I think there may be a misunderstanding. 
I think we are almost there.

I did not understand your comment:

> I will need to modify the `PicocliCommands` constructor (maybe add a boolean argument) , and add in the clear command to the passed in command-line object. If I am going to modify the passed in `CommandLine` object anyway, I can wrap in the factory with `MyFactory` internally in the `PicocliCommands` constructor. In order to do this, I would need a getter to get the original factory on the original `CommandLine`. (I could not find this getter on `CommandLine`.)

Perhaps I misunderstand, but I think we can leave the `PicocliCommands` constructor as is, or am I missing something?

About the factory, and combining it with a custom factory, I imagine applications would use it like this:

```java
MyCustomFactory customFactory = createCustomFactory(); // your application custom factory
PicocliCommands.Factory factory = new PicocliCommands.Factory(customFactory); // chain the factories
CliCommands commands = new CliCommands();
CommandLine cmd = new CommandLine(commands, factory);
...
factory.setReader(reader); // set LineReader after it is created
```

This way, we no longer need the `PicocliCommands.includeClearScreenCommand` method, and applications can use the new ClearScreen command declaratively with `@Command(subcommands = { PicocliCommands.ClearScreen.class, ...})`.

This looks nice to me, but what do you think?

sualeh:
@remkop  - thanks for the last snippet. I think I have finally understood (sorry for being dense), and pushed what I think are the final changes. Please take a look. Thanks.

remkop:
Great! Let me check when I get to my PC.

remkop (changes requested):
Thank you for the changes, @sualeh, this looks pretty good now!

For finishing touches, can you add some javadoc?
Also I believe the factory and the clear-screen subcommand should require a `LineReaderImpl`; this is not great, but we cannot call `clearScreen` on any `LineReader`. This is also something we should probably raise with the JLine maintainers: they should consider adding the `clearScreen` method to the `LineReader` interface.

sualeh (commented):
@remkop - I have made the clean-up changes you suggested. 
- I want to remove `workDir` from `PicocliCommands`. Please let me know if you need a separate pull request for that, or if I can include it in the same PR. 
- I will test the suggestion in https://github.com/jline/jline3/issues/610 and see if that works - then we can go back to using `LineReader`

remkop:
I'm fine with removing workDir (thanks! :-)), no need for a separate pull request. 

Happy to use the LineReader interface if possible, let me know how it goes.

sualeh:
@remkop - I think we are done. I have removed `workDir`, and the clear screen is not working - see my comments on https://github.com/jline/jline3/issues/610

remkop (approved):
This looks good, thank you!

I will merge this soon.
This will be included in the upcoming picocli 4.6 release.

remkop (approved):
Ah very nice!
Using a public API like Terminal is a lot nicer than having to use internal API like LineReaderImpl.
Thanks for following up!

remkop:
Merged.

Issue #1259, opened 2020-11-13
Consider providing API for modifying the model at initialization


remkop:
This is a follow-up ticket from discussion on #1256 and #1217.

The use case raised by @MarkoMackic is as follows (Marko, please comment if I missed anything):

* At model construction time, we want the ability to modify the model, based on some runtime environment information (system properties, environment variables, configuration files, etc.). One example is the ability to include or exclude some subcommand based on a system property.
* We prefer to use the annotation API because of its ease of use, rather than constructing the model using the programmatic API.
* We want to do this modification at model _construction/initialization time_, so that this modified model can be used by other components like generating man pages and generating shell completion scripts.

My initial reaction is that this is already possible with code like this:

```java
static CommandLine createDynamicallyModifiedCommand() {
    CommandLine result = new CommandLine(new MyApp());
    if (someCondition()) { // could be a system property or something else
        result.getCommandSpec().subcommands().get("a")
                 .getCommandSpec().subcommands().get("b")
                 .removeSubcommand("c");
    }
    return result;
}

// runs MyApp
public static void main(String... args) {
    System.exit(createDynamicallyModifiedCommand().execute(args));
}

// generates man pages for MyApp
static class MyManPageGenerator implements Callable<Integer> {
        @Option(names = {"-d", "--outdir"}, defaultValue = ".", paramLabel = "<outdir>",
                description = {"Output directory to write the generated AsciiDoc files to. " +
                        "If not specified, files are written to the current directory."})
        File outdir;

        @Option(names = {"-v", "--verbose"},
                description = {
                        "Specify multiple -v options to increase verbosity.",
                        "For example, `-v -v -v` or `-vvv`"})
        boolean[] verbosity = new boolean[0];

    public Integer call() throws Exception {
        CommandSpec spec = createDynamicallyModifiedCommand().getCommandSpec();
        ManPageGenerator.generateManPage(outdir, null, verbosity, false, spec);
    }

    public static void main(String... args) throws Exception {
        new CommandLine(new MyManPageGenerator()).execute(args);
    }
}
```

@MarkoMackic can you clarify why the above is unsatisfactory and what you would like to see instead?

MarkoMackic:
[...]

PR #1266, opened 2020-11-26
Support user-defined model transformations after initialization and before parsing

MarkoMackic:
Proposes API for #1259

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -4682,7 +4682,7 @@ public class CommandLine {
 
         /** Returns transformer for command
          * @since 4.6 */
-        Class<? extends IModelTransformer> modelTransformer() default NoOpModelTransforemer.class;
+        Class<? extends IModelTransformer> modelTransformer() default NoOpModelTransformer.class;
     }
     /** A {@code Command} may define one or more {@code ArgGroups}: a group of options, positional parameters or a mixture of the two.
      * Groups can be used to:
@@ -4834,7 +4834,7 @@ public class CommandLine {
         CommandSpec transform(CommandSpec commandSpec);
     }
 
-    private static class NoOpModelTransforemer implements IModelTransformer {
+    private static class NoOpModelTransformer implements IModelTransformer {
         public CommandSpec transform(CommandSpec commandSpec) { return commandSpec; }
     }
 
@@ -7008,7 +7008,7 @@ public class CommandLine {
             void initExitCodeOnExecutionException(int exitCode) { if (initializable(exitCodeOnExecutionException, exitCode, ExitCode.SOFTWARE)) { exitCodeOnExecutionException = exitCode; } }
             void updateName(String value)                   { if (isNonDefault(value, DEFAULT_COMMAND_NAME))                {name = value;} }
             void updateModelTransformer(Class<? extends IModelTransformer> value) {
-                if (isNonDefault(value, NoOpModelTransforemer.class)) { this.modelTransformer = DefaultFactory.create(defaultFactory(), value); }
+                if (isNonDefault(value, NoOpModelTransformer.class)) { this.modelTransformer = DefaultFactory.create(defaultFactory(), value); }
             }
             void updateHelpCommand(boolean value)           { if (isNonDefault(value, DEFAULT_IS_HELP_COMMAND))             {isHelpCommand = value;} }
             void updateSubcommandsRepeatable(boolean value) { if (isNonDefault(value, DEFAULT_SUBCOMMANDS_REPEATABLE))      {subcommandsRepeatable = value;} }
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -4681,7 +4681,7 @@ public class CommandLine {
 
 
         /** Returns transformer for command
-         * @since X.X */
+         * @since 4.6 */
         Class<? extends IModelTransformer> modelTransformer() default NoOpModelTransforemer.class;
     }
     /** A {@code Command} may define one or more {@code ArgGroups}: a group of options, positional parameters or a mixture of the two.
@@ -4825,7 +4825,7 @@ public class CommandLine {
     /**
      * Provides a way to modify how command model is built. Commands may configure a model transformer using
      * {@link Command#modelTransformer()} annotation attribute.
-     * @since X.X */
+     * @since 4.6 */
     public interface IModelTransformer {
         /**
          * Returns CommandSpec after doing transformation.
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -4662,9 +4662,9 @@ public class CommandLine {
         ScopeType scope() default ScopeType.LOCAL;
 
 
-        /** Returns preprocessor for command spec
-         * @since 4.6 */
-        Class<? extends IModelTransformer> modelTransformer() default NoOpPreprocessor.class;
+        /** Returns transformer for command
+         * @since X.X */
+        Class<? extends IModelTransformer> modelTransformer() default NoOpModelTransforemer.class;
     }
     /** A {@code Command} may define one or more {@code ArgGroups}: a group of options, positional parameters or a mixture of the two.
      * Groups can be used to:
@@ -4805,18 +4805,18 @@ public class CommandLine {
     }
 
     /**
-     * Provides a way to modify how command model is built. Commands may configure a preprocessor using
+     * Provides a way to modify how command model is built. Commands may configure a model transformer using
      * {@link Command#modelTransformer()} annotation attribute.
      * @since X.X */
     public interface IModelTransformer {
         /**
-         * Returns CommandSpec after doing preprocessing.
-         * @return preprocessed CommandSpec
+         * Returns CommandSpec after doing transformation.
+         * @return transformed CommandSpec
          */
         CommandSpec transform(CommandSpec commandSpec);
     }
 
-    private static class NoOpPreprocessor implements IModelTransformer {
+    private static class NoOpModelTransforemer implements IModelTransformer {
         public CommandSpec transform(CommandSpec commandSpec) { return commandSpec; }
     }
 
@@ -11082,7 +11082,7 @@ public class CommandLine {
                 result.updateArgSpecMessages();
 
                 if (annotationsAreMandatory) {validateCommandSpec(result, cmd != null || hasCommandAnnotation, userObject.toString()); }
-                result = DefaultFactory.create(defaultFactory(), cmd == null  ? NoOpPreprocessor.class :  cmd.modelTransformer()).transform(result);
+                result = DefaultFactory.create(defaultFactory(), cmd == null  ? NoOpModelTransforemer.class :  cmd.modelTransformer()).transform(result);
                 result.validate();
                 return result;
             }
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -217,18 +217,18 @@ public class CommandLine {
      * @param factory the factory used to create instances of {@linkplain Command#subcommands() subcommands}, {@linkplain Option#converter() converters}, etc., that are registered declaratively with annotation attributes
      * @throws InitializationException if the specified command object does not have a {@link Command}, {@link Option} or {@link Parameters} annotation
      * @since 2.2 */
+    public CommandLine(Object command, IFactory factory) {
+        this(command, factory, true);
+    }
+
     private CommandLine(Object command, IFactory factory, boolean userCalled) {
         this.factory = Assert.notNull(factory, "factory");
         interpreter = new Interpreter();
         commandSpec = CommandSpec.forAnnotatedObject(command, factory);
         commandSpec.commandLine(this);
-        if(userCalled) { this.applyModelTransformations(); }
+        if (userCalled) { this.applyModelTransformations(); }
         commandSpec.validate();
         if (commandSpec.unmatchedArgsBindings().size() > 0) { setUnmatchedArgumentsAllowed(true); }
-    }
-
-    public CommandLine(Object command, IFactory factory) {
-        this(command, factory, true);
     }
 
     private CommandLine copy() {
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -6985,9 +6985,8 @@ public class CommandLine {
 
                 updateScopeType(cmd.scope());
 
-                updateModelTransformer(cmd.modelTransformer());
-
                 if (factory != null) {
+                    updateModelTransformer(cmd.modelTransformer());
                     updateVersionProvider(cmd.versionProvider(), factory);
                     initDefaultValueProvider(cmd.defaultValueProvider(), factory);
                 }
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -148,7 +148,7 @@ public class CommandLine {
     public static final String VERSION = "4.5.3-SNAPSHOT";
 
     private final Tracer tracer = new Tracer();
-    private final CommandSpec commandSpec;
+    private CommandSpec commandSpec;
     private final Interpreter interpreter;
     private final IFactory factory;
 
@@ -217,13 +217,18 @@ public class CommandLine {
      * @param factory the factory used to create instances of {@linkplain Command#subcommands() subcommands}, {@linkplain Option#converter() converters}, etc., that are registered declaratively with annotation attributes
      * @throws InitializationException if the specified command object does not have a {@link Command}, {@link Option} or {@link Parameters} annotation
      * @since 2.2 */
-    public CommandLine(Object command, IFactory factory) {
+    private CommandLine(Object command, IFactory factory, boolean userCalled) {
         this.factory = Assert.notNull(factory, "factory");
         interpreter = new Interpreter();
         commandSpec = CommandSpec.forAnnotatedObject(command, factory);
         commandSpec.commandLine(this);
+        if(userCalled) { this.applyModelTransformations(); }
         commandSpec.validate();
         if (commandSpec.unmatchedArgsBindings().size() > 0) { setUnmatchedArgumentsAllowed(true); }
+    }
+
+    public CommandLine(Object command, IFactory factory) {
+        this(command, factory, true);
     }
 
     private CommandLine copy() {
@@ -239,6 +244,17 @@ public class CommandLine {
         result.interpreter.converterRegistry.clear();
         result.interpreter.converterRegistry.putAll(interpreter.converterRegistry);
         return result;
+    }
+
+
+    protected void applyModelTransformations() {
+        if (this.commandSpec.modelTransformer != null) {
+            this.commandSpec = this.commandSpec.modelTransformer.transform(this.commandSpec);
+        }
+        for (CommandLine cmd : this.getSubcommands().values()) {
+
+            cmd.applyModelTransformations();
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


remkop (changes requested):
From a functionality perspective, this proposal seems too limited:
all that any preprocessor can do is prevent a subcommand from being added.

What if there are applications that want to dynamically add a subcommand, or add an option, or remove a subcommand from the parent and add it to a different subcommand, or make any other change to the model? 

From a design perspective, I prefer interfaces to classes, and ideally an interface with just a single abstract method (to allow the use of lambdas and Groovy closures in various scenarios.)

MarkoMackic:
Let's do a more generic one  :)

MarkoMackic:
@remkop What do you think about this approach ?

MarkoMackic:
There is one test failing :) I'll fix that

src/test/java/picocli/CommandSubcommandFilterTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Rule
+    public final SystemErrRule systemErrRule = new SystemErrRule().enableLog().muteForSuccessfulTests();
+
+    @Rule
+    public final SystemOutRule systemOutRule = new SystemOutRule().enableLog().muteForSuccessfulTests();
+
+    @CommandLine.Command(name = "aa")
+    private static class AA implements Callable<Integer>
+    {
```
remkop, line 31: The code style I use is opening bracket on the same line, separated by a space.

src/test/java/picocli/CommandSubcommandFilterTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+        private static class AFilter implements CommandLine.IPreprocessor
+        {
+            public CommandLine.Model.CommandSpec preprocess(CommandLine.Model.CommandSpec commandSpec)
+            {
+                // verify it's context aware
+                assertEquals(2, commandSpec.subcommands().size());
+
+                if(Boolean.valueOf(System.getProperty("disable_aa"))) commandSpec.removeSubcommand("aa");
```
remkop, line 69: The code style I use is space between the `if` and the opening bracket `(`, and redundant braces for the statement following the conditional.

src/main/java/picocli/CommandLine.java
```diff
@@ -11059,7 +11081,8 @@ static CommandSpec extractCommandSpec(Object command, IFactory factory, boolean
 
                 result.updateArgSpecMessages();
 
-                if (annotationsAreMandatory) {validateCommandSpec(result, hasCommandAnnotation, userObject.toString()); }
+                if (annotationsAreMandatory) {validateCommandSpec(result, cmd != null || hasCommandAnnotation, userObject.toString()); }
+                result = DefaultFactory.create(defaultFactory(), cmd == null  ? NoOpPreprocessor.class :  cmd.preprocessor()).preprocess(result);
```
remkop, line 11085: Just thinking out loud: perhaps we should first construct the full command hierarchy, and then have a second pass over the hierarchy (recursively visiting subcommands and sub-subcommands) to apply the preprocessor?

That way the preprocessor can operate on the full hierarchy and there are no timing issues where the preprocessor is run before the hierarchy has been fully constructed.

src/main/java/picocli/CommandLine.java
```diff
@@ -4799,6 +4804,22 @@ private static String format(String formatString, Object... params) {
...
+    /**
+     * Provides a way to modify how command model is built. Commands may configure a preprocessor using
+     * {@link Command#preprocessor()} annotation attribute.
+     * @since X.X */
+    public interface IPreprocessor {
+        /**
+         * Returns CommandSpec after doing preprocessing.
+         * @return preprocessed CommandSpec
+         */
+        CommandSpec preprocess(CommandSpec commandSpec);
```
remkop, line 4816: I like this interface signature. It allows for very flexible usage.

remkop (changes requested):
Thank you for the update, I like the new interface signature very much.
I had some feedback, can you take a look?

On a separate note, I am starting to think perhaps there is a better name than "preprocessor"... 

I don't have any ideas yet, open to suggestions.
I am not sure if the name should emphasize the timing, the fact that this is something that is applied between constructing the model and using the model to parse command line args. Or whether the name should emphasize that this is something that allows applications to modify/transform the model.
Still thinking, suggestions welcome!

src/main/java/picocli/CommandLine.java
MarkoMackic: Well I think that's not really necessery, where do you see this more fit ? What kind of timing issues ? I'm open to suggestions, if there is any appearent improvement :)

MarkoMackic:
I'll fix the codestyle. We could call it `modelPreprocessor`. That name empasizes it's done at model build time.

remkop:
`IModelTransformer`?

MarkoMackic:
That makes sense. I'll refactor tomorrow evening.

src/main/java/picocli/CommandLine.java
remkop: Oops, I only just saw your comment now.
What I meant was, the current proposal applies the transformation immediately after the command was created, so while the construction of the full hierarchy is still in progress. Imagine we have this hierarchy:

```
Top
  - Sub1
    - Sub1Sub1
    - Sub1Sub2
  - Sub2
    -Sub2Sub1
    -Sub2Sub2
```

The transformation for `Sub1Sub1` will be applied before `Sub1Sub2` was created, and even before `Sub2` and its subcommands are created. So if the transformation for  `Sub1Sub1` wanted to do something with  `Sub1Sub2`, that is currently not possible.

The current implementation does not allow some transformations, because the command hierarchy has not been fully constructed yet.

 I think it would be better if transformations were applied after the command hierarchy has been completely constructed.

src/main/java/picocli/CommandLine.java
```diff
@@ -4799,6 +4804,22 @@ private static String format(String formatString, Object... params) {
...
+     * @since X.X */
+    public interface IModelTransformer {
+        /**
+         * Returns CommandSpec after doing preprocessing.
+         * @return preprocessed CommandSpec
+         */
+        CommandSpec transform(CommandSpec commandSpec);
+    }
+
+    private static class NoOpPreprocessor implements IModelTransformer {
```
remkop, line 4819: Let's make this NoOpModelTransformer too.

MarkoMackic: Sure.

src/main/java/picocli/CommandLine.java
MarkoMackic: @remkop I don't think it's a good practice. This is not localized . And all preprocessing/transformation depends only on external state, so if there are common things, they could be unified by parent transformer class. The transformer doesn't have to be local to the subcommand.

MarkoMackic:
@remkop Please replace the X.X in this branch with the release this is going to get into. And read the comment about this after build transformation.

MarkoMackic:
And also PR title should be changed, because the functionality impact is wider, but I don't have good words to define it

src/main/java/picocli/CommandLine.java
remkop: @MarkoMackic well, we don't want to fail some people's transformations because they do things differently than we think they should... I think the transformation should not be applied to a partial model. Transformations should be able to trust that the model is complete when the `transform` method is called.

MarkoMackic: @remkop ok, let's try it, f.y.i, it shouldn't affect current API ( it should be still context aware ), but I can make it traverse the model after it's built. But there are issues with that, sp let's first communicate those. Let's take your tree for example :
```
Top
  - Sub1
    - Sub1Sub1
    - Sub1Sub2
  - Sub2
    -Sub2Sub1
    -Sub2Sub2
```

So let's say we define 2 transformers 1 for Sub1 and second for Sub2,  Sub1 transformer moves subcommand Sub2Sub1 ( eg. Sub1 transformer would look something like this ( pseudo code ) `spec.getParentCommand().subcommands("Sub2").subcommands("Sub2Sub1").moveTo(spec)` , and than you have transformer in Sub2 that says `spec.removeSubcommand("Sub2Sub1")`. They are clashing, 2 ways possible, if Sub1 transformer is executed first, the second one will fail, , and the other way around  , and it's up to library user to do that sensitive handling. Operating on partial model is much more stable, your model preprocessor can rest assured that hierarchy down the tree from what you're processing is 100% built, can't operate on parallel or branches up the tree. I think this is just more secure approach, and there are no API limitations, just some rules are enforced, ( parent can only process it's children and itself )

MarkoMackic: I see a couple of advantages over transforming done model, but they are not in out-of-scope modification, they could be just to inspect the model and make decisions. But even that is relatively hard, let's imagine with the same tree 
```
Top
  - Sub1
    - Sub1Sub1
    - Sub1Sub2
  - Sub2
    -Sub2Sub1
    -Sub2Sub2
```


So again we have model transformer on Sub1 and Sub2, and model transformer at Sub1 says something like this

`(spec) -> if (spec.getParentCommand().subcommands("Sub2").hasSubcommand("Sub2Sub1") spec.removeSubcommand("Sub1Sub2") ` 

and model transformer on Sub2 says 

`(spec) -> if (Boolean.valueOf(System.getProp("whatever"))) spec.removeSubcommand("Sub2Sub1")`

but than we'd have to construct dependency tree, because Sub2 transformer has to run before Sub1 transformer etc.. and it wouldn't work for cyclic deps .. much work for in my opinion small benefit.

remkop: If we were to implement transformation after the model is completely constructed, how would we go about it?

MarkoMackic: @remkop we'll recursive tree traversal inside constructor, after spec creation ? maybe provide a method to replace spec if there isn't one?

MarkoMackic: But what do you think about the notes above ? is it worth it ? We shouldn't rush.

remkop: I think your argument is basically that, even if we do transformations after the model is fully built, that does not eliminate the possibility for people to get themselves into trouble (for example by having 2 transformations that operate on the same node).

You may be right.
My intuition is that the "right" thing to do is to wait with transformations until the model is fully built.
One advantage is that it makes it possible to (in the future) have different traversal orders: breadth-first, depth-first, and variations of these. 
It just does not feel right to do transformations while the model is not fully built, sorry if that is vague.

Is the effort that much? I think the key question is where to put the call to start the transformations.

MarkoMackic: Ok, I'll refactor due to your request, but I hold my attitude that it's a wrong decision :)

MarkoMackic:
@remkop Please review, CommandSpec is not final anymore because it's replaced by preprocessors. This would allow preprocessor to return totally new command spec ( maybe you want only to have methods implemented in CommandSpec that are able to modify it, than interface signature would be `void(CommandSpec)`  )

src/main/java/picocli/CommandLine.java
```diff
@@ -217,15 +217,20 @@ public CommandLine(Object command) {
...
         this.factory = Assert.notNull(factory, "factory");
         interpreter = new Interpreter();
         commandSpec = CommandSpec.forAnnotatedObject(command, factory);
         commandSpec.commandLine(this);
+        if(userCalled) { this.applyModelTransformations(); }
         commandSpec.validate();
         if (commandSpec.unmatchedArgsBindings().size() > 0) { setUnmatchedArgumentsAllowed(true); }
     }
 
+    public CommandLine(Object command, IFactory factory) {
```
remkop, line 230: Let's move the javadoc from the private constructor to the public one.

src/main/java/picocli/CommandLine.java
```diff
@@ -217,15 +217,20 @@ public CommandLine(Object command) {
      * @param factory the factory used to create instances of {@linkplain Command#subcommands() subcommands}, {@linkplain Option#converter() converters}, etc., that are registered declaratively with annotation attributes
      * @throws InitializationException if the specified command object does not have a {@link Command}, {@link Option} or {@link Parameters} annotation
      * @since 2.2 */
-    public CommandLine(Object command, IFactory factory) {
+    private CommandLine(Object command, IFactory factory, boolean userCalled) {
         this.factory = Assert.notNull(factory, "factory");
         interpreter = new Interpreter();
         commandSpec = CommandSpec.forAnnotatedObject(command, factory);
         commandSpec.commandLine(this);
+        if(userCalled) { this.applyModelTransformations(); }
```
remkop, line 225: Please put a space between `if` and the opening bracket `(`.

src/main/java/picocli/CommandLine.java
```diff
@@ -4799,6 +4822,22 @@ private static String format(String formatString, Object... params) {
...
+     * @since X.X */
+    public interface IModelTransformer {
+        /**
+         * Returns CommandSpec after doing transformation.
+         * @return transformed CommandSpec
+         */
+        CommandSpec transform(CommandSpec commandSpec);
+    }
+
+    private static class NoOpModelTransforemer implements IModelTransformer {
```
remkop, line 4837: Small typo: `NoOpModelTransforemer` should be `NoOpModelTransformer`.

src/main/java/picocli/CommandLine.java
```diff
@@ -4799,6 +4822,22 @@ private static String format(String formatString, Object... params) {
         public String[] getVersion() throws Exception { throw new UnsupportedOperationException(); }
     }
 
+    /**
+     * Provides a way to modify how command model is built. Commands may configure a model transformer using
+     * {@link Command#modelTransformer()} annotation attribute.
+     * @since X.X */
```
remkop, line 4828: Let's add some text about the timing:

```java
     * Provides a way to modify how the command model is built.
     * This is useful for applications that need to modify the model dynamically depending on the runtime environment.
     * <p>
     * Commands may configure a model transformer using the
     * {@link Command#modelTransformer()} annotation attribute, or via the
     * {@link CommandSpec#modelTransformer(IModelTransformer)} programmatic API.
     * <p>
     * The transformers are invoked only once, after the full command hierarchy is constructed.
```

src/main/java/picocli/CommandLine.java
```diff
@@ -6964,6 +7007,9 @@ void initDefaultValueProvider(Class<? extends IDefaultValueProvider> value, IFac
             void initExitCodeOnInvalidInput(int exitCode)       { if (initializable(exitCodeOnInvalidInput, exitCode, ExitCode.USAGE)) { exitCodeOnInvalidInput = exitCode; } }
             void initExitCodeOnExecutionException(int exitCode) { if (initializable(exitCodeOnExecutionException, exitCode, ExitCode.SOFTWARE)) { exitCodeOnExecutionException = exitCode; } }
             void updateName(String value)                   { if (isNonDefault(value, DEFAULT_COMMAND_NAME))                {name = value;} }
+            void updateModelTransformer(Class<? extends IModelTransformer> value) {
+                if (isNonDefault(value, NoOpModelTransforemer.class)) { this.modelTransformer = DefaultFactory.create(defaultFactory(), value); }
```
remkop, line 7011: Please use the `factory` that is passed as parameter to the `updateCommandAttributes` method, not the default factory.

src/main/java/picocli/CommandLine.java
```diff
@@ -5864,6 +5903,8 @@ public void clear() {
             private Integer exitCodeOnInvalidInput;
             private Integer exitCodeOnExecutionException;
 
+            private IModelTransformer modelTransformer = null;
```
remkop, line 5906: Can we add public getter and setter accessors for this attribute to `CommandSpec`?
Following the naming convention I use here, that would be

```java
/** Some javadoc here
  * @since 4.6 */
public IModelTransformer modelTransformer() { return modelTransformer; }

/** Some javadoc here
  * @since 4.6 */
public CommandSpec modelTransformer(IModelTransformer modelTransformer) { this.modelTransformer = modelTransformer; return this; }

**Review by remkop — CHANGES_REQUESTED** [review]
Looks good, thank you! 
I think we are getting close.
I added some feedback for minor adjustments, can you take a look?

We should also mention this feature in the user manual ([index.adoc](https://github.com/remkop/picocli/blob/master/docs/index.adoc)).
Perhaps add a small section called **Model Transformations**, between [@spec annotation](https://picocli.info/#spec-annotation) and [Custom Factory].(https://picocli.info/#_custom_factory). If you can provide an initial draft, perhaps with a usage example, then I take care of the rest.

**Comment by MarkoMackic** [comment]
This is going with 4.6 ?

File: src/main/java/picocli/CommandLine.java (hunk 11 above)
Comment (MarkoMackic) [reply src/main/java/picocli/CommandLine.java:5906]: That's ok, than we need to add public API to CommandLine to manually trigger the transformation ? I'll see into this.

Comment (MarkoMackic) [reply src/main/java/picocli/CommandLine.java:5906]: Actually you don't need programmatic API for this :) because you're constructing command line on the fly ( i.e. your code does construction ) , but maybe it can serve some purpose.

File: src/main/java/picocli/CommandLine.java (hunk 10 above)
Comment (MarkoMackic) [reply src/main/java/picocli/CommandLine.java:7011]: Ok.

File: src/main/java/picocli/CommandLine.java (hunk 9 above)
Comment (MarkoMackic) [reply src/main/java/picocli/CommandLine.java:4828]: Ok.

File: src/main/java/picocli/CommandLine.java (hunk 8 above)
Comment (MarkoMackic) [reply src/main/java/picocli/CommandLine.java:4837]: Ok.

File: src/main/java/picocli/CommandLine.java (hunk 7 above)
Comment (MarkoMackic) [reply src/main/java/picocli/CommandLine.java:225]: Ok.

File: src/main/java/picocli/CommandLine.java (hunk 6 above)
Comment (MarkoMackic) [reply src/main/java/picocli/CommandLine.java:230]: Ok.

File: src/main/java/picocli/CommandLine.java (hunk 11 above)
Comment (MarkoMackic) [reply src/main/java/picocli/CommandLine.java]: I've done it.

**Comment by MarkoMackic** [comment]
@remkop Could you please write the docs ?

File: src/main/java/picocli/CommandLine.java [hunk 12]
```diff
@@ -4799,6 +4822,28 @@ private static String format(String formatString, Object... params) {
...
+     * Provides a way to modify how the command model is built.
+     * This is useful for applications that need to modify the model dynamically depending on the runtime environment.
+     * <p>
+     * Commands may configure a model transformer using the
+     * {@link Command#modelTransformer()} annotation attribute, or via the
+     * {@link CommandSpec#modelTransformer(IModelTransformer)} programmatic API.
+     * <p>
+     * The transformers are invoked only once, after the full command hierarchy is constructed.
+     */
+    public interface IModelTransformer {
```
remkop, line 4835: Let's add a `@since 4.6` tag to this javadoc.

src/test/java/picocli/CommandModelTransformersTest.java
```diff
@@ -0,0 +1,203 @@
...
+            return 0;
+        }
+    }
+
+
+    @CommandLine.Command(name = "a", subcommands = {AA.class}, modelTransformer = A.AFilter.class)
+    private static class A implements Callable<Integer> {
+
+        private static class AFilter implements CommandLine.IModelTransformer {
+            public CommandLine.Model.CommandSpec transform(CommandLine.Model.CommandSpec commandSpec) {
```
remkop, line 56: If we import `CommandLine.Model.CommandSpec`, it will make the code nicer to read. Maybe also import `CommandLine.Command` and `CommandLine.IModelTransformer`.

src/test/java/picocli/CommandModelTransformersTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+        assertEquals(2, cl.execute("d"));
+
+        assertEquals(0, cl.execute("b"));
+
+        // should throw unmatched argument exception
+        cl.parseArgs("b");
+
+        // should throw unmatched argument exception
+        cl.parseArgs("c");
```
remkop, line 199: Is this statement reached, given that `parseArgs("b")` will throw an exception?

remkop (approved):
This looks good, thank you very much!

I added some minor feedback, would be nice if you can take care of them, if not, I can do it.
I will update the user manual, no worries.

src/test/java/picocli/CommandModelTransformersTest.java
MarkoMackic: Nope, I'll fix it.

src/main/java/picocli/CommandLine.java
MarkoMackic: Ok.

src/test/java/picocli/CommandModelTransformersTest.java
MarkoMackic: Ok.

MarkoMackic:
@remkop All good :) Please make one final review, add this to user manual and we're ready to go.

remkop (approved):
Looks good, many thanks!

MarkoMackic:
@remkop When is 4.6 going to be released ?

remkop:
I have taken next week off from work. That is when I am planning to merge approved PRs and complete other outstanding work for 4.6 and do the release.

remkop:
Merged and added documentation.
Feedback welcome!

Issue #802, opened 2019-09-05
Allow interactive options that echo to the console


portus84:
Hi, now interactive mode hide text digitized by user. There is a mode to use interactive mode which show text that user insert during write on console?

remkop:
Currently the picocli library does not provide this functionality, but I think it would be a good feature to have.

Perhaps we should add an `echo` attribute, that allows applications to control whether the user input is echo-ed to the console or not. I'm also thinking to add a `prompt` attribute so that applications can customize the text displayed to the end user for an interactive option. For example:

```
@Option(names = "-x", interactive = true, echo = true, prompt = "Enter your PIN: ")
String pin;
```

remkop:
Note that you can work around this in the application, by calling `Console.readLine`:

```java
public class Interactive implements Runnable {
    @Option(names = {"-u", "--user"}, interactive = true,
            description = "user id")
    private String user;

    public void run() {
        if (user == null) {
            user = new String(System.console().readLine("We really need a user id: "));
        }
        System.out.printf("Ok, we got: %s%n", user);
        // do stuff...
    }

    public static void main(String[] args) {
        int exitCode = new CommandLine(new Interactive()).execute(args);
    }
}
```

xinchao-zhang:
> Note that you can work around this in the application, by calling `Console.readLine`:
> 
> ```java
> public class Interactive implements Runnable {
>     @Option(names = {"-u", "--user"}, interactive = true,
>             description = "user id")
>     private String user;
> 
>     public void run() {
>         if (user == null) {
>             user = new String(System.console().readLine("We really need a user id: "));
>         }
>         System.out.printf("Ok, we got: %s%n", user);
>         // do stuff...
>     }
> 
>     public static void main(String[] args) {
>         int exitCode = new CommandLine(new Interactive()).execute(args);
>     }
> }
> ```

Then, the interactive flag is useless in this case and can be dropped, right?

remkop:
@xinchao-zhang Yes, I think you are correct.

sakata1222:
@remkop Hi, I'm interested in 'echo` and `prompt` feature. Can I try to create a PR?

remkop:
Yes, great!

PR #1284, opened 2020-12-13
Support echo and prompt for an interactive option

sakata1222:
PR for #802.

- If `echo` is true, just uses `BufferedReader#readLine` not `Console#readPassword` to read user input.
- If `prompt` is not empty, the specified string will be displayed to the end user for an interactive option.
- The default values:
  - `echo` : false
  - `prompt` : empty string

Here is the result of a sample class added by this PR. I use Windows 10.

```bash
PS C:\xxxxx\picocli> java -version
openjdk version "11.0.9.1" 2020-11-04
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.9.1+1)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.9.1+1, mixed mode)

PS C:\xxxxx\picocli> java -cp "build/libs/picocli-4.5.3-SNAPSHOT.jar;picocli-examples/build/libs/picocli-examples.jar" picocli.examples.interactive.UserPasswordDemo "sample action"
Enter your user: dummy
Enter your password:
User: dummy, Password: secret
Action: sample action
```

### DIFF
--- a/picocli-examples/src/main/java/picocli/examples/interactive/UserPasswordDemo.java
+++ b/picocli-examples/src/main/java/picocli/examples/interactive/UserPasswordDemo.java
@@ -1,0 +1,34 @@
+package picocli.examples.interactive;
+
+import picocli.CommandLine;
+import picocli.CommandLine.Parameters;
+
+public class UserPasswordDemo implements Runnable {
+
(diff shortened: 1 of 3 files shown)


remkop:
Merged.
Thank you for the contribution! This will be included in the 4.6 release.
(I still need to update the release notes.)

Issue #1274, opened 2020-12-06
ReflectionConfigGenerator should not generate method section in subclass for private superclass methods (was: Bug in ReflectionConfigGenerator)


sebthom:
I am trying to compile a Java cli app with GraalVM native-image and realized that it emits warnings like `WARNING: Method com.example.MyCommand.setVerbosity(boolean[]) not found.`

The issue was, that the `setVerbosity` method was declared as private method in an abstract parent command class, but ReflectionConfigGenerator generates this `reflect-config.json`

```json
[
  {
    "name" : "com.example.MyCommand",
    "allDeclaredConstructors" : true,
    "allPublicConstructors" : true,
    "allDeclaredMethods" : true,
    "allPublicMethods" : true,
    "methods" : [
      { "name" : "setVerbosity", "parameterTypes" : ["[Z"] }
    ]
  },
  {
    "name" : "com.example.AbstractCommand",
    "allDeclaredConstructors" : true,
    "allPublicConstructors" : true,
    "allDeclaredMethods" : true,
    "allPublicMethods" : true,
    "fields" : [
      { "name" : "commandSpec" }
    ],
    "methods" : [
      { "name" : "setVerbosity", "parameterTypes" : ["[Z"] }
    ]
  }
]
```

Looking at the ReflectionConfigGenerator I realized that the logic is broken at https://github.com/remkop/picocli/blob/f2f2ac3230b9f6b5f52d159bda515ce805c0be88/picocli-codegen/src/main/java/picocli/codegen/aot/graalvm/ReflectionConfigGenerator.java#L482

I actually believe it should say
```java
if (scope.get().getClass().equals(method.getDeclaringClass())) {
```
instead of
```java
if (!scope.getClass().equals(method.getDeclaringClass())) {
```
which always evaluates to `true`

remkop:
Thanks for raising this!
I’ll take a look.

remkop:
I finally have time to look at this. Is my understanding below correct?

Given the code below to reproduce the issue:

```java
package picocli.codegen.aot.graalvm;

import picocli.CommandLine.Command;
import picocli.CommandLine.Model.CommandSpec;
import picocli.CommandLine.Option;
import picocli.CommandLine.Spec;

@Command
public class Issue1274AbstractCommand {
    @Spec
    CommandSpec commandSpec;

    @Option(names = {"-v", "--verbose"})
    private void setVerbosity(boolean[] verbosity) {
    }
}
```

```java
[...]

Issue #1217, opened 2020-10-15
Preprocessor callbacks during parsing (was: Make string options behave same as boolean options)


remkop:
From [discussion](https://groups.google.com/g/picocli/c/__tziJIw3lE/m/UCliKrsPBQAJ) on the picocli google group mailing list, topic raised by @maxandersen:

> when I have:
> 
> ```
> @CommandLine.Option(names = {"--live" })
> boolean live;
> ```
> 
> then `jbang --live=true` and `jbang --live=false` works, and `jbang --live false` is not interpreting the false argument as a possible value. and `jbang --live=custom` i can even intercept with custom value handler.
> 
> now...how can I get the same for 
> 
> ```
> @CommandLine.Option(names = {"--live" })
> String live;
> ```
> 
> with this I cannot force that only `--live=myarg` gets taken as input and that `--live myarg`  should not be treated as a value to `live`.
> 
> any way for this to work ?
> 

My reply:

> The intention was that `--option=value` and `--option value` would be treated identically.
> There is indeed some special logic for booleans, that takes the separator into account.
> This is because boolean options have arity=0, meaning they are not expected to have a parameter.
> If the user specified `--option=value`, then the presence of the separator "forces" us to accept "value" as a parameter, and that is what the logic does.
> 
> This is a special case. Generally I would think that treating  `--option=value` and `--option value`  differently would be confusing to users, but there may be other use case that I don't understand yet.
> 
> Currently the `IParameterConsumer` interface does not give any indication of whether the value was attached with a separator or not, that may have been a bad decision.

@maxandersen's reply:

> I think it would been great to know if parameter is done via = or not as I really dont see another way of allowing to have options that can have a default beavhior but then be able to specify another option....well you can but that would require having one boolean nand then another flag for the option.
> 
> i.e. in my case its jbang where I want to do `jbang edit --open myfile.java` (which says please make temporary project for myfile.java for me to edit, oh and please open my default editor on that project) 
> but I also want to be able to write `jbang edit --open=idea myfile.java` which says the same but says, use the idea command/editor in this case.
> 
> There is no real way for me to know wether `myfile.java` is actually the editor he wants to use or not ....
> 
> I've used parameterconsumer in other cases pretty nicely for numeric values where if the value was a number i would acccept it but not if a string/filename that existed - but it just gets dirty.
> 
> And when I look at all my special cases I could have handled it much more uniformly if I had a "OnlyExplicitlyAssignedParameterConsumer.class"  or something ;) 
[...]

Issue #294, opened 2018-03-05
[Feature request] Native zsh completion


gyscos:
Picocli currently can create bash completion files, and zsh has a compatibility tool to read bash completion files.

However, this only exposes basic autocompletion features (what bash itself support), and does not take advantage of the features zsh offer.

For instance, zsh can show a description for each command/option, as well as option aliases (short vs long options). Here is an example for the git command:

![screenshot from 2018-03-05 10-48-22](https://user-images.githubusercontent.com/564903/36993485-d0e31074-2062-11e8-87de-6733da0b11e3.png)

Therefore, I propose that a new `Autocomplete.zsh` static method be added, to output a zsh-native completion script that includes more details, like command and option description and aliases.

remkop:
I think that's great idea and that would be a very cool feature!
Will you be able to provide a pull request?

hristo-vrigazov:
+1, this would be awesome

yschimke:
FWIW to get even the existing zsh support working through bash, I needed the following in my zsh script

https://github.com/yschimke/homebrew-tap/blob/master/rsocket-cli.rb#L18

Below is from a homebrew recipe

```
    system "echo 'autoload -U +X bashcompinit && bashcompinit' > #{prefix}/bash/completion.zsh"
    system "#{bin}/rsocket-cli --completionScript >> #{prefix}/bash/completion.zsh"
zsh_completion.install "#{prefix}/bash/completion.zsh" => "_rsocket-cli"
```

yschimke:
The other relevant thing is that the file should be in FPATH and start with the #compdef line so shell login doesn't involve running all the scripts

https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org

remkop:
@yschimke thanks for the tip!

yschimke:
The testing and debugging section of that doc is good also e.g.


| _complete_help | Ctrl+x h | displays information about context names, tags, and completion functions used when completing at the current cursor position |

byteit101:
It's *super* hacky,  and only supports the bare minimum needed for one-level descriptions on my specific app, but I made a first attempt at this on https://github.com/byteit101/picocli/commit/7f9d1cfd5e48fa3729adb3e36098f2bf1e32cdd0

I do plan on eventually getting around to cleaning it up to make it into an integrateable PR, but unfortunately I'm rather busy with other things for the next month or so, if anyone else wants to run with it.

remkop:
[...]

PR #1285, opened 2020-12-14
Documentation, code samples: adding tab(s) to the right

deining:
This PR extends the current infrastructure for displaying tabs on source code blocks.
After merging this PR, you can position tabs on source code blocks both to the left and to the right.
This might be useful in various places, As an example, I added two additional tabs 'Invocation' and 'Usage help' in chapter `3.1. Options` of the manual. A screenshot illustrates this use case:

image

IMHO this is a very compact way to illustrate and convey the various aspects of picocli usage.
If you like this kind of tab display, I'm willing to identify further code samples that might benefit from this additional tabs and author the content for those tabs.
Please note that I had to add the bootstrap css framework in order to implement this kind of tab representation. This slightly changed the display of the manual page. If you don't like certain aspects of the display, let me know and I will try to correct this.


remkop:
Hi @deining, that looks interesting. Please give me a few days: I have some limited time now and I want to focus on getting a 4.6 release out. I would like to include #1274 (bugfix), #1217 (parser callback) and #294 (zsh completion) in that release.
It may take me a few days before I can look at this PR. Please bear with me for a while.

remkop:
Sorry for keeping this open for so long; I wasn't able to make a decision quickly, and then I did not look at this for a long time.
Apologies! :pray: 

I think I prefer the way that the information is currently presented in the manual.
My thinking is as follows:

The Java/Groovy/Kotlin code blocks are interchangeable. 
These blocks present the same information, the same logic, only using different syntax.
This allows users to pick the programming language they are most familiar with, and zoom in on the example that is easiest to absorb.
If users never click on anything, they will simply see the example in the default language (Java) and not miss anything.

The Invocation and Usage help code blocks are different in this respect:
They form a sequence, like a story, with a beginning and an end, and users will probably need to look at both.
Users may not realize they need to click on a tab to see the full story, which could potentially be confusing.
Since users need to look at both the Invocation and Usage Help of examples anyway, it is more natural to simply show them both in subsequent paragraphs.

Again, apologies for the long wait.
I want to say again that your contributions to the picocli manual are invaluable. Thank you very very much!

PR #1286, opened 2020-12-22
documentation fix: @ParentObject -> @ParentCommand

lalmeras:
I find no @ParentObject annotation. Does it need to be replaced by @ParentCommand, as it seems to me it is the right annotation ?


remkop:
Oops! Nice catch! 

Merged. I will update the release notes soon.
Thanks for the contribution!

PR #1288, opened 2020-12-30
Make interface public

maxandersen:


maxandersen:
as discussed in #1217 otherwise I can't implement the interface ;)

remkop:
Thank you, I didn’t realize!

PR #1290, opened 2020-12-31
JLine: change keystroke syntax 'Ctl-D' to more common used syntax 'Ctrl-D'

deining:
Though certainly not wrong, `Ctl-D` seems to be rarely used, so I propose to change it to `Ctrl-D` ( that's what I see on English keyboards and in many other places).


remkop:
Merged, thank you!

(TODO: update release notes)

remkop:
Updated RELEASE-NOTES.

Issue #1291, opened 2021-01-01
Cannot compile with picocli-shell-jline3 dependency using Maven


sualeh:
Since the change to [Fix gradle warnings](https://github.com/remkop/picocli/commit/758558b9817660b6c6e497aab28cb8ac7421127e), Maven compilation of any project with [picocli-shell-jline3](https://search.maven.org/search?q=a:picocli-shell-jline3) as a dependency is broken, since the artifact published to The Central Repository has a changed the scope of transitive dependencies to jline - see the following changes to scope in the published pom.xml:

```xml
    <dependency>
      <groupId>org.jline</groupId>
      <artifactId>jline</artifactId>
      <version>3.17.1</version>
      <scope>runtime</scope>
    </dependency>
```
whereas in former versions of picocli it was

```xml
    <dependency>
      <groupId>org.jline</groupId>
      <artifactId>jline</artifactId>
      <version>3.16.0</version>
      <scope>compile</scope>
    </dependency>
```

In order words, the change in how the Gradle dependency is defined has broken the Maven scope, and compilation fails.

sualeh:
@deining - copying you for information too.

remkop:
@sualeh Thank you for raising this!

Will applications be able to work around this by explicitly adding this dependency to their POM? 

```xml
    <dependency>
      <groupId>org.jline</groupId>
      <artifactId>jline</artifactId>
      <version>3.16.0</version>
      <scope>compile</scope>
    </dependency>
```

If so, we can try to fix this in the next release, but in the meantime add a note for this to the documentation.

sualeh:
Here is the way I suggest:

```xml
    <dependency>
      <groupId>info.picocli</groupId>
      <artifactId>picocli-shell-jline3</artifactId>
      <version>4.6.0</version>
      <exclusions>
        <exclusion>
          <groupId>org.jline</groupId>
          <artifactId>jline</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
    <dependency>
      <groupId>org.jline</groupId>
      <artifactId>jline</artifactId>
      <version>3.17.1</version>
    </dependency>
```

dejan2609:
Related issue #1294 is reported by @remal (I will try to expand my PR asap).

sualeh:
@remkop  - may I request a point release with this fix?

remkop:
Yes will do.

remkop:
Thanks again for raising this!
This is fixed in the 4.6.1 release.

[...]

PR #1292, opened 2021-01-02
[#1291] build.gradle for 'picocli-shell-jline3' submodule is changed: Gradle configuration is switched (from 'implementation' to 'api') for 'org.jline:jline' dependency

dejan2609:
note: solves broken build issues for any Maven project that uses 'picocli-shell-jline3' version 4.6.0

related issue: https://github.com/remkop/picocli/issues/1291 **_Cannot compile with picocli-shell-jline3 dependency using Maven_**

### DIFF
--- a/picocli-shell-jline3/build.gradle
+++ b/picocli-shell-jline3/build.gradle
@@ -13,7 +13,7 @@ targetCompatibility = 1.8
 
 dependencies {
     implementation     rootProject
-    implementation     "org.jline:jline:$jline3Version"
+    api                "org.jline:jline:$jline3Version"
     implementation     "org.jline:jline-console:$jline3Version"
     testImplementation "junit:junit:$junitVersion"
 }


remal:
I suppose the same change should be applied to all other modules, as we have the same issue for `picocli-spring-boot-starter` module: #1294

picocli-shell-jline3/build.gradle
```diff
@@ -13,7 +13,7 @@ targetCompatibility = 1.8
 
 dependencies {
     implementation     rootProject
```
remal, line 15: Shouldn't we change this to `api` too?

remal (changes requested):

picocli-shell-jline3/build.gradle
dejan2609: Thanx for a review @remal ! I will push another commit for `picocli-spring-boot-starter` tomorrow (and hopefully all these commits will be squashed into one if/when this gets merged into master).

As for a `rootProject`... To be honest: this is the first that I see this Gradle directive used as a dependency (I presume that it references code/dependency  from root/src (that somehow didn't find its way into some separate submodule). 

:fast_forward: Going further: if we compare transitive dependencies for `picocli-shell-jline3` recent versions (4.5.2 vs. 4.6.0):
- https://search.maven.org/artifact/info.picocli/picocli-shell-jline3/4.5.2/jar
- https://search.maven.org/artifact/info.picocli/picocli-shell-jline3/4.6.0/jar

we will found that Maven scope changed from **_compile_** to **_runtime_** for these **_three_** dependencies: 
1. `info.picocli:picocli` (root project, mentioned above)
2. `org.jline:jline` (https://github.com/jline/jline3/blob/master/jline/pom.xml)
3. `org.jline:jline-console` (https://github.com/jline/jline3/blob/master/console/pom.xml)

It seems that Gradle configuration for all of those three dependencies should be changed from **_implementation_** to _**api** in order to re-introduce Maven scope **_compile_** (when those dependencies are used transitively, that is).

I will have to sleep on this :sleeping:

remkop:
Thank you both for the contribution!
This was a great help in understanding what needed to be done.
Picocli 4.6.1 has been released with these changes.

dejan2609:
No problem @remkop ! I see that you were super-fast and that changes are landed into master: https://github.com/remkop/picocli/commits/v4.6.1 😃

remkop:
@dejan2609 Thank you for your help! 👍🙏

PR #1296, opened 2021-01-04
Docu, new code samples: Kotlin versions, minor changes

deining:
This PR adds Kotlin versions for the new code samples in picocli 4.6 (`IModelTransformer`, `IParameterProcessor`).
While working on the `IParameterProcessor` sample, I made minor changes to the code sample and to the documentation in order to improve readability and intelligibility. I hope you like them.
Additionally, I added a third column to the comparison table `PicocliBaseScript2` vs. `PicocliBaseScript`.

### DIFF
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -10050,12 +10050,13 @@ interface IModelTransformer {
 
 This allows applications to dynamically add or remove options, positional parameters or subcommands, or modify the command in any other way, based on some runtime condition.
 
-[source,java]
+.Java
+[source,java,role="primary"]
 ----
 @Command(modelTransformer = Dynamic.SubCmdFilter.class)
 class Dynamic {
 
-    private static class SubCmdFilter implements IModelTransformer {
+    static class SubCmdFilter implements IModelTransformer {
         public CommandSpec transform(CommandSpec commandSpec) {
             if (Boolean.getBoolean("disable_sub")) {
                 commandSpec.removeSubcommand("sub");
@@ -10066,7 +10067,29 @@ class Dynamic {
 
     @Command
     private void sub() {
-        // subcommand business logic
+        // subcommand, business logic
+    }
+}
+----
+
+.Kotlin
+[source,kotlin,role="secondary"]
+----
+@Command(modelTransformer = Dynamic.SubCmdFilter::class)
+class Dynamic {
+
+    class SubCmdFilter : CommandLine.IModelTransformer {
+        override fun transform(commandSpec: CommandSpec): CommandSpec {
+            if (Boolean.getBoolean("disable_sub")) {
+                commandSpec.removeSubcommand("sub")
+            }
+            return commandSpec
+        }
+    }
+
+    @Command
+    private fun sub() {
+        // subcommand, business logic
     }
 }
 ----
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -2346,6 +2346,7 @@ When `-x VALUE` is specified on the command line, this results in an error: `Mis
 Users can use the <<Double dash (`--`),end-of-options delimiter>> and disambiguate the input with
 `-x &dash;&dash; VALUE`, but this may not be obvious to many users.
 One idea is to <<Show End of Options,show the end-of-options>> delimiter in the usage help.
+Another idea is to make use of the <<IParameterPreprocessor Parser Plugin,IParameterPreprocessor Parser Plugin>> introduced with picocli 4.6.
 
 An alternative is to avoid the use of optional parameters and use the default arity in this scenario to eliminate the ambiguity altogether.
 
@@ -4776,7 +4777,7 @@ class ExecParameterConsumer : IParameterConsumer {
 }
 ----
 
-==== `IParameterPreprocessor` Parser Plugin
+==== IParameterPreprocessor Parser Plugin
 
 Introduced in picocli 4.6, the `IParameterPreprocessor` is also a parser plugin, similar to `IParameterConsumer`, but more flexible.
 
@@ -4798,7 +4799,7 @@ edit [--open[=<editor>]] <file>
 One of the <<Optional Parameter Limitations, limitations of options with an optional parameter>> is that they are difficult to combine with positional parameters.
 
 With a custom parser plugin, we can customize the parser, such that `VALUE` in `--option=VALUE` is interpreted as the option parameter, and in `--option VALUE` (without the `=` separator), VALUE is interpreted as the positional parameter.
-The code below demonstrates:
+The code below demonstrates this:
 
 .Java
 [source,java,role="primary"]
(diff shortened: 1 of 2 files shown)


docs/index.adoc
```diff
@@ -4798,26 +4799,27 @@ edit [--open[=<editor>]] <file>
...
 ----
 @Command(name = "edit")
 class Edit {
 
-    @Parameters(index = "0", description = "The file to edit.")
+    @Parameters(index = "0", arity="0..1", description = "The file to edit.")
     File file;
 
-   enum Editor { defaultEditor, eclipse, idea, netbeans }
+   enum Editor { eclipse, idea, netbeans }
```
remkop, line 4813: I think we should keep `defaultEditor`: it is good to have a separate value to represent the absence of choice by the end user; I am imagining that in the application this would come from some configuration (so not hard-coded).

docs/index.adoc
```diff
@@ -4798,26 +4799,27 @@ edit [--open[=<editor>]] <file>
...
 
    @Option(names = "--open", arity = "0..1", preprocessor = Edit.MyPreprocessor.class,
        description = {
-           "Optionally specify the editor to use; if omitted the default editor is used. ",
+           "Optionally specify the editor to use (${COMPLETION-CANDIDATES}). " +
+           "If omitted eclipse is used as default. ",
            "Example: edit --open=idea FILE opens IntelliJ IDEA (notice the '=' separator)",
-           "         edit --open FILE opens the specified file in the default editor"
+           "         edit --open FILE opens the specified file in eclipse"
        })
```
remkop, line 4821: Similar to the above, I think it is meaningful to have a `defaultEditor` choice, so let's keep the original description.

docs/index.adoc
```diff
@@ -4798,26 +4799,27 @@ edit [--open[=<editor>]] <file>
...
-   Editor editor = Editor.defaultEditor;
+   static Editor editor = Editor.eclipse;
```
remkop, line 4822: Similar to previous (default should be `defaultEditor`, not `eclipse`). 
Also, no need for the `static` keyword (see below).

docs/index.adoc
```diff
@@ -4828,36 +4830,71 @@ class Edit {
            // or the name of the editor to use...
            if (" ".equals(info.get("separator"))) { // parameter was not attached to option
 
-               // act as if the user specified --open=defaultEditor
-               args.push(Editor.defaultEditor.name());
+               // act as if the user specified --open=eclipse
+               args.push(editor.name());
```
remkop, line 4834: I think it is good to explicitly use the enum constant that represents the default, so `Editor.defaultEditor` (or `Editor.eclipse` if Eclipse were the default).

This is mostly for clarity: so that people reading this code example in the documentation would not have to refer back in the code to what the actual value would be.

I realize that in "real" code there is a tension here with the principle of having information in only once place. To be honest, I am not sure what I would do in a "real" application in this case. :-)

But for documentation purposes, I believe that duplicating the default value in the preprocessor is better because it requires less work on the part of the reader.

docs/index.adoc
```diff
@@ -4776,7 +4777,7 @@ class ExecParameterConsumer : IParameterConsumer {
 }
 ----
 
-==== `IParameterPreprocessor` Parser Plugin
+==== IParameterPreprocessor Parser Plugin
```
remkop, line 4780: This probably looks better. :-) Should we do the same for IParameterConsumer?

remkop (changes requested):
Wow, awesome work again!

The only think I would like to change is that I think we should keep `Editor.defaultEditor` to represent the absence of choice by the end user, rather than just picking one of the actual editors as the default. I put more detailed comments with the proposed changes.

Other than that, it looks great!

deining:
> Wow, awesome work again!

Great you like it!
I realized that the comparison table `PicocliBaseScript2` vs. `PicocliBaseScript` is present in the release notes, too. For the sake of consistency, I would like to modify this table, too. I hope this is fine for you.

> The only think I would like to change is that I think we should keep `Editor.defaultEditor` to represent the absence of choice by the end user, rather than just picking one of the actual editors as the default. I put more detailed comments with the proposed changes.

I will work on this shortly, you may expect an updated PR soon.

> Other than that, it looks great!

Thank you!

While working on this, two further issues came up that I would like to raise:

- I'm wondering, whether the API Docs for the package `picocli.groovy` are up to date. I would suspect these docs at this [URL](https://picocli.info/apidocs/picocli/groovy/package-summary.html), but these documents are 4.0.0-alpha-3. I didn't find a link to these API docs in any `README.md`file in the repository either, I think this is something that should be added.
- `ParameterConsumer` vs `IParameterPreprocessor`: AFAICS, the `IParameterPreprocessor` is superior to the `ParameterConsumer` in all aspects. Is there any reason to still stick to the `ParameterConsumer`? If not, this interface should be marked as deprecated IMHO. Also, chapter `11.15. Custom Parameter Processing` lists both plugin in chronological order concerning their authoring date. I think they should be ordered according to relevance for the user, therefore I would suggest to swap the first two subchapters.

remkop:
> I realized that the comparison table `PicocliBaseScript2` vs. `PicocliBaseScript` is present in the release notes, too. For the sake of consistency, I would like to modify this table, too. I hope this is fine for you.

I'm okay to leave the release notes as they are, and just evolve the user manual. I see the release notes as more of a snapshot of how things were at some point in time, so there is some value in not updating them.

> * I'm wondering, whether the API Docs for the package `picocli.groovy` are up to date. I would suspect these docs at this [URL](https://picocli.info/apidocs/picocli/groovy/package-summary.html), but these documents are 4.0.0-alpha-3. I didn't find a link to these API docs in any `README.md`file in the repository either, I think this is something that should be added.

Oh, good catch! I have completely forgotten about this... 
The Groovy classes were previously in the main picocli jar, now they are in a separate jar...  I suspect they are still there from when these classes were in the picocli jar.

It makes sense to publish Javadoc for the other jars, not just for the picocli jar, but also for picocli-groovy, picocli-jline2/3, picocli-spring-boot and picocli-codegen. The best way to deal with this is probably to update the `copyDocs` task in `build.gradle`. Also, like you said, we need to link to their top-level pages from the user manual and project README.

> * `IParameterConsumer` vs `IParameterPreprocessor`: AFAICS, the `IParameterPreprocessor` is superior to the `IParameterConsumer` in all aspects. Is there any reason to still stick to the `IParameterConsumer`? If not, this interface should be marked as deprecated IMHO. Also, chapter `11.15. Custom Parameter Processing` lists both plugin in chronological order concerning their authoring date. I think they should be ordered according to relevance for the user, therefore I would suggest to swap the first two subchapters.

Swapping these in the manual makes sense; `IParameterPreprocessor` is more flexible than `IParameterConsumer`. That said,  `IParameterConsumer` is simpler, so I can imagine people preferring to use that one. I am not in a rush to deprecate `IParameterConsumer`. It is probably a good idea to update the javadoc for `IParameterConsumer` to mention that `IParameterPreprocessor` is a more powerful and flexible alternative.

docs/index.adoc
remkop: That said, there are several other places in the manual where backticks are used for class names in headings.
Maybe best to keep the backticks.

deining:
> we need to link to their top-level pages from the user manual and project README.

As an addition, we could develop/author an overview page that contains links to the apidocs of all subprojects. Once we made this overview page avaible at the subdomain apidocs.picocli.info, this URL could serve as single starting point for accessing all apidocs (just a proposal/ an idea).

remkop:
I created https://github.com/remkop/picocli/issues/1298 and put some ideas on how to deal with the Javadoc for the other modules there, let's discuss further on that ticket.

docs/index.adoc
deining: I never wanted to removed the backticks. They simply were a hindrance to use the section title as cross reference target, that's why I removed them. Fortunaltely, I could overcome this difficulty, and the backticks are now in place again.

deining:
I resolved all issues from your review meanwhile.
Thanks for considering my PR.

remkop (approved):

remkop:
Merged and published the HTML. Many thanks!

PR #1313, opened 2021-01-23
Bump jline3Version in order to avoid stackoverflow error

rupert-madden-abbott:
Using picocli-shell-jline3:4.6.1, and copying the example containing [this README](https://github.com/remkop/picocli/tree/master/picocli-shell-jline3) exactly leads to a stack overflow error. Somebody else [reported this issue to jline](https://github.com/jline/jline3/issues/640) where they have recommended bumping the jline version to 3.18.0.

I've tested locally, encountered the  stack overflow, then bumped the version and confirmed the stack overflow is fixed.


remkop:
@rupert-madden-abbott thank you for the pull request!
This looks like a good change, and I am happy to merge it, but should we also update the JLine version in the README?

rupert-madden-abbott:
@remkop Thanks very much.

I've had a look at the README and, personally, I don't think maintaining a version compatibility range in a README is a good idea unless you have tests (automated or otherwise) which are specifically testing picocli-shell-jline3 versions against different jline versions.

I'm not sure when this stackoverflow issue was introduced. It may have only been in jline 3.17.1. In that case, the latest version of picocli-shell-jline3 may be compatible with both 3.17.0 (and less) and 3.18.0 (and more). But I haven't tested those versions so I can't say for sure.

All I know is that version 3.17.1 should not be used. If you are happy that the version compatibility table is otherwise correct, perhaps we could just add a note saying to not use 3.17.1 with a link to this PR or to the jline issue?

remkop:
I like your suggestion. 
Let’s do that.

rupert-madden-abbott:
@remkop awesome, I've pushed an update. See you what you think.

remkop:

PR #1314, opened 2021-01-25
Fix use of deprecated Maven properties

charphi:
Fix the following warnings in Maven when using configuration specified in codegen documentation:
```
The expression ${groupId} is deprecated. Please use ${project.groupId} instead.
The expression ${artifactId} is deprecated. Please use ${project.artifactId} instead.
It is highly recommended to fix these problems because they threaten the stability of your build.
For this reason, future Maven versions might no longer support building such malformed projects.
```


remkop:

remkop:
Release notes updated.

PR #1321, opened 2021-02-03
Fix bad encoding when creating default PrintWriter

### LINKED ISSUE #1320: Bad encoding on default picocli.CommandLine#getOut and picocli.CommandLine#getErr
Depending on your console settings, the default `PrintWriter` of `picocli.CommandLine#getOut` may use a bad encoding.

Picocli code:
```java
public PrintWriter getOut() {
    if (this.out == null) {
        this.setOut(new PrintWriter(System.out, true));
    }

    return this.out;
}
```

This code creates a `PrintWriter` using `Charset.defaultCharset()` which can be different from the current console encoding.  
A possible solution would be to use `Charset.forName(System.getProperty("sun.stdout.encoding"))`.
See https://github.com/keerath/openjdk-8-source/blob/5f6e9d42a9f6b6736100c9c6f43f5f5ea1570cfb/jdk/src/share/classes/java/lang/System.java#L1189

Here is a code to reproduce (Windows 10, command prompt, encoding cp437):

```java
///usr/bin/env jbang "$0" "$@" ; exit $?
//REPOS jitpack
//DEPS com.github.remkop.picocli:picocli:4.6.1
//DEPS com.github.remkop.picocli:picocli-codegen:4.6.1

import picocli.CommandLine;

import java.io.OutputStreamWriter;
import java.nio.charset.Charset;
import java.util.concurrent.Callable;

@CommandLine.Command(name = "EncodingIssue")
class EncodingIssue implements Callable<Void> {

    @CommandLine.Spec
    private CommandLine.Model.CommandSpec spec;

    public static void main(String... args) {
        System.exit(new CommandLine(new EncodingIssue()).execute(args));
    }

    @Override
    public Void call() throws Exception {
        String problem = "[abcµ]" + System.lineSeparator();
        // ko
        spec.commandLine().getOut().append("1" + spec.commandLine().getColorScheme().errorText(problem)).flush();
        spec.commandLine().getOut().append("2" + problem).flush();
        new OutputStreamWriter(System.out, Charset.defaultCharset()).append("3" + problem).flush();
        // ok
        System.out.append("4" + problem).flush();
        new OutputStreamWriter(System.out, getStdoutEncoding()).append("5" + problem).flush();
        return null;
    }

    static Charset getStdoutEncoding() {
        String charsetName = System.getProperty("sun.stdout.encoding");
        if (charsetName != null) return Charset.forName(charsetName);
        return Charset.defaultCharset();
    }
}
```

charphi:
Fix #1320

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -1206,7 +1206,7 @@ public class CommandLine {
      * help with a {@code --help} or similar option, the usage help message is printed to the standard output stream so that it can be easily searched and paged.</p>
      * @since 4.0 */
     public PrintWriter getOut() {
-        if (out == null) { setOut(new PrintWriter(System.out, true)); }
+        if (out == null) { setOut(newPrintWriter(System.out, getStdoutEncoding())); }
         return out;
     }
 
@@ -1233,7 +1233,7 @@ public class CommandLine {
      * should use this writer to print error messages (which may include a usage help message) when an unexpected error occurs.</p>
      * @since 4.0 */
     public PrintWriter getErr() {
-        if (err == null) { setErr(new PrintWriter(System.err, true)); }
+        if (err == null) { setErr(newPrintWriter(System.err, getStderrEncoding())); }
         return err;
     }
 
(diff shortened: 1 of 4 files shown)


remkop:
Thanks for the PR!
Merged and documented in the release notes.

PR #1328, opened 2021-02-20
Upgrade Gradle to 6.8.2 and enable Official Gradle Wrapper Validation GitHub Action

Fiouz:
Official Gradle Wrapper Validation Action enabled to ensure binary wrapper is not tampered with.

See: https://github.com/marketplace/actions/gradle-wrapper-validation


remkop:

Issue #1227, opened 2020-10-25
option to generate single html/man page


maxandersen:
would be awesome if the man page generator would be able to generate a single html page instead of *per* subcommand. The current end result ends up being very fragmented.

reason: https://github.com/jbangdev/jbang/pull/444

remkop:
This looks very interesting and I am certainly open to improving this. 

Anyone able to provide a pull request?

remkop:
I haven’t given any thought to what the structure of the generated AsciiDoc would become. Simple concatenation?
What about nested sub-subcommands?

What about links, simply make all command names links to other sections in the same document? Do subcommands link back to their parent?

Anything else worth considering that doesn’t come up for single command doc pages?

maxandersen:
Includes would work as long as follow same convention. 

My thinking is to get to something like manual at cli.github.com.

remkop:
Looking at https://cli.github.com/manual/ , I see each command and subcommand has a separate page, and the Table of Contents (TOC) navigation bar on the left brings them all together. The TOC especially is nice: it expresses the command hierarchy and is intuitive to navigate. 

One way I can see this can be done with AsciiDocter is to have a single document for each top-level command. This doc can have a section for each subcommand, sub-sections for each sub-subcommand, etc. 
AsciiDocter's [table of contents](https://asciidoctor.org/docs/user-manual/#user-toc) feature would generate the navigation bar on the left.
The contents of that document could be just titles for each section and subsection, and the contents of each of these sections could be an include that points to the file currently generated by `picocli-codegen`.

@grumpyf0x48 I see you have been working on https://github.com/jbangdev/jbang/pull/444, would you be interested in upstreaming some of your work to do this in picocli-codegen directly?

remkop:
Note that https://github.com/remkop/picocli/pull/1340 will improve this a little bit in the upcoming picocli release.

bentatham:
This would be great for a single PDF file, as well as single-page HTML.

remkop:
@bentatham Do you feel like providing a pull request for this?

PR #1340, opened 2021-03-04
add xref for subcommands

crotwell:
subcommands get separate html files, nice if html generated from asciidoc could link them.

Added an xref for each subcommand and updated tests. I believe an xref like this doesn't hurt anything, and helps with linking is possible (html and pdf). That said, I know just enough asciidoc to be dangerous, so...


remkop:
Merged.
Thank you for the contribution! I really like this, and suspect there are many other parts of the document that can be linked this way. For example:

* provide a link to the parent command (maybe from the synopsis)
* provide links to the options from the synopsis
* ... other ideas?

Also related to https://github.com/remkop/picocli/issues/1227 in the sense of improving usability of the generated manpage docs.

Issue #1359, opened 2021-04-11
Add a doc section for packaging with JReleaser


aalmiray:
Follow up to #727

aalmiray:
I suppose this issue may be closed now.

PR #1360, opened 2021-04-11
[#1359] DOC: add JReleaser as packaging option

aalmiray:


remkop:
Merged.
I will update the release notes shortly.
The rendered HTML will be published with the next picocli release at the latest, if I don't get around to publishing them earlier.

PR #1362, opened 2021-04-20
Fix #1351

### LINKED ISSUE #1351: completionCandidates() called during help output even if not going to be used
I have some cli commands that interact with the server. One of parameters has a completionCandidates setting that talks to the server to figure out the list of possible values, which can be dynamic. It looks like when I ask for "--help" of the command it also invokes the completionCandidates, even though the result is not going to be used. It gets it just in case it will need to use in the description if `"${COMPLETION-CANDIDATES}"` is specified. But in my case it's not used in description.  My situation is even worse, sometimes I have no connection to the server, and therefore I'm unable to display the command help at all.

It would be nice, and more optimal in case of expensive operation, if the completionCandidates for help of commands was invoked only if needed. I.e. when `"${COMPLETION-CANDIDATES}"` is actually specified in description.

The code in question is in `picocli.CommandLine.Model.ArgSpec.expandVariables(String[])`

wtd2:
Hello, I'm with @sustc11810424 and @Lanninger08 to fix issue #1351.

If we've understood it right, we just need to avoid iterating through `completionCandidates` unnecessarily by first checking any existence of `${COMPLETION-CANDIDATES}`.

A very simple solution with a test case is presented in the PR.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -8680,16 +8680,17 @@ public class CommandLine {
             private String[] expandVariables(String[] desc) {
                 if (desc.length == 0) { return desc; }
                 StringBuilder candidates = new StringBuilder();
-                if (completionCandidates() != null) {
-                    boolean isCompletionCandidatesUsed = false;
-                    for (String s: desc) {
-                        if (s.contains(DESCRIPTION_VARIABLE_COMPLETION_CANDIDATES)) {
-                            isCompletionCandidatesUsed = true;
-                            break;
-                        }
+                boolean isCompletionCandidatesUsed = false;
+                for (String s: desc) {
+                    if (s.contains(DESCRIPTION_VARIABLE_COMPLETION_CANDIDATES)) {
+                        isCompletionCandidatesUsed = true;
+                        break;
                     }
-                    if (isCompletionCandidatesUsed){
-                        for (String c : completionCandidates()) {
+                }
+                if (isCompletionCandidatesUsed) {
+                    Iterable<String> iter = completionCandidates();
+                    if (iter != null) {
+                        for (String c : iter) {
                             if (candidates.length() > 0) { candidates.append(", "); }
                             candidates.append(c);
                         }
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -8680,7 +8680,16 @@ public class CommandLine {
             private String[] expandVariables(String[] desc) {
                 if (desc.length == 0) { return desc; }
                 StringBuilder candidates = new StringBuilder();
-                if (completionCandidates() != null) {
+
+                boolean flag = false;
+                for (String s: desc) {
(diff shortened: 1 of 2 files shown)


src/main/java/picocli/CommandLine.java
```diff
@@ -8680,7 +8680,16 @@ public boolean required() {
             private String[] expandVariables(String[] desc) {
                 if (desc.length == 0) { return desc; }
                 StringBuilder candidates = new StringBuilder();
-                if (completionCandidates() != null) {
+
+                boolean flag = false;
```
remkop, line 8684: Can we rename `flag` to `isCompletionCandidatesUsed`?

src/test/java/picocli/Issue1351.java
```diff
@@ -0,0 +1,60 @@
...
+    class TestCommand {
+        @CommandLine.Option(names = "-o", completionCandidates = MyIterable.class,
+                description = "Candidates: A, B, C")
+        String option;
+    }
+
+    @Test
+    public void testIssue1351() {
+        CommandLine.usage(new TestCommand(), System.out);
+        assertEquals(0, flag);
```
remkop, line 58: This is a good test to verify that MyIterable is not invoked if the description does not contain `${COMPLETION-CANDIDATES}`.

Can we also add a test where we verify the opposite: that MyIterable _is_ invoked when the description _does_ contain `${COMPLETION-CANDIDATES}`?

remkop (changes requested):
Thank you for the PR!
This looks pretty good. I could merge this as is, but can you take a look at my comments? 
There are some small enhancements that would make it even better.

wtd2:
Hi @remkop, thanks for your kind advice. We have modified the codes according to your suggestions.

src/main/java/picocli/CommandLine.java
```diff
@@ -8681,9 +8681,18 @@ public boolean required() {
                 if (desc.length == 0) { return desc; }
                 StringBuilder candidates = new StringBuilder();
                 if (completionCandidates() != null) {
```
remkop, line 8683: This call to `completionCandidates()` on line 8683 may also be slow...
It may be better to postpone this call until after we check that `isCompletionCandidatesUsed == true`.

Also, it may be good to avoid calling `completionCandidates()` twice, so how about doing something like this:

```java
if (isCompletionCandidatesUsed) {
    Iterable<String> iter = completionCandidates();
    if (iter != null) {
        for (String c : iter) {
            ....
```

remkop (changes requested):
Almost there, I found one more optimization, please check...

src/main/java/picocli/CommandLine.java
wtd2: Thank you @remkop, and we have updated according to the advice.

remkop:
Merged. Thank you for the contribution! :+1: 

wtd2:
Thanks @remkop for merging.

remkop:
I updated the release notes now.
I mentioned all of you (@wtd2,  @sustc11810424 and @Lanninger08), I hope that is okay.

PR #1364, opened 2021-04-26
Variable interpolation in Kotlin

MagnusMG:
Changed the user manual (index.adoc) by adding a caution block in the section about variable interpolation.


remkop:
Merged and updated release notes.
The rendered HTML will be published with the next picocli release.

PR #1370, opened 2021-04-30
fix #1366 point 1 (the help command arity)

### LINKED ISSUE #1366: few points from a picocli beginner
Hi,

Moving from commons-cli to picocli both using annotation and programmatic API, it is really nice!

1. Using something very similar to the documentation regarding the add of the help subcommand, I got an 'help help' message like the following:

```
Displays help information about the specified command

Usage: Main help [-h] [COMMAND...]

When no COMMAND is given, the usage help for the main command is displayed.
If a COMMAND is specified, the help for that command is shown.

      [COMMAND...]   The COMMAND to display the usage help message for.
  -h, --help         Show usage help for the help command and exit.
````

Why displaying `[COMMAND...]` and not just `[COMMAND]`, does help for multiple commands in one run possible?
Because in my simple test case, it always displays help for the first command only.
2. Could it be a long-term dev to have a third approach using an externalized declaration file similar to what is done in JPA/ORM for which can be used annotations, runtime/programmatic API, and XML (or another format)? Is there any standard that could be interesting to look at for this? (in the scope of CWL?)
3. I don't despair of making my case with many subcommands works using only the programmatic API!!
 
Thanks

### PR DESCRIPTION
**Pull request by peutch**
Hi, I hope it is good enough.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -14753,7 +14753,7 @@ public class CommandLine {
 
         @Parameters(paramLabel = "COMMAND", arity="0..1", descriptionKey = "helpCommand.command",
                 description = "The COMMAND to display the usage help message for.")
-        private String command;
+        private String commands;
 
         private CommandLine self;
         private PrintStream out;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/main/java/picocli/CommandLine.java [hunk 1]
```diff
@@ -14751,9 +14751,9 @@ private BuiltIn() {} // private constructor: never instantiate
                 description = "Show usage help for the help command and exit.")
         private boolean helpRequested;
 
-        @Parameters(paramLabel = "COMMAND", descriptionKey = "helpCommand.command",
+        @Parameters(paramLabel = "COMMAND", arity="0..1", descriptionKey = "helpCommand.command",
                 description = "The COMMAND to display the usage help message for.")
-        private String[] commands = new String[0];
+        private String command;
```
remkop, line 14756: Let's keep the name `commands` unchanged, and only change the type from `String[]` to `String`.

That will allow previously generated `reflect.json` files to continue to work with both older and newer versions of picocli.

remkop (changes requested):
This is a good change, thank you for the pull request.
I would like to make a small change for backwards compatibility, can you take a look?

src/main/java/picocli/CommandLine.java
peutch: Of course, I'll be reviewing my copy of Plural Logic book! ;-)
What is more would it have been possible to keep String[] and just set the arity?
I did not try.

remkop:
Merged. Thank you for the contribution. Apologies for the delay.

(Note to self: TODO update release notes)

PR #1382, opened 2021-06-07
fixes  #1381

### LINKED ISSUE #1381: Default value of option in repeated subcommand not applied correctly
Say I have the following command in which mysubcommand is a repeatable subcommand:
`mycommand mysubcommand param mysubcommand -t test param`
In this case the first subcommand's default value for the -t option is not set correctly by Interpreter.applyDefaultValues(). The reason seems to be that parsing the second subcommand adds the -t option to the set of initialized arguments: `Set<ArgSpec> initialized`. This set is then also used when setting the default values of the first subcommand. Thus  `applyDefaultValues()` falsely believes it is already initialized.

sfeuerhahn:
This is an attempt to fix #1381 . It fixed my use case but I am not 100% sure if it raises other issues.


remkop:
Merged.
This will be included in the next picocli release.

sfeuerhahn:
Thank you for this really great library!

Issue #1398, opened 2021-07-29
Behaviour doesn't match documentation for boolean Option arity


joca-bt:
https://picocli.info/apidocs/picocli/CommandLine.Option.html#arity-- says:

- By default, flags (boolean options) have arity zero
- By default picocli does not expect boolean options (also called "flags" or "switches") to have a parameter.

However specifying a parameter is not resulting in error by default:

```
@Command(name = "test")
public class Main implements Runnable {
    @Option(names = { "--option" })
    private boolean option;

    @Override
    public void run() {
    }

    public static void main(String[] args) {
        int exitCode = new CommandLine(new Main()).execute(args);
        System.exit(exitCode);
    }
}
```

The error only appears when we explicitly add `arity = "0"` to the above `@Option` declaration:

`option '--option' (<option>) should be specified without 'true' parameter`

 which, by the documentation, should already be the default.

remkop:
You are correct, the documentation is incorrect.
The arity for boolean options is actually `"0..1"` now.

Do you feel like contributing a pull request for this?

joca-bt:
Revert the code behaviour or doc?

remkop:
We’ll change the documentation. 
There’s no reason to break existing applications that depend on the current behaviour. :-)

PR #1400, opened 2021-08-01
Documentation: Update boolean arity

joca-bt:
The default arity for boolean options is 0..1 and no longer 0.

Closes https://github.com/remkop/picocli/issues/1398.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -3734,7 +3734,7 @@ public class CommandLine {
          * command line, a {@link MissingParameterException} is thrown by the {@link #parse(String...)} method.
          * <p>
          * In many cases picocli can deduce the number of required parameters from the field's type.
-         * By default, flags (boolean options) have arity zero,
+         * By default, flags (boolean options) have arity "0..1",
          * and single-valued type fields (String, int, Integer, double, Double, File, Date, etc) have arity one.
          * Generally, fields with types that cannot hold multiple values can omit the {@code arity} attribute.
          * </p><p>
@@ -3746,7 +3746,7 @@ public class CommandLine {
          * </p>
          * <b>A note on boolean options</b>
          * <p>
-         * By default picocli does not expect boolean options (also called "flags" or "switches") to have a parameter.
+         * By default picocli allows boolean options (also called "flags" or "switches") to have an optional parameter.
          * You can make a boolean option take a required parameter by annotating your field with {@code arity="1"}.
          * For example: </p>
          * <pre>&#064;Option(names = "-v", arity = "1") boolean verbose;</pre>
@@ -3756,12 +3756,11 @@ public class CommandLine {
          * on the command line, or a {@link MissingParameterException} is thrown by the {@link #parse(String...)}
          * method.
          * </p><p>
-         * To make the boolean parameter possible but optional, define the field with {@code arity = "0..1"}.
+         * To remove the optional parameter, define the field with {@code arity = "0"}.
          * For example: </p>
-         * <pre>&#064;Option(names="-v", arity="0..1") boolean verbose;</pre>
-         * <p>This will accept any of the below without throwing an exception:</p>
+         * <pre>&#064;Option(names="-v", arity="0") boolean verbose;</pre>
+         * <p>This will reject any of the below:</p>
          * <pre>
-         * -v
          * -v true
          * -v false
          * </pre>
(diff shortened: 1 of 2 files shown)


joca-bt:
There might be some doc missing. @remkop

src/main/java/picocli/CommandLine.java
```diff
@@ -5603,8 +5593,8 @@ static Range adjustForType(Range result, IAnnotatedElement member) {
             return result.isUnspecified ? defaultArity(member) : result;
         }
         /** Returns the default arity {@code Range}: for interactive options/positional parameters,
-         * this is 0; for {@link Option options} this is 0 for booleans and 1 for
-         * other types, for {@link Parameters parameters} booleans have arity 0, arrays or Collections have
+         * this is 0; for {@link Option options} this is "0..1" for booleans and 1 for
+         * other types, for {@link Parameters parameters} booleans have arity 1, arrays or Collections have
```
joca-bt, line 5597: According to this table https://picocli.info/#_default_arity. For boolean parameters it should be 1. Which one is correct @remkop?

remkop:
Some background on how this is implemented:
`OptionSpec` (and `PositionalParamSpec`) have a field called `arity` of type `Range`.

If a positional parameter is defined without explicitly setting the arity, picocli automatically assigns arity="1".
If a named option is defined without explicitly setting the arity, picocli automatically assigns arity="0" for booleans, and "1" for any other type. (Even though, as you found, _effectively_ the arity for booleans is "0..1". Keep reading...)
However, in both cases, the arity Range is marked as "unspecified=true".

During parsing, picocli creates a "derived arity" for each option it recognizes: named options with a value attached to the option name will have derived arity="1". For example, `--file=/tmp/abc`, or `-fMyFile`.

Picocli uses this "derived arity" if the option was defined without an explicit arity (so when "unspecified=true").

Effectively, this means that from a validation point of view, booleans have arity="0..1".
(It kind of grew like this over time... Perhaps this can be simplied, but would not want to impact behaviour...)

Anyway, that is why the docs say that arity=0 for booleans, when arity=0..1 is more accurate. (It does not tell the whole story, but I guess that is okay.)
And for positional parameters, arity must be "1", regardless of type.

src/main/java/picocli/CommandLine.java
remkop: You are correct: for boolean positional parameters, the `defaultArity` method returns `Range.valueOf("1").unspecified(true)`.

For named boolean options, the story is a bit more complex.
Strictly speaking, the `defaultArity` method returns `Range.valueOf("0").unspecified(true)`.  _Effectively_, that is "0..1". Shall we change this javadoc to something like this?

> (...); for {@link Option options} this is "0" for booleans (but effectively "0..1" since {@code unspecified} is set to {@code true}) and 1 for other types, for {@link Parameters parameters} booleans have arity 1 (...)

joca-bt: That seems quite a bit of implementation details, not sure it should be in the Javadoc. I'm not sure what would be the best way of phrasing that to put in the documentation. Any suggestions? Feel free to change the MR at will.

remkop: OK, another idea is to use your phrasing, with one small change, add the word _effectively_:

> this is effectively `"0..1"` for booleans

And then we can add an Implementation Notes section to the Javadoc for that method, like this:

```
<p><b><em>Implementation Notes</em></b></p>
<p>The returned {@code Range} for boolean options has an <em>effective</em> arity of "0..1".
This is implemented by returning a {@code Range} with arity "0", 
and its {@code unspecified} property set to {@code true}. 
This implementation may change in the future.</p>
```

Thoughts?

src/main/java/picocli/CommandLine.java
```diff
@@ -3746,22 +3746,12 @@ private static String format(String formatString, Object... params) {
          * </p>
          * <b>A note on boolean options</b>
          * <p>
-         * By default picocli does not expect boolean options (also called "flags" or "switches") to have a parameter.
-         * You can make a boolean option take a required parameter by annotating your field with {@code arity="1"}.
```
remkop, line 3750: I would like to keep this sentence:

> You can make a boolean option take a required parameter by annotating your field with {@code arity="1"}.

src/main/java/picocli/CommandLine.java
```diff
@@ -3746,22 +3746,12 @@ private static String format(String formatString, Object... params) {
...
+         * By default picocli allows boolean options (also called "flags" or "switches") to have an optional parameter.
```
remkop, line 3749: Small addition:

> (...) an optional parameter, which must be either `true` or `false`.

src/main/java/picocli/CommandLine.java
```diff
@@ -3746,22 +3746,12 @@ private static String format(String formatString, Object... params) {
...
-         * <p>
-         * Because this boolean field is defined with arity 1, the user must specify either {@code <program> -v false}
-         * or {@code <program> -v true}
-         * on the command line, or a {@link MissingParameterException} is thrown by the {@link #parse(String...)}
-         * method.
-         * </p><p>
-         * To make the boolean parameter possible but optional, define the field with {@code arity = "0..1"}.
-         * For example: </p>
-         * <pre>&#064;Option(names="-v", arity="0..1") boolean verbose;</pre>
-         * <p>This will accept any of the below without throwing an exception:</p>
```
remkop, line 3762: I would like to keep this example. Let's have an example for both arity=1 and arity=0.

remkop (changes requested):
I think we are almost there.
I have a few suggested changes. Can you take a look?

src/main/java/picocli/CommandLine.java
remkop: Great, thank you!

src/main/java/picocli/CommandLine.java
joca-bt: I think this one is very redundant, it's a _boolean_ option.

remkop: I see what you are saying, but I think it makes sense to be very specific about what parameters are allowed. For example, the following parameters would be rejected: TRUE, FALSE, True, False, 1, 0, YES, NO, Yes, No. 

Contrast this to some languages like Groovy and Python where Strings with these values _are_ translated to true/false boolean values.

remkop:
Merged. 
Thank you for the pull request and thank you for your patience!

PR #1404, opened 2021-08-05
Print paramLabel only when it could exist

joca-bt:


remkop:
Can you give some detail about what you are currently seeing that is different from what you expected to see? Ideally with an example that can be reproduced. 

Also, I’m a bit surprised that all tests still pass with this change. (Away from my PC, cannot check.) Does that mean that this area isn’t covered by any tests? Maybe we should take this opportunity to add tests if that’s the case.

joca-bt:
Before fix:
`option '--help' (<help>) should be specified only once`

After fix:
`option '--help' should be specified only once`

remkop:
Thanks for the clarification. 
Can we add some tests to the pull request?

joca-bt:
Do you mind adding them? Won't have time to do that anytime soon.

remkop:
> Do you mind adding them? Won't have time to do that anytime soon.

I currently have almost no time to spend on picocli. 
If you don't have time either that is fine of course. 
This will just have to wait until someone can add some tests.

remkop:

Issue #1352, opened 2021-03-18
`AutoComplete` fails with NPE if message bundle key for command name is used


bobtiernay-okta:
If the following is used for a command:
```java
@Command(name = "${project.cli.command}", ...)
```

`AutoComplete` will fail with:
```
java.lang.NullPointerException
	at picocli.AutoComplete.generateFunctionCallsToArrContains(AutoComplete.java:563)
	at picocli.AutoComplete.generateEntryPointFunction(AutoComplete.java:544)
	at picocli.AutoComplete.bash(AutoComplete.java:471)
	at picocli.AutoComplete.bash(AutoComplete.java:436)
	at picocli.AutoComplete$App.call(AutoComplete.java:186)
	at picocli.AutoComplete$App.call(AutoComplete.java:96)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1953)
	at picocli.CommandLine.access$1300(CommandLine.java:145)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2352)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2346)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2311)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2179)
	at picocli.CommandLine.execute(CommandLine.java:2078)
	at picocli.AutoComplete.main(AutoComplete.java:74)
Usage: picocli.AutoComplete [-fhVw] [-c=<factoryClass>] [-n=<commandName>]
                            [-o=<autoCompleteScript>] [@<filename>...]
                            <commandLineFQCN>
Generates a bash completion script for the specified command class.
      [@<filename>...]       One or more argument files containing options.
      <commandLineFQCN>      Fully qualified class name of the annotated
                               `@Command` class to generate a completion script
                               for.
  -c, --factory=<factoryClass>
                             Optionally specify the fully qualified class name
                               of the custom factory to use to instantiate the
                               command class. When omitted, the default picocli
                               factory is used.
  -n, --name=<commandName>   Optionally specify the name of the command to
                               create a completion script for. When omitted,
                               the annotated class `@Command(name = "...")`
                               attribute is used. If no `@Command(name = ...)`
                               attribute exists, '<CLASS-SIMPLE-NAME>' (in
                               lower-case) is used.
  -o, --completionScript=<autoCompleteScript>
                             Optionally specify the path of the completion
[...]

PR #1430, opened 2021-09-11
[#1388] Fix subcommand aliases autocomplete regression

### LINKED ISSUE #1388: 4.3.2->4.6.1 regression, incorrect autocompletion is generated for subcommand aliases, because of which autocomplete does not work for subcommand specified by alias.
TLDR autocompletion does not work for aliases.

Command defined as:

```
@CommandLine.Command(description = "Receive avro messages from kafka",
        usageHelpWidth = 120,
        name = "receive-avro-from-kafka",
        aliases = {"rafk"})
public class ReceiveAvroMessagesFromKafka extends JsonLikeKafkaMessageReceiver implements Callable<Integer> {
```

will generate autocompletion:

```
function _complete_campTools() {
...
  local cmds4=(receive-avro-from-kafka)
  local cmds5=(receive-avro-from-kafka)
...  
  if CompWordsContainsArray "${cmds5[@]}"; then _picocli_campTools_rafk; return $?; fi
  if CompWordsContainsArray "${cmds4[@]}"; then _picocli_campTools_receiveavrofromkafka; return $?; fi
...
```

_complete_campTools() is top level function. My interpretation is, that cmds4 and cmds5 are generated duplicated. Both subcomand autocomplete functions are equal, but we test presence of wrong word in COMP_WORDS array, meaning that we cannot ever call rafk alias related function. Fixing it to `local cmds5=(rafk)` fixes the issue and autocomplete works again.

Not sure if that's enough to find the issue, or whether some sub-standard coding on my part causes that. If you need smth more, I can help / share project.

NewbieOrange:
This seems to fix #1388 regression, and I have added a new test for the issue.

However, this cause #1352 `testIssue1352_SubcommandNameResourceBundle` to fail, which seemed to be caused by `descriptor.commandName` is `${project.cli.command}`, and it's from `commandLine.getSubcommands().entrySet().?.getKey()`, indicating there might be some issue in resource bundle propagation for subcommands.

@remkop any thoughts?

### DIFF
--- a/src/main/java/picocli/AutoComplete.java
+++ b/src/main/java/picocli/AutoComplete.java
@@ -560,7 +560,7 @@ public class AutoComplete {
             int count = functionCalls.size();
             CommandSpec spec = descriptor.commandLine.getCommandSpec();
             String full = spec.qualifiedName(" ");
-            String withoutTopLevelCommand = full.substring(spec.root().name().length() + 1);
+            String withoutTopLevelCommand = full.substring(spec.root().name().length() + 1).replace(spec.name(), descriptor.commandName);
 
             functionCalls.add(format("  if CompWordsContainsArray \"${cmds%2$d[@]}\"; then %1$s; return $?; fi\n", descriptor.functionName, count));
             buff.append(      format("  local cmds%2$d=(%1$s)\n", withoutTopLevelCommand, count));
--- a/src/main/java/picocli/AutoComplete.java
+++ b/src/main/java/picocli/AutoComplete.java
@@ -560,7 +560,8 @@ public class AutoComplete {
             int count = functionCalls.size();
             CommandSpec spec = descriptor.commandLine.getCommandSpec();
             String full = spec.qualifiedName(" ");
-            String withoutTopLevelCommand = full.substring(spec.root().name().length() + 1).replace(spec.name(), descriptor.commandName);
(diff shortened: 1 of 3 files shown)


NewbieOrange:
By adding a workaround in `createSubHierarchy(String scriptName, CommandLine commandLine, List<CommandDescriptor> out)` I can get `testIssue1352_SubcommandNameResourceBundle` to pass, but it is a weird solution.

```java
private static void createSubHierarchy(String scriptName, CommandLine commandLine, List<CommandDescriptor> out) {
// ...
    String commandName = entry.getKey(); // may be an alias
    if (commandName.startsWith("$")) {
        commandName = spec.resourceBundle().getString(commandName.substring(2, commandName.length() - 1));
    }
// ...
}
```

NewbieOrange:
The subcommands now get their name inerpolated in `addSubcommand()`, fixing #1352 test regression.

Not sure it is the correct solution - but all tests now pass.

NewbieOrange:
Still need some changes, not ready for merge yet.

remkop:
> Still need some changes, not ready for merge yet.

Okay, understood. Thank you for looking at this!

NewbieOrange:
> Okay, understood. Thank you for looking at this!

Fixed the issue now.

remkop:
Apologies for the long wait. I will try to look at this soon!

PR #1431, opened 2021-09-13
Whou/fix window to unix newline

wenhoujx:
fix a bunch of `^M` window newline chars.


remkop:
Newlines for HTML should be automatically converted from CRLF to LF when committing to git.
I updated the `.gitattributes` file accordingly.

Thank you for pointing this out!

wenhoujx:
oh, TIL. thanks!

PR #1467, opened 2021-11-17
[#1466] Fix autocomplete for aliased subcommands

### LINKED ISSUE #1466: Autocompletion script doesn't properly handle aliases
I have a command class similar to the following, with sub-commands being programmatically added to this command using `commandLine.addSubcommand()`. 

```java
@Command(name = "retrieve", aliases = {"get"}, description = "Get data from various sources")
public class CompositeGetCommand {}
```
After sourcing the generated auto-completion script, something like `mycmd retrieve <TAB><TAB>` will correctly list the appropriate sub-commands. However, `mycmd get <TAB><TAB>` acts like the autocomplete script doesn't recognize the `get` alias, therefore listing the sub-commands of the `mycmd` parent command instead:

```
$ ./mycmd retrieve <TAB><TAB>
sub1 sub2 ...
$ ./mycmd get <TAB><TAB>
get retrieve ...
```

Looking at the auto-completion script, a duplicate cmd entry is generated for every alias, for example:

```bash
  ...
  local cmds7=(retrieve)
  local cmds8=(retrieve)
  ...
  local cmds77=(retrieve sub1)
  local cmds78=(retrieve sub2)
  ...
  local cmds100=(retrieve sub1)
  local cmds101=(retrieve sub2)
  ...
```

If I manually change this to the following, auto-completion starts working as expected:

```bash
  ...
  local cmds7=(retrieve)
  local cmds8=(get)
  ...
  local cmds77=(retrieve sub1)
  local cmds78=(retrieve sub2)
  ...
  local cmds100=(get sub1)
  local cmds101=(get sub2)
  ...
```

So, it seems like picocli is correctly recognizing aliases, but then incorrectly using the command name instead of alias when generating the autocomplete entries for these aliases.

---
Edit: The description above is for picocli 4.6.1. I just tried with picocli 4.6.2, where the behavior is slightly different with the following auto-complete script snippet being generated:

```bash
  ...
  local cmds7=(retrieve)
  local cmds8=(get)
  ...
  local cmds77=(retrieve sub1)
  local cmds78=(retrieve sub2)
  ...
  local cmds100=(retrieve sub1)
  local cmds101=(retrieve sub2)
  ...
```

With this, auto-completion works fine for both `mycmd retrieve` and `mycmd get`, but not for `mycmd get sub1` as `cmds100` and `cmds101` are still not using the proper alias name.

rsenden:
This PR fixes #1466


remkop:
This is great, thank you!
Can you add a test for this, or modify one of the existing tests?

rsenden:
I just added aliases to the existing AutoComplete tests, and updated the expected outputs. The only thing that I'm not sure about is why in AutoCompleteTest#testComplete() I need to repeat each alias three times. Any idea?

remkop:
> The only thing that I'm not sure about is why in AutoCompleteTest#testComplete() I need to repeat each alias three times. Any idea?

I will have to take a look in the debugger to see what is going on. It may take some time before I can get to this.

remkop:
I finally had some time to take a look at this.
It appears that the `AutoComplete#filterAndTrimMatchingPrefix` method does not filter out duplicates, even though it should. This can be fixed by modifying the first line of this method to use a `Set` instead of a `List`:

```java
    private static void filterAndTrimMatchingPrefix(String prefix, List<CharSequence> candidates) {
        Set<CharSequence> replace = new HashSet<CharSequence>();
        ...
```

The expectations in `testComplete` can then be changed to also have only one occurrence of each of the subcommand aliases, which makes much more sense.

Can you change the PR accordingly?

src/test/resources/picocompletion-demo-help_completion.bash
```diff
@@ -4,7 +4,7 @@
 # =======================
 #
 # Bash completion support for the `picocompletion-demo-help` command,
-# generated by [picocli](http://picocli.info/) version %1$s.
+# generated by [picocli](http://picocli.info/) version 4.6.3-SNAPSHOT.
```
remkop, line 7: Please leave the `%1$s` variable, it will be replaced by the latest version in the test.

src/test/resources/picocompletion-demo-help_completion.bash
```diff
@@ -275,7 +396,7 @@ function _picocli_picocompletion-demo-help_sub2_subsub2() {
   local commands=""
   local flag_opts=""
   local arg_opts="-u --timeUnit -t --timeout"
-  local timeUnit_option_args="%2$s" # --timeUnit values
+  local timeUnit_option_args="NANOSECONDS MICROSECONDS MILLISECONDS SECONDS MINUTES HOURS DAYS" # --timeUnit values
```
remkop, line 399: Please leave the `%2$s` variable, it will be replaced by the supported TimeUnit constants for the version of Java that is running the test.

src/test/resources/picocompletion-demo_completion.bash
```diff
@@ -4,7 +4,7 @@
 # =======================
 #
 # Bash completion support for the `picocompletion-demo` command,
-# generated by [picocli](http://picocli.info/) version %1$s.
+# generated by [picocli](http://picocli.info/) version 4.6.3-SNAPSHOT.
```
remkop, line 7: Please leave the `%1$s` variable, it will be replaced by the latest version in the test.

src/test/resources/picocompletion-demo_completion.bash
```diff
@@ -256,7 +377,7 @@ function _picocli_picocompletion-demo_sub2_subsub2() {
   local commands=""
   local flag_opts=""
   local arg_opts="-u --timeUnit -t --timeout"
-  local timeUnit_option_args="%2$s" # --timeUnit values
+  local timeUnit_option_args="NANOSECONDS MICROSECONDS MILLISECONDS SECONDS MINUTES HOURS DAYS" # --timeUnit values
```
remkop, line 380: Please leave the `%2$s` variable, it will be replaced by the supported TimeUnit constants for the version of Java that is running the test.

remkop (changes requested):
I also found some minor things with variables used in the test result templates that should be preserved. Can you take a look?

rsenden:
PR has been updated with all requested changes

remkop:
Merged. 

PR #1473, opened 2021-11-19
Fix incorrect autocomplete if file name contains space

### LINKED ISSUE #1458: Bad Autocomplete Output When Tab Completing a File With a Space
I am currently experiencing an issue with tab completion on an installed application running PicoCLI 4.6.1 and java 11 on Fedora 34.

Here is a snippet of the command line class for this project:
```
...
import picocli.CommandLine.*;
import picocli.CommandLine.Model.*;
...

@Command(name = "bcc")
public class BccCommands implements Runnable{

    @Spec
    CommandSpec spec;

    @Option(names = {"-s", "--snapshot"}, description = "Loads the specified snapshot file immediately upon launch.  Use full paths.")
    public File snapshotFile;

   ...
}

...
	_cmd = new BccCommands();
       	_cmdParser = new picocli.CommandLine(_cmd);
        _cmdParser.setCaseInsensitiveEnumValuesAllowed(true);
        _cmdParser.setOverwrittenOptionsAllowed(true);
...
```
I am currently generating the AutoComplete using Maven and picocli.AutoComplete to generate a bcc11_completion script.

With the generated completion script in `/etc/bash_completions.d/` I can  do the following:
```
bcc11 --snap<TAB><TAB>
```
and get
```
bcc11 --snapshot 
```
as expected

However when I try to add the file path and use tab completion to fill in the file path like this:
```
bcc11 --snapshot /home/zpater/Snapshots/2<TAB><TAB>
```
I get the following output:
image

but I would expect something similar to this:
image

If I tab complete on a combination that doesn't find a file with a space in it, I get the expected result:
image

There seems to be an issue with the completion script I generated and spaces. Is there anything I can do to resolve this issue?

NewbieOrange:
By setting the bash separator to `\n`, `COMPREPLY` now contains file names with spaces without splitting them, and can autocomplete correctly.

Screenshot1

After some preliminary testing, only files with line feed `\n` in their name cannot be autocompleted correctly, but `ls` seem to behave the same, so looks good to me.

Screenshot2

This pull request will close #1458.

### DIFF
--- a/src/main/java/picocli/AutoComplete.java
+++ b/src/main/java/picocli/AutoComplete.java
@@ -288,7 +288,7 @@ public class AutoComplete {
         final String commandName;
         final CommandLine commandLine;
         final CommandLine parent;
-        
+
         CommandDescriptor(String functionName, String commandName, CommandLine commandLine, CommandLine parent) {
             this.functionName = functionName;
             this.commandName = commandName;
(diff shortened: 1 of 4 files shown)


remkop:
Thank you for another great contribution, @NewbieOrange !
I am looking at another PR related to autocompletion first, and I will merge yours after that.
Thank you again!

PR #1475, opened 2021-11-22
Typo in annotation target-type error message

rgoldberg:
An annotation target-type error message has a typo that says "Cannot only process" when it should say "Can only process".


remkop:
Good catch! Thank you for the contribution! 👍 

Note to self: still need to update the release notes.

PR #1477, opened 2021-11-23
Remove execution and extension characters in the generated autocomplete script

andreaTP:
First of all, thanks for the amazing project and all the effort in it @remkop !
This is a follow up from an issue experienced in the Keycloak project:
https://github.com/keycloak/keycloak/pull/8875

Sorry for directly opening a PR instead of an Issue, but I do believe that this change is more self-explanatory in code 🙂 

Rationale:
- `./` breaks the autocomplete script on ZSH
- `.bash` and `.sh` autocompletion extensions are always automatically generated

Sanitizing the internally used `scriptName` sounds like a first step in the right direction, wdyt?


remkop:
@andreaTP Thank you for the contribution!
Looks good, I will merge this soon (looking at another completion-related PR first).

remkop:

andreaTP:
Thank you!

PR #1484, opened 2021-11-30
Fixed org.junit.Assert.assertThat deprecation warning

rgoldberg:
Fixed org.junit.Assert.assertThat deprecation warning


remkop:

TODO update release notes.

PR #1499, opened 2021-12-01
Improved DefaultFactory#create(Class<T>)

rgoldberg:
Improved `DefaultFactory#create(Class<T>)`:

- Only handle `cls` argument being a `Map` or subtype differently than normal if `cls` is an interface
- Create `TreeMap` if `cls` argument is `SortedMap` or sub-interface
- Simplified normal instantiation:
  - no longer call deprecated method
  - only try `setAccessible(true)` if it might make a difference

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -5505,12 +5505,10 @@ public class CommandLine {
                     return cls.cast(new LinkedHashMap<Object, Object>());
                 }
             }
+            Constructor<T> constructor = cls.getDeclaredConstructor();
             try {
-                @SuppressWarnings("deprecation") // Class.newInstance is deprecated in Java 9
-                T result = cls.newInstance();
-                return result;
-            } catch (Exception ex) {
-                Constructor<T> constructor = cls.getDeclaredConstructor();
+                return constructor.newInstance();
+            } catch (IllegalAccessException ex) {
                 constructor.setAccessible(true);
                 return constructor.newInstance();
             }
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -5505,12 +5505,19 @@ public class CommandLine {
                     return cls.cast(new LinkedHashMap<Object, Object>());
                 }
             }
-            Constructor<T> constructor = cls.getDeclaredConstructor();
             try {
-                return constructor.newInstance();
-            } catch (IllegalAccessException ex) {
-                constructor.setAccessible(true);
-                return constructor.newInstance();
+                @SuppressWarnings("deprecation") // Class.newInstance is deprecated in Java 9
+                T result = cls.newInstance();
+                return result;
+            } catch (Exception ex) {
+                // TODO log the error at debug level
+                Constructor<T> constructor = cls.getDeclaredConstructor();
+                try {
+                    return constructor.newInstance();
+                } catch (IllegalAccessException iaex) {
+                    constructor.setAccessible(true);
+                    return constructor.newInstance();
+                }
             }
         }
         private static ITypeConverter<?>[] createConverter(IFactory factory, Class<? extends ITypeConverter<?>>[] classes) {
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -32,7 +32,6 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.ByteOrder;
 import java.nio.charset.Charset;
-import java.nio.charset.UnsupportedCharsetException;
 import java.text.BreakIterator;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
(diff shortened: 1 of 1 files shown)


src/main/java/picocli/CommandLine.java
```diff
@@ -5484,32 +5483,32 @@ public Help create(CommandSpec commandSpec, Help.ColorScheme colorScheme) {
...
             }
+            Constructor<T> constructor = cls.getDeclaredConstructor();
             try {
-                @SuppressWarnings("deprecation") // Class.newInstance is deprecated in Java 9
-                T result = cls.newInstance();
-                return result;
-            } catch (Exception ex) {
-                Constructor<T> constructor = cls.getDeclaredConstructor();
+                return constructor.newInstance();
+            } catch (IllegalAccessException ex) {
```
remkop, line 5511: ...but I do like this idea of setting accessible=true only in the catch block of IllegalAccessException, but this try/catch should remain in the catch(Exception) block of `Class.newInstance`.

remkop (changes requested):
Thank you; can you look at my feedback and let me know your thoughts?

src/main/java/picocli/CommandLine.java
rgoldberg: Is this what you want?

```java
            try {
                @SuppressWarnings("deprecation") // Class.newInstance is deprecated in Java 9
                T result = cls.newInstance();
                return result;
            } catch (IllegalAccessException ex) {
                Constructor<T> constructor = cls.getDeclaredConstructor();
                constructor.setAccessible(true);
                return constructor.newInstance();
            }
```

src/main/java/picocli/CommandLine.java
```diff
@@ -5484,32 +5483,32 @@ public Help create(CommandSpec commandSpec, Help.ColorScheme colorScheme) {
...
-            if (Map.class.isAssignableFrom(cls)) {
-                try { // if it is an implementation class, instantiate it
-                    return cls.cast(cls.getDeclaredConstructor().newInstance());
-                } catch (Exception ignored) { }
-                return cls.cast(new LinkedHashMap<Object, Object>());
             }
+            Constructor<T> constructor = cls.getDeclaredConstructor();
             try {
-                @SuppressWarnings("deprecation") // Class.newInstance is deprecated in Java 9
-                T result = cls.newInstance();
```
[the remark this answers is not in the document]
rgoldberg: Just wondering why you prefer `Class::newInstance` over `Constructor::newInstance`.

I don't know of any reason to prefer the former over the latter except that it's less of a change, but maybe I'm missing something.

Whereas not using something that's deprecated because it bypasses compile-time exception checking seems like a good reason to prefer the latter over the former.

I'll do whatever you decide, I just want to learn about the pros & cons as I haven't used either of these methods extensively.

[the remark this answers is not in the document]
remkop: To be honest I don't remember exactly.
There was a situation where `Constructor.newInstance` gave some error, but `Class.newInstance` worked.
Unfortunately I don't remember the context or environment. 
It feel safer to leave it the way it is.

src/main/java/picocli/CommandLine.java
remkop: Almost, I had this in mind:

```java
try {
    @SuppressWarnings("deprecation") // Class.newInstance is deprecated in Java 9
    T result = cls.newInstance();
    return result;
} catch (Exception ex) {
    // TODO log the error at debug level
    Constructor<T> constructor = cls.getDeclaredConstructor();
    try {
        return constructor.newInstance();
    } catch (IllegalAccessException ex) {
        constructor.setAccessible(true);
        return constructor.newInstance();
    }
}
```
but not sure how much of an improvement this is.
The logging at DEBUG level when `Class.newInstance` fails feels like a nice improvement.

src/main/java/picocli/CommandLine.java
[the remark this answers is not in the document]
rgoldberg: Good enough reason to revert back to `Class` instead of `Constructor`. Will resubmit the PR sometime today with the change.

If you ever remember the situation where the latter errors while the former doesn't, please let me know so I can learn about the differences, too. Thanks.

src/main/java/picocli/CommandLine.java
rgoldberg: I can see how debugging can help learn the exact differences between the 2 methods. I will modify the PR later today & resubmit.

remkop: @rgoldberg Did you have a chance to look at this?

rgoldberg: Sorry. Couldn't look at this for a while. Will implement the changes soon.

src/main/java/picocli/CommandLine.java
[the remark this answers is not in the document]
rgoldberg: Pushed requested changes

src/main/java/picocli/CommandLine.java
rgoldberg: Pushed requested changes

rgoldberg:
I pushed the requested changes.

remkop:

PR #1518, opened 2021-12-16
Simplified detection of getter & setter in TypedMember constructor

rgoldberg:
Simplified detection of getter & setter in TypedMember constructor


remkop:
Hi @rgoldberg, just FYI: I am on the Apache Logging PMC and as you can imagine a bit preoccupied with the Log4j issue.
I will get back to this but just wanted to let you know that it may take a while. Thanks!

rgoldberg:
Yeah, you must be very busy.  No urgency.  Good luck with it.

I'm just cleaning up code that seemed awkward when I first read it.  I think the modifications should make it easier to understand, and incidentally improve performance (probably negligibly, but the straightforward code is my main aim, not massive performance gains).

remkop:
I finally had some time to look at this.

PR #1529, opened 2021-12-23
Allow aliases of a CommandSpec that is already a subcommand to be properly & consistently modified.

### LINKED ISSUE #1528: Modifying a `CommandSpec`'s aliases if it's already a subcommand
Currently, if `CommandSpec#aliases(String...)` is called to modify the aliases of a `CommandSpec` that is already registered as a subcommand of another `CommandSpec`, then the aliases registered in the parent are not updated.

The aliases in the parent should be updated to a) keep things consistent, and b) allow aliases to be modified.

rgoldberg:
Allow aliases of a CommandSpec that is already a subcommand to be properly & consistently modified.

Resolves #1528

Doesn't yet contain tests.  Wanted feedback before I implement tests.

Will add tests and whatever else is requested once the current state of the PR has been reviewed.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -7090,18 +7090,18 @@ public class CommandLine {
              * @return this CommandSpec for method chaining
              * @since 3.1 */
             public CommandSpec aliases(String... aliases) {
-                Set<String> existingAliasSet = this.aliases;
+                Set<String> previousAliasSet = this.aliases;
                 this.aliases = new LinkedHashSet<String>(Arrays.asList(aliases == null ? new String[0] : aliases));
                 if (parent != null) {
                     //remove & add aliases
-                    Set<String> newAliasSet = new LinkedHashSet<String>(this.aliases);
-                    newAliasSet.removeAll(existingAliasSet);
+                    Set<String> addedAliasSet = new LinkedHashSet<String>(this.aliases);
+                    addedAliasSet.removeAll(previousAliasSet);
                     Tracer t = new Tracer();
-                    for (String alias : newAliasSet) {
+                    for (String alias : addedAliasSet) {
                         parent.addAlias(alias, name, commandLine, t);
                     }
-                    existingAliasSet.removeAll(this.aliases);
-                    for (String alias : existingAliasSet) {
+                    previousAliasSet.removeAll(this.aliases);
+                    for (String alias : previousAliasSet) {
                         parent.removeAlias(alias, commandLine, t);
                     }
                 }
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -6428,9 +6428,7 @@ public class CommandLine {
                 if (subSpec.name == null) { subSpec.name(actualName); }
                 subSpec.parent(this);
                 for (String alias : subSpec.aliases()) {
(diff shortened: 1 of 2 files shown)


src/main/java/picocli/CommandLine.java
```diff
@@ -7081,7 +7090,21 @@ public String qualifiedName(String separator) {
              * @return this CommandSpec for method chaining
              * @since 3.1 */
             public CommandSpec aliases(String... aliases) {
+                Set<String> existingAliasSet = this.aliases;
                 this.aliases = new LinkedHashSet<String>(Arrays.asList(aliases == null ? new String[0] : aliases));
+                if (parent != null) {
+                    //remove & add aliases
+                    Set<String> newAliasSet = new LinkedHashSet<String>(this.aliases);
```
remkop, line 7097: Another minor detail: I would call this variable `addedAliasSet` since the intention is to subtract the old set so that only the _added_ alias remain, and then perform some operation on only the _added_ aliases.

remkop (approved):
I finally had time to look at this.
Looks good. Thank you for the contribution and your patience!

I added some minor feedback but I'm not too fussed, your proposed code is also fine.

Can you add some tests?
I would like to see a test that demonstrates the problem (fails with the current impl, and passes with the proposed change).

rgoldberg:
Thanks. Will look over feedback and modify the PR in the next few days.

Was planning to add tests (as per my initial description in this PR), just didn't want to make them until I had your feedback (e.g., if you wanted the existing behavior, my whole PR would never be accepted, so tests wouldn't ever be used, etc.)

remkop:
FYI, I will probably do the 4.6.3 release when this is merged. :-)

rgoldberg:
Sorry for delay. Will try to get to this soon. Thanks for heads up.

remkop:
Hi @rgoldberg, just wanted to check if this is still on your radar. :-)

remkop:
Hi @rgoldberg, if you don't think you will be able to work on this in the next week or so, that is not a problem: we can simply add it to the next release.
That said, I can wait a bit, I would like to include it, but I would like to do a picocli release in the next week or so.
Can you let me know your plans?

rgoldberg:
Will get it done this weekend. Sorry for delay.

src/main/java/picocli/CommandLine.java
```diff
@@ -7081,7 +7090,21 @@ public String qualifiedName(String separator) {
              * @return this CommandSpec for method chaining
              * @since 3.1 */
             public CommandSpec aliases(String... aliases) {
+                Set<String> existingAliasSet = this.aliases;
```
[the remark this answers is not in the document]
rgoldberg: Done

src/main/java/picocli/CommandLine.java
rgoldberg: Done

remkop:
@rgoldberg I can see your feedback on my comments, that you commented "Done", but I don't see any additional commits yet. Did you push your changes?

rgoldberg:
No. I committed the changes locally. Writing the test now. Will push all soon.

rgoldberg:
Added tests.  Fails on current master, but passes on this PR branch.

remkop (approved):
Looks great, thank you so much!

PR #1530, opened 2021-12-23
Simplified CommandSpec#validateSubcommandName(…)

rgoldberg:
Simplified CommandSpec#validateSubcommandName(…) by returning ASAP when the correct return value is first known.

It makes it much more understandable.


remkop:
Merged.

PR #1541, opened 2022-01-03
[#1537] AbbreviationMatcher now treats aliases of the same object as one match

### LINKED ISSUE #1537: Error: "value" not unique - when matching name and alias for same object
Found that having e.g. the `name` and `aliases` begin with the same letters result in strange bahaviour when it comes to allowing abbreviated commands. Say I have the following snippet out of the generated usage text;
```
COMMANDS
  chemical-files, chem-formats ...
```
and then running with `.. chem` will fail with the following stack trace (running `CommandLine.execute(String)`):
```
picocli.CommandLine$ParameterException: Error: 'chem' is not unique: it matches 'chemical-files', 'chem-formats'
	at picocli.CommandLine$AbbreviationMatcher.match(CommandLine.java:18426)
	at picocli.CommandLine$Interpreter.processArguments(CommandLine.java:13306)
	at picocli.CommandLine$Interpreter.parse(CommandLine.java:13122)
	at picocli.CommandLine$Interpreter.processSubcommand(CommandLine.java:13395)
	at picocli.CommandLine$Interpreter.processArguments(CommandLine.java:13310)
```
I guess that's because the code builds up a Map<String,Object> of each name/alias and the AbbreviationMatcher doesn't check that they reference the same objects?

NewbieOrange:
Fixes #1537.


remkop:
I was finally able to review this.

Issue #1540, opened 2022-01-03
[Documentation] Looking for more examples


binkley:
I'm looking for more examples of mixing Picocli with JLine for a rich-featured command line tool.
My use case is simple: rolling dice expressions (ref: https://github.com/binkley/kotlin-dice)

My goals are multiple modes of operation:

- Support-type flags, such as `--help` should have colorized output
- Testing -- test `main()` and the modes when possible without hosing streams or hanging
- REPL -- full colorizing and keyboard controls (_eg_, `Ctrl-a`, _et al_)
- Rolls on the command line -- no colorizing, just print to STDOUT, suitable in scripting
- Rolls from STDIN (ie, pipe, etc) -- no colorizing, just print to STDOUT, suitable in scripting

In these, distinguish STDOUT from STDERR.

remkop:
You may be interested in the [Micronaut Starter CLI](https://github.com/micronaut-projects/micronaut-starter/tree/3.2.x/starter-cli/src/main/java/io/micronaut/starter/cli); that uses JLine and picocli.

binkley:
Thanks for that.  I like Micronaut.  Might the example be pulled into the docs as a reference?
If the link is there, apologies for missing it.

remkop:
Another example is the quarkus cli: https://github.com/quarkusio/quarkus/tree/main/devtools/cli/src/main/java/io/quarkus/cli

It may be good to link to both frameworks or even more if possible.

binkley:
Cool.  I am inclined for a docs PR, but it will take me time.
My example usages might not cover enough.  My goal of a UNIX-friendly command line tool might not match other use cases.

remkop:
The source code of the [micronaut-starter CLI](https://github.com/micronaut-projects/micronaut-starter/tree/3.3.x/starter-cli/src/main/java/io/micronaut/starter/cli) may be of interest: it shows how JLine2 and picocli can be combined to provide a complex, full-featured interactive command line REPL application.

PR #1542, opened 2022-01-03
Review, please -- "simple" example for command-line program

binkley:
For issue #1540

This PR is more for review & guidance than for merging (though it is mergeable best I can tell).
I'm looking to see if I'm heading in the right direction before I go too far along.

Things I'd like to understand/address include:
- How should I test this code?
- Is it OK to create an executable uberjar?  (I'm using this for manual command-line testing)
- Is there a better way to detect "isatty"?
- Did I pick the right directory and/or package for this work?

### DIFF
--- a/picocli-shell-jline3/src/main/java/picocli/tool/jline3/ExampleTool.java
+++ b/picocli-shell-jline3/src/main/java/picocli/tool/jline3/ExampleTool.java
@@ -11,7 +11,9 @@ import picocli.CommandLine.Command;
 import picocli.CommandLine.Option;
 import picocli.CommandLine.Parameters;
 
+import java.io.BufferedReader;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.util.List;
 import java.util.concurrent.Callable;
 
@@ -58,19 +60,38 @@ final class Options implements Callable<Integer> {
 
     @Override
     public Integer call() {
+        // Check if we are reading from a pipe or shell redirection
+        final boolean isatty = null != System.console();
+
         // Return failure to demonstrate that `-h` does not fail: standard
         // options are handled by Picocli: this `call()` method is not
         // invoked by standard options
-
-        if (arguments.isEmpty())
+        if (!isatty) {
+            // TODO: Bleh.  Nicer in more modern Java
+            try (final BufferedReader lineReader = lineReader()) {
+                while (true) {
+                    final String line = lineReader.readLine();
+                    if (null == line) break;
+                    final Integer result = process(line);
+                    if (0 != result)
+                        return result;
+                }
+            } catch (final IOException e) {
+                return 2;
+            }
+        } else if (arguments.isEmpty())
             return runRepl(prompt);
         else for (final String argument : arguments) {
-            final Integer result = process(argument);
-            if (0 != result)
-                return result;
-        }
+                final Integer result = process(argument);
+                if (0 != result)
+                    return result;
+            }
 
         return 0;
+    }
+
+    private static BufferedReader lineReader() {
+        return new BufferedReader(new InputStreamReader(System.in));
     }
 }
 
@@ -123,12 +144,14 @@ final class Repl {
 /**
  * Your processing logic goes here. It could be human-edited text from the
  * REPL, it could be tests, it could be text passed from the command line, it
- * could be text read from <code>STDIN</code> via a pipe or shell redirection.
+ * could be text read from <code>STDIN</code> via a pipe or shell
+ * redirection.
  */
 final class YourProcessing {
     // TODO: It is challenging in Java to return _all of_ a status code for
     //  the command line, output results, and to let caller handle exceptions
     static Integer process(final String input) {
+        System.out.println(input); // Dummy example
         return 0;
     }
 }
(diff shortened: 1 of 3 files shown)


binkley:
The initial commits are all code.

Before I write up separate documentation, I'd like working code to back up those docs.

remkop:
Hi, sorry for the late response. I have been traveling with family. It may take me some time to get to this.

picocli-shell-jline3/src/main/java/picocli/tool/jline3/ExampleTool.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+        // Use case #2 -- pipe or shell redirection
+        else if (null == System.console()) return runFromInput();
+        // Use case #3 -- interactive REPL
+        else return runRepl(prompt);
+    }
+
+    /** @todo Nicer way using more modern Java. */
+    private static Integer runFromInput() {
+        try (final BufferedReader lineReader = lineReader()) {
```
remkop, line 73: This method seems to be a rudimentary re-implementation of JLine functionality.
I don't think we should suggest to users that they do this.
My recommendation to users would be to familiarize themselves with what JLine offers and use that instead.

picocli-shell-jline3/src/main/java/picocli/tool/jline3/ExampleTool.java
```diff
@@ -0,0 +1,146 @@
...
+
+final class Repl {
+    /**
+     * This is an <em>absolutely minimal</em> REPL.
+     *
+     * @param prompt the prompt to show on left-side asking user for input
+     *
+     * @return the exit code for <code>main()</code>
+     *
+     * @todo Newer versions of Java permit defining this as a lambda
```
remkop, line 99: Do we need this TODO?

picocli-shell-jline3/src/main/java/picocli/tool/jline3/ExampleTool.java
```diff
@@ -0,0 +1,146 @@
...
+     */
+    static Integer runRepl(final String prompt) {
+        try (final Terminal terminal = TerminalBuilder.builder().name("example tool").build()) {
+            final LineReader lineReader = LineReaderBuilder.builder().terminal(terminal).build();
+            do {
+                final String input;
+                try {
+                    input = lineReader.readLine(prompt);
+                } catch (final UserInterruptException | EndOfFileException e) {
+                    // Handling user banging on <kbd>Ctrl-C</kbd>
```
remkop, line 109: Do we want to catch these exceptions here?
My preference would be to let this method throw these exceptions and handle them in the place where this method is called.

picocli-shell-jline3/src/main/java/picocli/tool/jline3/ExampleTool.java
```diff
@@ -0,0 +1,146 @@
...
+                // TODO: It is challenging in Java to return _both_ a status
+                //  code for the command line, and to let caller handle
+                //  exceptions
+                final Integer result = process(input);
+                // Another policy decision: Should invalid input kill the
+                // REPL, or notify the user and continue?
+                if (0 != result) return result;
+            } while (true);
+        } catch (final IOException e) {
+            // TODO: It is challenging in Java to return _both_ a status code
```
remkop, line 126: If processing the exception is challenging them perhaps we should not catch it here, or perhaps we should wrap it in a custom exception and re-throw it .

picocli-shell-jline3/src/main/java/picocli/tool/jline3/ExampleTool.java
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Your processing logic goes here. It could be human-edited text from the
+ * REPL, it could be tests, it could be text passed from the command line, it
+ * could be text read from <code>STDIN</code> via a pipe or shell
+ * redirection.
+ */
+final class YourProcessing {
+    // TODO: It is challenging in Java to return _all of_ a status code for
+    //  the command line, output results, and to let caller handle exceptions
```
remkop, line 141: Can we remove this TODO?

remkop (changes requested):
Overall, this approach is not one I would take.

This approach uses picocli as an entry point, from which a JLine (or JLine-like) REPL is started, after which the `YourProcessing` class somehow needs to deal with the user input.

One problem of this approach is that it is not clear is how the application would tokenize the user input: given an input string of `mycommand --option --option2=value a b x,y,z` we want to break this up into `mycommand`, `--option`, `--option2=value`, `a`, `b`, and `x,y,z`, so we can invoke `CommandLine.execute(args)` with these tokens. JLine3 provides this functionality, and much more, out of the box. IMO it pays to spend the time to become familiar with the functionality that JLine3 offers rather than re-implementing this functionality ourselves.

The example in the https://github.com/remkop/picocli/blob/master/picocli-shell-jline3/README.md uses the JLine3 API extensively, and is more in line with what I would recommend to application authors. 

The Example in the README shows how to wire together picocli commands with JLine3.
Once that wiring is in place, your REPL application can support any number of commands, that can simply use the picocli API (`Runnable` or `Callable` classes with `@Option` or `@Parameters`-annotated fields) to define these commands.

There is a learning curve to this, but on the other hand it allows for application code to work at a much higher level of abstraction.

Issue #1526, opened 2021-12-20
In ZSH, use of compinit clobbers existing completions


bherw:
I'm generating completions for a bunch of commands in a row and trying to load all their completions sequentially like so:

```shell
if [ -d "$SCRIPTDIR/../bash_completion.d" ]; then
	for bcp in "$SCRIPTDIR/../bash_completion.d"/*.sh; do
		source "$bcp"
	done
fi
```

This works great in bash, but unfortunately in zsh the "compinit" call at the top of each script resets all the rest of the completions, resulting in only the last command alphabetically in the directory getting any completions.

I've worked around this for now by adding an `alias compinit=` before the loop.

Based on https://zsh.sourceforge.io/Doc/Release/Completion-System.html, compinit defines compdef, so I think testing if it's already loaded as follows should be enough to avoid clobbering completions:
```shell
if ! type compdef > /dev/null; then
    compinit
fi
```

remkop:
Thank you for raising this!
Can you confirm that this works for both bash and zsh?

bherw:
I haven't tried editing the autocomplete scripts directly yet, but bash seems fine with it if I put it in a `if [ -n "$ZSH_VERSION" ]` block. The issue only affects the use of ZSH's compinit in that block; bash worked fine already, so as long as it's not causing a syntax error it should be fine.

I'll see if I can find some time to make a PR; should be a quick change.

remkop:
> I'll see if I can find some time to make a PR; should be a quick change.

That would be great, thank you! Most of my time is occupied with Log4j recently...

remkop:
@bherw Have you had a chance to look at this?

PR #1548, opened 2022-01-17
[#1526] Fix use of compinit clobbering existing completions

bherw:
Created a couple of test programs and tested the autocomplete scripts in zsh and bash.


remkop:

PR #1616, opened 2022-02-27
1615: Fix bug in getCJKAdjustedLength() that double counts supplement…

### LINKED ISSUE #1615: getCJKAdjustedLength is miscalculating for supplementary code points
getCJKAdjustedLength() calculates the column width of text and adjusts for CJK values that may be double widths.

But this method is miscalculating for supplementary code points line "𝑓". In this case the method is returning 2 rather than the expected 1.

gwalbran:
…ary code points

getCJKAdjustedLength calculates the column width of text strings.

But it operates on chars which causes it to double count supplementary code
points. Supplimentary code points have a high and a low code unit(char).

This commit reworks this logic to be based on code points. The code assembles
the code points by hand as String.codePoints() was only added in 1.8 and
picocli has a source level of 1.5

closes #1615

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -7950,16 +7950,16 @@ public class CommandLine {
              * @since 4.0 */
             public UsageMessageSpec autoWidth(boolean detectTerminalSize) { autoWidth = detectTerminalSize; return this; }
             /**
-             * Given a character, is this character considered to be a CJK character?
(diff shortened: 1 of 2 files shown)


remkop (approved):
This looks great, thank you very much!
I am working on some improvements to the build (currently CI is broken), and I will merge this PR after I get the build fixed.

remkop:
Merged. Thank you again for the contribution!

PR #1627, opened 2022-03-07
User guide, chapter 3.2.3. Forcing Interactive Input: code sample: add Kotlin version

deining:
While running the sample code, I realized that there is a minor glitch here: if there is no description given for the interactive option, there are opening and closing braces which make no sense without the description string:

```
Enter value for --interactive (): 
```

Would be nice if these braces were supressed if no description string is given.

### DIFF
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -689,6 +689,25 @@ public class Main implements Runnable {
 }
 ----
 
+.Kotlin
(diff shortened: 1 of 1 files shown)


remkop (approved):

remkop:

I raised #1629 for the issue with the standard prompt for interactive options. Thanks for raising this!

PR #1677, opened 2022-05-12
Doc improvements

Vampire:
A couple of improvements and mainly fixes to the manual.

I didn't find whether you replace versions manually or somehow automated.
In the last paragraph there is a link to `CommandLine.java` that still pointed to `4.6.3-SNAPSHOT` while the remaining document points to `4.7.0-SNAPSHOT`, so however you do the replacement, that spot was missed.

### DIFF
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -1070,7 +1070,7 @@ command line with lots of options or with long arguments for options.
 
 Starting from v2.1.0, picocli supports "argument files" or "@-files".
 Argument files are files that themselves contain arguments to the command.
-When picocli encounters an argument beginning with the character `@',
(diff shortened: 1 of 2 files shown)


remkop (approved):

remkop:
Looks great, thank you!

That 4.6.3-SNAPSHOT version in the doc should have been automatically replaced but wasn't, so thank you for pointing that out.  👍 
I need to fix that in the build.

remkop:
Turns out I had a typo in the regex for the link to a stable version of the `CommandLine.java` source code. 
Fixed now.  ✔️ 
Thank you again for the contributions! 🙏

Issue #1706, opened 2022-06-10
Option descriptions not read from parent command resource bundle


rsenden:
Suppose we have the following command hierarchy:

```
  RootCommand (resourceBundle=RootMessages)
  - ModuleCommand (resourceBundle=ModuleMessages)
     - Cmd1
     - Cmd2
```

Behaviour:
- As expected: If we define `usage.footer` in the RootMessages bundle, this footer will be displayed on all commands.
- As expected: If we define `option1` in the ModuleMessages bundle, any `option1` in any of the leaf commands displays the proper description.
- Not as expected: If we define `option2` in the RootMessages bundle, any `option2` in any of the leaf commands displays an empty description

Since our modules are maintained by different teams, we want to have a dedicated resource bundle for each module. However, as we also have some common options that are used throughout the application (mostly through generic mixins), we would like to define the descriptions for those generic options in the RootMessages bundle.

Unfortunately that doesn't seem to work, so for now we have to replicate those command option descriptions across every ModuleMessages bundle in each module. Obviously this is something that we would like to avoid.

Is there any way to make this work with the current picocli version, or will this require a fix in picocli?

remkop:
> * As expected: If we define `usage.footer` in the RootMessages bundle, this footer will be displayed on all commands.
> * As expected: If we define `option1` in the ModuleMessages bundle, any `option1` in any of the leaf commands displays the proper description.
> * Not as expected: If we define `option2` in the RootMessages bundle, any `option2` in any of the leaf commands displays an empty description

To clarify: you are saying that the (non-empty) description for `option2` in the RootMessages bundle gets lost because of the intermediate ModuleMessages bundle?

And, the behaviour for options is different from the behaviour for things like the footer?
So the scenario is that ModuleMessages bundle does not contain a `usage.footer` key, only the top-level RootMessages bundle has this key, but still, while Cmd1 and Cmd2 can access the `usage.footer` keys from the top-level RootMessages bundle, they cannot access the `option2` in the RootMessages bundle?

That seems a bit strange, to be honest... Can you verify that this is really what is happening?

Generally speaking, the easiest way with current picocli would be to have a single ResourceBundle that is set at the top-level command, that has all the keys for all commands. This can be done by prefixing the option names with the command name.

For example:

```
# single resource bundle set on RootCommand, other commands do not have a resource bundle
RootCommand.ModuleCommand.Cmd1.option1 = description for this option1
[...]

PR #1709, opened 2022-06-15
Updated unit tests

wtfacoconut:
Hey @rsenden, 
Here's a summary of changes for the i18n unit test case for your update to resolve issue #1706 .

## The primary change what was made:
- Added new class with option to act ac a Mixin option between parent command and subcommand
- Added description for the new option in `SharedMessages` resource bundle
- Created new subcommand class called `sub2` and added the previously mentioned option as a Mixin. Also, `sub2` has been to use a different resource bundle than the parent command.
- Added the new `sub2` command and new Mixin option to the parent command

## Other changes that were made so that other existing i18n tests don't break:
- Updated the japanese resource (`SharedMessages_ja`) with description for the new mixin option
- Added the method `stripFactoryInstanceHashcodes` to TestUtil to strip to remove hash codes (object instance IDs?) from trace output so that `testTracingWithResourceBundle` will not fail due to my updates to the test command.
- Updated the `expect` string variables for many of the other test cases.

## One area of concern:
During testing, if there's a parent command that uses a resource bundle where the "usage.header" key is defined, but a subcommand's resource bundle doesn't (and also doesn't have a usage header or description defined in the subcommand's @Command annotation) then the subcommand will adopt the "usage.header" key from the parent resource bundle to be used as the subcommand's description when the parent command's help output is displayed. Please see image below:
image

Please let me know if this is expected behavior or if we should instead be expecting to see a blank/emty description of that subcommand from the parent command's help message output. If this is not intended, then for this scenario, I'll update the existing tests to fail if they see anythin for the subcommand's help message output.


wtfacoconut:
Sorry, pull request sent to the wrong repo.

remkop:
For what it’s worth, the behaviour described in the area of concern sounds reasonable to me.

wtfacoconut:
> For what it’s worth, the behaviour described in the area of concern sounds reasonable to me.

Hey @remkop. Do you think I should raise this as a new issue?

remkop:
@wtfacoconut Sorry I was unclear; I meant that the behaviour you describe sounds fine:
if a subcommand does not have an entry in its resource bundle then this subcommand showing the entry from the parent command resource bundle sounds exactly right. (So I don't think we need a new ticket, unless I am missing something.)

If you can work with @rsenden on this that would be great.
I won't be able to review any pull requests until after this weekend.
Many thanks for contributing to this!

wtfacoconut:
No worries and thank you for the clarification. Yes, I'll tag team with rsenden. 👍

Issue #1708, opened 2022-06-13
The setUsageHelpLongOptionsMaxWidth method launchs exception when running on terminal with small width


fabio-franco:
Hi, I saw that the default behavior of the `setUsageHelpLongOptionsMaxWidth` method is to trigger an exception if the width of the terminal is small. However, I think that makes sense to set the width of options when possible (mid/big screen) but just skip it when it is not possible, instead of failure with this exception:

```
Exception in thread "main" picocli.CommandLine$InitializationException: Invalid usage long options max width 50. Value must not exceed width(55) - 20
        at picocli.CommandLine$Model$UsageMessageSpec.longOptionsMaxWidth(CommandLine.java:7687)
        at picocli.CommandLine.setUsageHelpLongOptionsMaxWidth(CommandLine.java:3376)
        at com.MyApp.main(MyApp.java:110)
```

For example, letting the behavior with an exception makes me do a workaround that works because I am using `setUsageHelpAutoWidth(true)`. I think that other people will potentially do a similar thing to avoid the failure (or try-catch it).

```
CommandLine cmd = new CommandLine(new MyApp())
			.setUnmatchedArgumentsAllowed(false)
			.setSubcommandsCaseInsensitive(true)
			.setOptionsCaseInsensitive(true)
			.setUsageHelpAutoWidth(true);
		
		int width = cmd.getUsageHelpWidth();
		int defaultLongOptionsWidth = 50;
		
		// This is a workaround to avoid setUsageHelpLongOptionsMaxWidth, which needs the defaultLongOptionsWidth 
		// to be bigger than terminal screen plus 20, otherwise it will failure with picocli.CommandLine$InitializationException
		if(width - defaultLongOptionsWidth > 20) {
			cmd.setUsageHelpLongOptionsMaxWidth(defaultLongOptionsWidth);
		}
```

Is that possible to change the behavior to avoid the exception by just ignoring the value passed to `setUsageHelpLongOptionsMaxWidth` when the terminal width is too small?

remkop:
Yes, that makes sense.
Will you be able to provide a pull request for this?

fabio-franco:
Sure, but probably I will do that on the weekend.

remkop:
I miss-clicked and accidentally merged before reviewing. 😅
I won’t be able to review it for a few more days; I hope you’re okay with the delayed feedback. (But I’m sure it’ll be solid stuff anyway.) 😄
Thanks again 🙏

PR #1712, opened 2022-06-19
[Issue 1708] Ignore invalid values in setUsageHelpLongOptionsMaxWidth

fabio-franco:
This is a fix proposal for https://github.com/remkop/picocli/issues/1708. 

The idea is to avoid exception when the screen is too small and just let setUsageHelpLongOptionsMaxWidth set the newValue when valid.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -7820,17 +7820,15 @@ public class CommandLine {
             /**
              * Sets the maximum usage help long options column max width to the specified value.
              * This value controls the maximum width of the long options column: any positional parameter labels or long options that are longer than the specified value will overflow into the description column, and cause the description to be displayed on the next line.
-             * @param newValue the new maximum usage help long options column max width. Must be 20 or greater.
+             * @param newValue the new maximum usage help long options column max width. Must be 20 or greater, otherwise the new value will be ignored.
              * @return this {@code UsageMessageSpec} for method chaining
-             * @throws IllegalArgumentException if the specified long options column max is less than 20
              * @since 4.2 */
             public UsageMessageSpec longOptionsMaxWidth(int newValue) {
-                if (newValue < DEFAULT_USAGE_LONG_OPTIONS_WIDTH) {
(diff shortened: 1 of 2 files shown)


remkop:

src/main/java/picocli/CommandLine.java
```diff
@@ -7820,17 +7820,15 @@ public UsageMessageSpec width(int newValue) {
...
              * @return this {@code UsageMessageSpec} for method chaining
-             * @throws IllegalArgumentException if the specified long options column max is less than 20
              * @since 4.2 */
             public UsageMessageSpec longOptionsMaxWidth(int newValue) {
-                if (newValue < DEFAULT_USAGE_LONG_OPTIONS_WIDTH) {
-                    throw new InitializationException("Invalid usage long options max width " + newValue + ". Minimum value is " + DEFAULT_USAGE_LONG_OPTIONS_WIDTH);
-                } else if (newValue > width() - DEFAULT_USAGE_LONG_OPTIONS_WIDTH) {
-                    throw new InitializationException("Invalid usage long options max width " + newValue + ". Value must not exceed width(" + width() + ") - " + DEFAULT_USAGE_LONG_OPTIONS_WIDTH);
+                if (newValue >= DEFAULT_USAGE_LONG_OPTIONS_WIDTH && newValue <= width() - DEFAULT_USAGE_LONG_OPTIONS_WIDTH) {
+                    longOptionsMaxWidth = newValue;
```
remkop, line 7828: I agree with not throwing an exception, but I would like to log an `INFO`-level trace message that tells the user that we are ignoring the specified value, and why (similar to the message text of the InitializationException).

remkop (commented):
I am okay with not throwing the exception but I do think we to log a message when ignoring the user-specified value. Can you take a look?

src/main/java/picocli/CommandLine.java
fabio-franco: Sure, it totally makes sense. I will update the branch and send you an update soon.

PR #1723, opened 2022-06-26
[Issue: #1708] Including log info for invalid values scenario

fabio-franco:
In this pr it was included some log info for invalid scenarios from this issue: https://github.com/remkop/picocli/issues/1708

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -7826,8 +7826,9 @@ public class CommandLine {
             public UsageMessageSpec longOptionsMaxWidth(int newValue) {
                 if (newValue >= DEFAULT_USAGE_LONG_OPTIONS_WIDTH && newValue <= width() - DEFAULT_USAGE_LONG_OPTIONS_WIDTH) {
                     longOptionsMaxWidth = newValue;
+                } else {
+                    CommandLine.tracer().info("Invalid usage long options max width %d. Value must not exceed width(%d) - %d", newValue, width(), DEFAULT_USAGE_LONG_OPTIONS_WIDTH);
                 }
-
                 return this;
             }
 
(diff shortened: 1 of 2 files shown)


src/main/java/picocli/CommandLine.java
```diff
@@ -7826,8 +7826,9 @@ public UsageMessageSpec width(int newValue) {
             public UsageMessageSpec longOptionsMaxWidth(int newValue) {
                 if (newValue >= DEFAULT_USAGE_LONG_OPTIONS_WIDTH && newValue <= width() - DEFAULT_USAGE_LONG_OPTIONS_WIDTH) {
                     longOptionsMaxWidth = newValue;
+                } else {
+                    CommandLine.tracer().info("Invalid usage long options max width %d. Value must not exceed width(%d) - %d", newValue, width(), DEFAULT_USAGE_LONG_OPTIONS_WIDTH);
```
remkop, line 7830: I think it would be good to have separate log messages for when the longOptionsMaxWidth is too small and for when it is too big.
That will allow us to give much clearer feedback to end users.
Something like this?

```java
if (newValue < DEFAULT_USAGE_LONG_OPTIONS_WIDTH) {
    CommandLine.tracer().info("Invalid usage long options max width %d. Minimum value is %d", newValue,  DEFAULT_USAGE_LONG_OPTIONS_WIDTH);
} else if (newValue > width() - DEFAULT_USAGE_LONG_OPTIONS_WIDTH) {
    CommandLine.tracer().info("Invalid usage long options max width %d. Value must not exceed width(%d) - %d", newValue , width(), DEFAULT_USAGE_LONG_OPTIONS_WIDTH);
} else {
    longOptionsMaxWidth = newValue;
}
```

remkop (changes requested):
I think it would be good to have separate log messages for when the longOptionsMaxWidth is too small and for when it is too big.
That will allow us to give much clearer feedback to end users.
Can you take a look?

remkop (approved):

remkop:
Merged. Thank you for the contribution! 🙏

Issue #1644, opened 2022-03-31
Picocli 4.6.3 : autocompletion of directory names not working anymore


philgdn:
I'm facing a regression with Picocli 4.6.3 : the autocompletion of filenames doesn't work anymore.

The help of my `OdsManager` app:

`OdsManager extractLoadOds [-hV] [-p PROPERTIES] [JOBFILE [--maxErrorsByExtraction MAX_ERRORS] [--[no-]cleanBeforeExtract] [-f | -s SINCE]] [[--commitAfter <commitLoadAfter>] `

When I enter `OdsManager extractLoadOds ` on the command line, with Picocli 4.6.2 hitting the TAB twice I was getting the possibility to autocomplete **directory names** for the JOBFILE parameter:

- if  I hit TAB twice I get the list of files and directories in the current directory 
- if I then type `jo` TAB TAB (for the existing `jobs/` directory)... I get no suggestions

This was working in 4.6.2 but doesn't work anymore in 4.6.3.

remkop:
Thank you for raising this!
I will take a look.

remkop:
I finally had some time to look at this.
This seems to be caused by this change: https://github.com/remkop/picocli/pull/1473: if I comment out the `local IFS=$'\n'` in the completion script section for the positional parameter, completion seems to work for partial directory names.

I need to spend more time debugging this.

NewbieOrange:
Is this reproducible with https://raw.githubusercontent.com/remkop/picocli/master/src/test/resources/picocompletion-demo_completion.bash?

philgdn:
Sorry I missed your message of September 1.

With the provided `picocompletion-demo_completion.bash` script I get the completion for the directory names

NewbieOrange:
Can you provide a minimal example to reproduce the issue?

philgdn:
Here are examples (_not so minimal sorry_) : [Examples.zip](https://github.com/remkop/picocli/files/9839753/Examples.zip)

- `SolifeOdsBatchManager_completion_462` was  generated by Picocli 4.6.2 and works as expected
- `SolifeOdsBatchManager_completion_463` was  generated by Picocli 4.6.3 : autocompletion don't work for _directories_

To reproduce the problem (with SolifeOdsBatchManager_completion_463) in a directory containing files and directories type (for example a directory `dir1` :

`SolifeOdsBatchManager extractLoadOds d` then TAB.

The autocompletion should complete the `dir1` directory... but it doesn't

[...]

PR #1759, opened 2022-07-29
Honor spaces in autocomplete candidates

### LINKED ISSUE #1754: Autocomplete mishandling suggestions with spaces
When an option provides a `completionProvider` that list strings with spaces, the autocomplete doesn't honor the completion provider, but instead splits all options on its own.

For instance, given the example:

```java
    @Command(mixinStandardHelpOptions = true)
    public static class DummyCommand {
        @Option(names = "--candidate", completionCandidates = CandidatesCompletion.class)
        public String candidate;

        public static class CandidatesCompletion implements Iterable<String> {

            @Override
            public Iterator<String> iterator() {
                return Arrays.asList("John Doe", "Michael Scott", "John Sena").iterator();
            }
        }
    }
```

I'd expect 3 candidates to be offered, with proper escaping for the whitespace. However, this is what it's actually being suggested:

```
$ myapp --candidate <TAB><TAB>
Doe      John     Michael  Scott    Sena
```

5 options, all single words, duplicates removed.

jsotuyod:
Fixes #1754
 
- Both positional params and option candidates are now stored in the autocomplete script as an array.
- By setting `IFS=$'\n'` for those scenarios, each array entry is treated as a single candidate,
allowing to honor any spaces in the original suggestions coded by the user.
- Values still need to be double-escaped for proper handling.


remkop:
Thank you for the PR! Checking...

jsotuyod:
@remkop any feedback on this?

remkop:
Oops! Sorry for the long wait. I have not been able to spend much time on picocli recently...
Also, this ticket seems related to https://github.com/remkop/picocli/issues/1644 , so I want to look at both together.
I have not forgotten about it though!

NewbieOrange (approved):
LGTM

remkop (approved):

remkop:
Merged.

remkop:
Started the release process now...

src/main/java/picocli/AutoComplete.java
```diff
@@ -750,7 +785,7 @@ private static String generatePositionalParamsCases(List<PositionalParamSpec> po
             int max = param.index().max();
             if (param.completionCandidates() != null) {
                 buff.append(format("%s    %s (( currIndex >= %d && currIndex <= %d )); then\n", indent, ifOrElif, min, max));
-                buff.append(format("%s      positionals=$( compgen -W \"$%s_pos_param_args\" -- \"%s\" )\n", indent, paramName, currWord));
+                buff.append(format("%s      positionals=$( compReplyArray \"${%s_pos_param_args[@]}\" )\n", indent, paramName, currWord));
```
remkop, line 788: We recently [discovered](https://github.com/remkop/picocli/issues/2053) that the format string has only 2 `%s` variables, but we specify 3 parameters (`indent`, `paramName`, `currWord`)...

This means that the last parameter (`currWord`) is not used.
Is that correct? Can the `currWord` parameter be removed?

src/main/java/picocli/AutoComplete.java
```diff
@@ -793,7 +828,8 @@ private static String generateOptionsCases(List<OptionSpec> argOptionFields, Str
             }
             if (option.completionCandidates() != null) {
                 buff.append(format("%s    %s)\n", indent, concat("|", option.names()))); // "    -u|--timeUnit)\n"
-                buff.append(format("%s      COMPREPLY=( $( compgen -W \"${%s_option_args}\" -- \"%s\" ) )\n", indent, bashify(option.paramLabel()), currWord));
+                buff.append(format("%s      local IFS=$'\\n'\n", indent));
+                buff.append(format("%s      COMPREPLY=( $( compReplyArray \"${%s_option_args[@]}\" ) )\n", indent, bashify(option.paramLabel()), currWord));
```
remkop, line 832: We recently [discovered](https://github.com/remkop/picocli/issues/2053) that the format string has only 2 `%s` variables, but we specify 3 parameters (`indent`, `bashify(option.paramLabel())`, `currWord`)...

This means that the last parameter (`currWord`) is not used.
Is that correct? Can the `currWord` parameter be removed?

remkop (commented):
Hi @jsotuyod! Thank you again for working on this PR! 
I realize it is a while ago that you worked on this, but a question recently came up:
@vorburger [ran ErrorProne against the picocli project](https://github.com/remkop/picocli/issues/2053), which flagged some items on AutoComplete.java:788 and 832:

> ... the format string has only 2 `%s` variables, but the code specifies 3 parameters (`indent`, `paramName`, `currWord`)...

This means that the last parameter (`currWord`) is not used. Is that correct? Can the `currWord` parameter be removed?

src/main/java/picocli/AutoComplete.java
jsotuyod: Yeah, it can be removed. Sorry for missing that

src/main/java/picocli/AutoComplete.java
vorburger: fixed in #2175

src/main/java/picocli/AutoComplete.java
vorburger: fixed in #2175

PR #1764, opened 2022-08-04
Add documentation about resource bundle entries for negatable options

vanitasvitae:
Hey,
not sure if I added this to the correct file, but I thought this line was missing from the documentation.

### DIFF
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -9924,6 +9924,11 @@ verbose = Show more detail during execution. \
           May be specified multiple times for increasing verbosity.
 ----
 
+For negatable options, the entry in the resource bundle must match the negated form:
+----
+no-verbose = Show verbose output
+----
+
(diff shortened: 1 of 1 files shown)


remkop (approved):

remkop:
@vanitasvitae Thank you for the pull request!

I was actually not aware that this is the case (that the negative form of negatable options is the key in resource bundles). 

Can you add a unit test to demonstrate?

remkop (changes requested):
Can you please add a unit test that demonstrates this behavior?

vanitasvitae:
After opening this PR, I noticed that this is in fact only true for negatable options which are supposed to be true by default (see https://picocli.info/#_negatable_options). I think this PR needs rewording :D

vanitasvitae:
In the end I guess the behavior for negatable options is no different than for normal options (resource entry must equal `names` value without dashes).
I guess there is no need to mention this explicitly in the documentation.

remkop:
Okay great, thank you for the confirmation! 🙏

PR #1778, opened 2022-08-18
Migrate to new Spring Boot auto configuration introduced in spring bo…

aspan:
…ot 2.7. The existing spring.factories support will be removed in spring boot 3 so this will make sure that the autoconfiguration will still work in when upgrading to spring boot 3.

### DIFF
--- a/picocli-spring-boot-starter/src/main/java/picocli/spring/boot/autoconfigure/PicocliAutoConfiguration.java
+++ b/picocli-spring-boot-starter/src/main/java/picocli/spring/boot/autoconfigure/PicocliAutoConfiguration.java
@@ -1,10 +1,10 @@
 package picocli.spring.boot.autoconfigure;
 
-import org.springframework.boot.autoconfigure.AutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Primary;
 import picocli.CommandLine;
 import picocli.CommandLine.IFactory;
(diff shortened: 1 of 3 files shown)


remkop:
Hi @aspan, you mentioned that this new mechanism was introduced in Spring Boot 2.7, which was [released](https://spring.io/blog/2022/05/19/spring-boot-2-7-0-available-now) only 3 months ago...

Will this new mechanism work with older versions of Spring (2.6.x and earlier)?

Is there any way to be compatible with both Spring 3.x and Spring 2.6.x and earlier?

aspan:
I don't think there is a way to be backwards compatible with spring boot 2.6. The support has not been removed yet in spring boot 3 so it still works there. See. [#29699](https://github.com/spring-projects/spring-boot/issues/29699) and [#29698](https://github.com/spring-projects/spring-boot/issues/29698). So I think that there needs to be a separate module for spring 3 if your aim is to be backwards compatible with spring boot 2.6.

remkop:
I prefer not to have separate modules for Spring 2.6 and 2.7/3.x...

Would it be possible to support both mechanisms?

So, we don't remove [picocli-spring-boot-starter/src/main/resources/META-INF/spring.factories](https://github.com/remkop/picocli/pull/1778/files#diff-940779d1f39f673b46d486b91977ca9816c704d308d7151bdcdd47d11a5cf656) --> this will allow it to work with Spring 2.6 and earlier

And we add [picocli-spring-boot-starter/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports](https://github.com/remkop/picocli/pull/1778/files#diff-708b9796b3c36d3ab1ac4efd6f27c612d8281e778494219563233fcc33f9aac3) --> this will allow it to work with Spring 2.7 and later

Can I ask you to test whether that approach would allow us to support both Spring 2.6 and 2.7/3.x?

snicoll:
We've had a similar request (registering the auto-configuration in both `META-INF/spring.factories` and the new location so that it works in both 2.x and 3.x. Picocli might be a good candidate as it relies on a small part of the Spring Boot API. Let me take that back to the team and I'll report back. Thanks!

remkop:
Awesome, thank you @snicoll! 🙏😊

remkop:
@aspan Will you be able to make these changes?

aspan:
I'll make the changes today.

remkop (approved):

remkop:
Merged. Thank you for the pull request!

PR #1796, opened 2022-09-01
Fixing broken links and typos

deining:
This PR
* fixes broken links (docu, javadocs)
* fixes several typos
* updates asciidoctorj to latest version 2.5.5.


remkop (approved):

remkop:
Thank you @deining !! 🙏 
Much appreciated!

(It is unfortunate that the GraalVM documentation now seems tightly coupled to a version number... :-(
Actually, looking closer now, it seems that generic URLs without version number are automatically redirected to the latest version...

If I open `https://www.graalvm.org/reference-manual/native-image/basics/` in my browser,
it is redirected to 
`https://www.graalvm.org/22.2/reference-manual/native-image/basics/`

Perhaps it is better to remove the version number from the URLs in the picocli docs and rely on the GraalVM site to redirect to the latest version. What do you think?

deining:
> If I open `https://www.graalvm.org/reference-manual/native-image/basics/` in my browser, it is redirected to `https://www.graalvm.org/22.2/reference-manual/native-image/basics/`
> 
> Perhaps it is better to remove the version number from the URLs in the picocli docs and rely on the GraalVM site to redirect to the latest version. What do you think?

Good idea, done with #1881.

PR #1802, opened 2022-09-07
Fix #1767 - NPE on OptionSpec.getValue() (IScoped)

### LINKED ISSUE #1767: NPE on OptionSpec.getValue()

rsenden:
This PR fixes https://github.com/remkop/picocli/issues/1767, based on the IScoped approach discussed in the issue

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -5995,6 +5995,15 @@ public class CommandLine {
          * @since 4.0
          */
         public interface IScope extends IGetter, ISetter {}
+        
+        /** This interface provides access to an {@link IScope} instance. 
+         */
+        public interface IScoped {
+            /** Get the {@link IScope} instance.
+             * 
+             *  @return {@link IScope} instance */
+            IScope getScope();
+        }
 
         /** Customizable getter for obtaining the current value of an option or positional parameter.
          * When an option or positional parameter is matched on the command line, its getter or setter is invoked to capture the value.
@@ -9166,9 +9175,25 @@ public class CommandLine {
              * @return whether this argument applies to all descendent subcommands of the command where it is defined
              * @since 4.3 */
             public ScopeType scopeType() { return scopeType; }
+            
+            /** Check whether the {@link #getValue()} method is able to get an actual value from the current {@link #getter()}. */
+            public boolean isValueGettable() {
+                if (getter instanceof IScoped) {
+                    IScoped scoped = (IScoped) getter;
+                    IScope scope = scoped.getScope();
+                    try {
+                        Object obj = scope.get();
+                        return obj != null;
+                    } catch (Exception e) {
+                        return false;
+                    }
+                }
+                return true;
+            }
 
             /** Returns the current value of this argument. Delegates to the current {@link #getter()}. */
             public <T> T getValue() throws PicocliException {
+                if ( !isValueGettable() ) { return null; }
                 try {
                     return getter.<T>get();
                 } catch (PicocliException ex) { throw ex;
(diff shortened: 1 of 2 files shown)


src/main/java/picocli/CommandLine.java
```diff
@@ -5995,6 +5995,15 @@ private Model() {}
          * @since 4.0
          */
         public interface IScope extends IGetter, ISetter {}
+        
+        /** This interface provides access to an {@link IScope} instance. 
+         */
```
remkop, line 6000: Can we add a  `@since` tag to the javadoc for this new interface?

```
@since 4.7
```

src/main/java/picocli/CommandLine.java
```diff
@@ -9166,9 +9175,25 @@ private String defaultValueFromProvider() {
              * @return whether this argument applies to all descendent subcommands of the command where it is defined
              * @since 4.3 */
             public ScopeType scopeType() { return scopeType; }
+            
+            /** Check whether the {@link #getValue()} method is able to get an actual value from the current {@link #getter()}. */
```
remkop, line 9179: Can we add a `@since` tag to the javadoc for this new public method?

```
@since 4.7
```

src/main/java/picocli/CommandLine.java
```diff
@@ -9166,9 +9175,25 @@ private String defaultValueFromProvider() {
...
+            public boolean isValueGettable() {
+                if (getter instanceof IScoped) {
+                    IScoped scoped = (IScoped) getter;
+                    IScope scope = scoped.getScope();
+                    try {
+                        Object obj = scope.get();
+                        return obj != null;
```
remkop, line 9186: The `IScope scope` object itself (what is returned from `scoped.getScope()`) may also be `null`.
We have a try/catch around this, so any NPEs would be caught and handled correctly, but should we do this instead?

```java
return scope != null && scope.get() != null;
```

remkop (changes requested):
Thank you for this PR!
I do prefer this approach to the IGetterAccessChecker alternative approach in https://github.com/remkop/picocli/pull/1801

I have some requests for minor changes, can you take a look?

src/main/java/picocli/CommandLine.java
rsenden: Implemented all changes as requested

remkop (approved):

remkop:
Merged. Thank you for the PR!

PR #1827, opened 2022-09-27
Change access modifier in READE.md

kogakzbj9:
Access modifier of MyCommand.x and positionals is private in  READE.  But they arer public in the actual sample code. 

https://github.com/remkop/picocli/blob/main/picocli-spring-boot-starter/src/test/java/picocli/spring/boot/autoconfigure/sample/MyCommand.java


remkop:
Hi @kogakzbj9; thank you for pointing out that inconsistency.
Yes, the sample code uses public fields in inner classes.
The reason that is done is for testing: the test code checks that the fields have the expected value afterwards.

In the documentation I do not want to suggest that fields should be public, so I prefer to keep the README as it is.
Thank you again for taking a close detailed look!

PR #1857, opened 2022-10-23
Document System.exit usage

### LINKED ISSUE #1855: Document a pattern that do not return exit code
I have been using picocli on the [micrometer-docs-generator](https://github.com/micrometer-metrics/micrometer-docs-generator) project and stumbled upon a problem that `System.exit` halt the maven process when exec-maven-java `exec:java` goal is used.

This is because `exec:java` uses the main maven process to invoke the target class and `System.exit` would simply halt the process that maven runs.

It is natural for a standalone CLI application to use the `System.exit`, but code invoked by some other application on the same process will have a problem using `System.exit`.

For my case, I initially tried to make the app a fat executable jar and make it invoked by `exec:exec` goal which spawns a new process, so it can use `System.exit`. https://github.com/micrometer-metrics/micrometer-docs-generator/pull/69
At the end, instead of the executable jar, I decided to go with simply throwing an exception AFTER `CommandLine#execute` is executed based on the returned exit code. 

```java
int exitCode = new CommandLine(new DocsGeneratorCommand()).execute(args);
if (exitCode != ExitCode.OK) {
    throw new IllegalStateException("DocsGeneratorCommand failed.");
}
```

I appreciate this pattern is also documented somewhere.
Since all I see is always using `System.exit`, it would be a problem for the environment where `System.exit` causes a negative effect such as `exec:java`.

ttddyy:
Documentation for a usecase not to use `System.exit`.

Resolves #1855


docs/index.adoc
```diff
@@ -3620,6 +3620,9 @@ fun main(args: Array<String>) {
 
 CAUTION: Older versions of picocli had some limited exit code support where picocli would call `System.exit`, but this is now deprecated.
 
+When an application is not appropriate to call `System.exit`, for example it runs on the caller's process, you can take a different action based on the returned `int`.
```
remkop, line 3623: This is the introduction section of the manual, I want to keep it very simple here, so I don't want to mention alternatives to calling `System.exit` just yet.

Let's remove the first sentence and only keep the following sentence that links to `<<Usage of System.exit>>`

docs/index.adoc
```diff
@@ -3620,6 +3620,9 @@ fun main(args: Array<String>) {
...
+See <<Usage of `System.exit`>> section for the usecases.
```
remkop, line 3624: typo: `usecases` -> `use cases`

docs/index.adoc
```diff
@@ -11401,6 +11404,31 @@ Multi-line text blocks can be used in command and option descriptions, headers a
 
 For more details, see https://www.infoq.com/articles/java-text-blocks/[this article] by Java Language Architect Brian Goetz.
 
+=== Usage of `System.exit`
+
```
remkop, line 11408: Instead of adding a new subsection **Usage of `System.exit`** to section **9. Executing Commands**, I propose we put all the text below in a new subsection under [21. Tips & Tricks](https://picocli.info/#_tips_tricks).

Then, we link to that new subsection from the [9.1. Exit Code](https://picocli.info/#_exit_code) subsection, by adding a sentence like this to the end of the 9.1. Exit Code subsection:

> There are cases where calling `System.exit` is not ideal. See < < NEW SECTION > > for details.

docs/index.adoc
```diff
@@ -11401,6 +11404,31 @@ Multi-line text blocks can be used in command and option descriptions, headers a
...
+In such scenario, you need to avoid `System.exit`.
+
+For example, picocli based <<TAB Autocomplete,TAB AutoComplete application>> solves this problem by having link:autocomplete.html#_maven_example[a system property] to determine whether to call `System.exit`.
+
+Another usecase is https://www.mojohaus.org/exec-maven-plugin/[`exec-maven-plugin`] with https://www.mojohaus.org/exec-maven-plugin/java-mojo.html[`exec:java` goal].
+This goal invokes the target class on the same maven process.
+The `System.exit` call finishes maven process bypassing the rest of the maven steps including its error handlings.
+
+In picocli, any exceptions thrown during the `CommandLine.execute` method are captured, printed, and converted to an exit code.
+You can check the exit code and perform an appropriate action, such as throwing an exception, instead of calling `System.exit`.
```
remkop, line 11421: This looks strange: you are correct that picocli captures any exceptions thrown during the `CommandLine.execute` method. That is the difference with the `CommandLine.parseArgs` method. If an application wants to throw an exception when a problem occurs, it should just use `parseArgs`. 

It would be strange to first capture an exception, convert to an error code, and then use that error code to throw another, less informative, exception...
The manual should not recommend that but instead should recommend using `parseArgs`.

docs/index.adoc
```diff
@@ -11401,6 +11404,31 @@ Multi-line text blocks can be used in command and option descriptions, headers a
...
+=== Usage of `System.exit`
+
+When an application runs as an independent process, `System.exit` returns an exit code to the caller.
+On the other hand, when an application runs as part of the caller process, `System.exit` halts the caller process and ends up an abrupt finish.
+
+In such scenario, you need to avoid `System.exit`.
+
+For example, picocli based <<TAB Autocomplete,TAB AutoComplete application>> solves this problem by having link:autocomplete.html#_maven_example[a system property] to determine whether to call `System.exit`.
+
+Another usecase is https://www.mojohaus.org/exec-maven-plugin/[`exec-maven-plugin`] with https://www.mojohaus.org/exec-maven-plugin/java-mojo.html[`exec:java` goal].
```
remkop, line 11416: typo: use case (space in between)

remkop (changes requested):
I made several suggestions for changes. Can you take a look?

docs/index.adoc
ttddyy: changed to recommend `parseArgs` and linked to the `9.7. DIY Command Execution` section

remkop:
Thank you!
I’m offline until mid next week, so feedback will take a while, thank you for your patience.

remkop:

PR #1908, opened 2023-01-02
Use the new official native-maven-plugin

tisonkun:
cc @remkop


remkop (approved):

remkop:
I will regenerate and publish the HTML when I get a chance.

Issue #725, opened 2019-06-15
Feature Request: fish completion


yschimke:
I think fish completion is generally much simpler than bash e.g. see examples below with baked in completions "atlassian github" etc.  But the functions at the top actually use the command itself to generate context sensitive completions by calling the code inside the app.  Good for URLs etc.

https://github.com/yschimke/cooee-cli/blob/master/src/dist/fish/cooee.fish

```
function __fish_complete_cooee_command
	cooee --fish-complete (commandline -ct)
end

function __fish_complete_add_command
	cooee --option-complete (commandline -ct)
end

function __fish_complete_remove_command
	cooee --option-complete (commandline -ct)
end

complete -c cooee -x -d 'Cooee CLI' -a '(__fish_complete_cooee_command)'
complete -c cooee -l debug -d 'Show debug output'
complete -c cooee -s h -l help -d 'Help options'
complete -c cooee -l login -d 'Login and link account'
complete -c cooee -l logout -d 'Unlink account'
complete -c cooee -s V -l version -d 'Output version and exit'
complete -c cooee -l repl -d 'REPL'

complete -c cooee -l list -d 'List apps'
complete -c cooee -l add -x -d 'Add app' -a '(cooee --option-complete add)'
complete -c cooee -l remove -x -d 'Remove app' -a '(cooee --option-complete remove)'

complete -c cooee -l token -x -d 'Provide specific authorization token'
#complete -c cooee -l tokenSet -x -d 'Set a named authorization token'

complete -c cooee -l authorize -x -d 'Authorize a provider' -a "atlassian github google trello strava"
complete -c cooee -l option-complete -x -a "option-complete command-complete authorize" -d 'Complete possible options'

complete -c cooee -l command-complete -d 'Complete possible command'
```

remkop:
Thanks for the suggestion!
Fish looks very interesting. 

Do you think you’ll be able to provide a pull request?

yschimke:
@remkop Fair question, It's unlikely I will in next 6 months due to 

1) I have working code that is app specific
2) I have other project commitments, work+personal projects+OkHttp
3) I'm hesitent to build something like this and not support it.  It would be a gamble that utility proved itself enough to gain support within existing users.

So feel free to close this and #726 but it's so palpably close to what I need to delete a tone of code in my own apps.

remkop:
No worries!

It is great if people can contribute a PR, so I always ask :-)

[...]

PR #1925, opened 2023-01-15
[WIP] Fish completion #725

volkov:
Fish completion POC (#725). Very dirty and buggy at the moment, but works nice for my scenario.

Definitely, it's early for reviewing of this changes, but early adopters of this thing are welcome.


remkop:
Awesome, @volkov thank you for working on this! 👍 
Will you be able to provide some unit tests?

volkov:
Yes, I have begun to move towards testing, however, my progress is currently moderate as I am not familiar with dejagnu
I think I'll be back in couple of weeks.

remkop:
Hi @volkov looks like you have made great progress! I took a brief look and I see you are also adding dejaGnu tests, much appreciated! I remember it took me a long time to wrap my head around the dejaGnu test framework, not trivial! 😅 

Were you able to make dejaGnu start a `fish` shell session instead of a `bash` session? I took a quick look at `src/test/dejagnu.tests/lib/library.exp` (I suspect that some work may be needed there) but I haven't figured it out yet. Looking at the `start_bash` function, it may be as simple as setting `--tool_exec` to `fish`, but perhaps that is not even needed, not sure...

If your dejaGnu tests work, can we call them from JUnit, maybe from `src/test/java/picocli/AutoCompleteDejaGnuTest.java`?

Again, thanks a lot for your contributions here! Much appreciated! 🙏

src/main/java/picocli/AutoComplete.java
```diff
@@ -534,6 +534,123 @@ public static String bash(String scriptName, CommandLine commandLine) {
...
+    }
+
+    private static void processLevel(String scriptName, StringBuilder result, List<CommandDescriptor> currentLevel,
+                                     List<String> currentLevelCommands, String levelName,
+                                     CommandDescriptor rootDescriptor) {
+        if (levelName.equals("")) {
+            levelName = "root";
+        }
+        result.append("\n# ").append(levelName).append(" completion\n");
+        result.append("set -l ").append(levelName).append(" ").append(String.join(" ", currentLevelCommands)).append(
```
remkop, line 581: `String.join` is Java 8 API, and picocli is compiled with Java 5, so we cannot use that method. We will have to write our own `join` utility method.

src/main/java/picocli/AutoComplete.java
```diff
@@ -534,6 +534,123 @@ public static String bash(String scriptName, CommandLine commandLine) {
...
+                "\n");
+        if (rootDescriptor != null) {
+            for (OptionSpec optionSpec : rootDescriptor.commandLine.getCommandSpec().options()) {
+                result.append("complete -c ").append(scriptName);
+                result.append(" -n \"not __fish_seen_subcommand_from $").append(levelName).append("\"");
+                result.append(" -l ").append(optionSpec.longestName().replace("--", ""));
+
+                if (optionSpec.completionCandidates() != null) {
+                    result.append(" -f -a '").append(String.join(" ", extract(optionSpec.completionCandidates()))).append("' ");
```
remkop, line 590: Similar as above, we cannot use `String.join`.

src/main/java/picocli/AutoComplete.java
```diff
@@ -534,6 +534,123 @@ public static String bash(String scriptName, CommandLine commandLine) {
...
+                result.append("complete -c ").append(scriptName);
+                result.append(" -n \"__fish_seen_subcommand_from ").append(commandDescriptor.commandName).append("\"");
+                if (!commandDescriptor.parentWithoutTopLevelCommand.equals("")) {
+                    result.append(" -n '__fish_seen_subcommand_from ").append(
+                            commandDescriptor.parentWithoutTopLevelCommand).append("'");
+                }
+                result.append(" -l ").append(optionSpec.longestName().replace("--", ""));
+
+                if (optionSpec.completionCandidates() != null) {
+                    result.append(" -a '").append(String.join(" ", extract(optionSpec.completionCandidates()))).append("' ");
```
remkop, line 628: Similar as above, we cannot use java 8 `String.join`.

remkop (changes requested):
I found some usages of Java 8 API, but picocli is compiled with Java 5, so we cannot use that (`String.join`). Can you take a look?

volkov:
> Were you able to make dejaGnu start a `fish` shell session instead of a `bash` session?

Yes, At the moment I use separate file [dejagnu.fishtests/lib/completion.exp](https://github.com/remkop/picocli/blob/5b1cbc89b60baa04c8695b49f4ccb5927e36885a/src/test/dejagnu.fishtests/lib/completion.exp) with `exp_spawn fish --no-config`

> If your dejaGnu tests work, can we call them from JUnit, maybe from src/test/java/picocli/AutoCompleteDejaGnuTest.java?

Fish tests work on my notebook, at the moment there is only subset of bash tests. Bash tests for some reason don't work on my notebook, but I didn't investigate it yet.

Yes, It's good idea to integrate them with junit.

Also I think its important to make them work in github actions.

I'll try to return to this pr in couple of weeks.

remkop (approved):

azdanov:
Hey. I found out about this PR from https://github.com/quarkusio/quarkus/issues/22710. I guess that currently it's out of date and would need to be rebased with main and fix any potential issues?

maxandersen:
i've rebased this PR with main in #2463

cmoulliard:
What is the status of this PR ? @volkov

PR #1955, opened 2023-02-12
Bump asciidoctorj to latest version

deining:
This PR bumps dependency `asciidoctorj` to latest released version.


remkop:
I wonder if this version can be moved to https://github.com/remkop/picocli/blob/main/dependencies.gradle so it will become maintained automatically by dependabot?

remkop (approved):

remkop:
I tried moving this version into depedencies.gradle but could not get it to work...
Giving up on that idea. ;-)

PR #1976, opened 2023-03-19
Fixed isJansiConsoleInstalled performance issue

### LINKED ISSUE #1975: ~150 millisecond hotspot under the picocli.CommandLine$Help$Ansi$Text.toString method
image

In my program's initialization, there's a bunch of `Class.forName` calls, half of them seem to come from `picocli's CommandLine$Help$Ansi$Text.toString` method.  It appears to be looking for `jansi` classes.  The above screenshot shows this is taking about 150 milliseconds to complete.  I can confirm that passing in `-Dpicocli.ansi=false` to the Java commandline works around this issue.  I'm running this on Windows.

https://github.com/remkop/picocli/blob/051a51d07ef917913aec53759ceeac0893591617/src/main/java/picocli/CommandLine.java#L17800-L17829

I don't have `org.fusesource.jansi` on my classpath.  I wouldn't expect the total time for the `isJansiConsoleInstalled` method to take 150 milliseconds, unless it was repeatedly called in a loop, which I don't think it is.  It should complete instantaneously.

Anyone know what's going on here?

ChrisTrenkamp:
Fixes #1975 

isJansiConsoleInstalled() was repeatedly called
in a loop from the TextTable#toString method,
which caused performance problems due to the
constant calls to Class#forName.  This change
caches the result from isJansiConsoleInstalled so
it doesn't repeatedly call Class#forName, which
cannot change at runtime.

### DIFF
--- a/picocli-tests-java567/src/test/java/picocli/HelpAnsiHeuristicsTest.java
+++ b/picocli-tests-java567/src/test/java/picocli/HelpAnsiHeuristicsTest.java
@@ -360,6 +360,11 @@ public class HelpAnsiHeuristicsTest {
         environmentVariables.clear(ANSI_ENVIRONMENT_VARIABLES);
         environmentVariables.set("CLICOLOR", "0"); // hint disabled
         System.setProperty("os.name", "Windows");
+        // Clear the globally cached jansiConsole value that might
+        // have been set in a previous test to force the
+        // Ansi#isJansiConsoleInstalled method to recalculate
+        // the cached value.
+        Ansi.jansiConsole = null;
         assertTrue(Ansi.isWindows());
         assertFalse(Ansi.isPseudoTTY());
         assertFalse(Ansi.forceDisabled());
--- a/picocli-tests-java9plus/src/test/java/picocli/AutoCompleteSystemExitTest.java
+++ b/picocli-tests-java9plus/src/test/java/picocli/AutoCompleteSystemExitTest.java
@@ -1,9 +1,11 @@
 package picocli;
 
 
+import org.junit.jupiter.api.BeforeAll;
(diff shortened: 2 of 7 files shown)


remkop (dismissed):

remkop:
@ChrisTrenkamp I had to enable CI/CD for this PR, but now I see some tests are failing:


```
## On Java 7 Ubuntu
picocli.HelpAnsiHeuristicsTest > testAnsiAutoJansiConsoleInstalledOverridesHintDisabled FAILED
    java.lang.AssertionError at HelpAnsiHeuristicsTest.java:373
```
and

```
## On Java 18 Windows
picocli.HelpAnsiTest > testAnsiEnabled FAILED
    java.lang.AssertionError at HelpAnsiTest.java:132
```

This may be due to the order in which tests are executed.
Perhaps all JansiConsoleInstalled-related tests need to be moved to separate test classes (a separate class per test)?

ChrisTrenkamp:
The failing test cases are passing now, but it had to be done by resetting the cached value.  Doesn't feel like the most elegant way to handle this.

remkop:
I see what you mean but I cannot think of an alternative either. 😅
I think this is fine. 
Thank you fixing the tests!

remkop (approved):

remkop:
Having some trouble with the CI/CD, but I do see this failing test:

```
picocli.HelpAnsiHeuristicsTest > testAnsiAutoJansiConsoleInstalledOverridesHintDisabled FAILED
    java.lang.AssertionError at HelpAnsiHeuristicsTest.java:378
```

remkop:
Merged. Sorry, it took some time to sort out the CI/CD issues.
Many thanks for the contribution!

PR #2036, opened 2023-06-01
Fixes #2035 - include mapFallbackValue for inherited opts

### LINKED ISSUE #2035: Option "mapFallbackValue" ignored when inherited to subcommand
With an @Option defined like this: 
```java
@Command(
    name = "main"
    subcommands = {
        SubCmd.class,
        CommandLine.HelpCommand.class
    }
)
public class Main {
    @Option(
            names = "-D",
            paramLabel = "<key=value>",
            mapFallbackValue = Option.NULL_VALUE,
            scope = ScopeType.INHERIT
    )
    private final Map<String, String> javaOpts = new HashMap<>();
}

@Command(name = "sub")
public class SubCmd {
    @ParentCommand
    private Main parent;
}
```

Running with args like ["sub", "-Dxyz"] results in error message:

    Value for option option '-D' (<key=value>) should be in KEY=VALUE format but was xyz

This seems like it might be because the ArgSpec Builder copy constructor doesn't copy the "mapFallbackValue" field in picocli 4.7.3:  https://github.com/remkop/picocli/blob/v4.7.3/src/main/java/picocli/CommandLine.java#L9496

zman0900:
This fixes issue #2035 seen in my own code when using "mapFallbackValue" on an inherited Option.  Also added some unit tests for both inherited Option and Parameters that all failed without this fix.

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -9526,6 +9526,7 @@ public class CommandLine {
                     setter = original.setter;
                     scope = original.scope;
                     scopeType = original.scopeType;
+                    mapFallbackValue = original.mapFallbackValue;
                     originalDefaultValue = original.originalDefaultValue;
                     originalMapFallbackValue = original.originalMapFallbackValue;
                 }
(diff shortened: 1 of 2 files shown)


remkop (approved):

remkop:

Issue #1965, opened 2023-02-24
Generated docs: git chokes on mixed line endings


deining:
**How to reproduce:**

```
git clone https://github.com/remkop/picocli.git
cd picocli/
```

I'm on a linux box. Immediately after cloning the repo, I do see modified files:

```
git status
On branch main
Your branch is up to date with 'origin/main'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   docs/A-Whirlwind-Tour-of-Picocli.html
        modified:   docs/announcing-picocli-1.0.html
        modified:   docs/autocomplete.html
        modified:   docs/build-great-native-cli-apps-in-java-with-graalvm-and-picocli.html
        modified:   docs/feedback.html
        modified:   docs/groovy-2.5-clibuilder-renewal-part1.html
        modified:   docs/groovy-2.5-clibuilder-renewal-part2.html
        modified:   docs/groovy-2.5-clibuilder-renewal.html
        modified:   docs/index.html
        modified:   docs/man/index.html
        modified:   docs/migrating-from-commons-cli.html
        modified:   docs/picocli-2.0-do-more-with-less.html
        modified:   docs/picocli-2.0-groovy-scripts-on-steroids.html
        modified:   docs/picocli-on-graalvm.html
        modified:   docs/picocli-programmatic-api.html
        modified:   docs/quick-guide.html
        modified:   docs/zh/picocli-2.0-do-more-with-less.html
        modified:   docs/zh/picocli-2.0-groovy-scripts-on-steroids.html

no changes added to commit (use "git add" and/or "git commit -a")
```

`git config --get core.autocrlf` doesn't print out anythig, so I think the default value of `false` applies.

I do have an idea what is causing this behavior: the generated html files do have **mixed line endings**.

For e.g. `docs/man/index.html/` has these mixed line endings (on a Linux checkout):

Line 1-1711: LF
Line 1712-1786: CR/LF
Line 1787-19079: LF

AFAICS, The CR/LF in line 1712 - 1786 are introduced by the gradle [asciidoctorj-tabbed-code-extension](https://github.com/bmuschko/asciidoctorj-tabbed-code-extension) that is used from the gradle `asciidoc` task.

https://github.com/remkop/picocli/blob/cc0602bf758042edce1759992ea79358e3651e87/gradle/docs.gradle#L2-L10

[...]

PR #2045, opened 2023-06-13
Commit the html files with the right line-ends

fridrich:
Every time git touches the repository, the CRLF->LF conversion is done. There are still html files in docs that have the wrong line-end. So every time one does for instance "git pull", there are local uncommitted changes. This pull request simply pushes the html files after the conversion, so that they have the right line-endings and so that there are not local changes generated every time.


remkop:
Sorry for the late response @fridrich, this dropped off my inbox horizon...

I remember [doing something similar recently](https://github.com/remkop/picocli/issues/1965) after feedback from @deining; and I am guessing that perhaps the latest 4.7.4 release has undone my manual changes.
I can merge your PR, I don't see any issue with that, but I suspect that after the next release we will have the same issue again. So I am wondering what causes this...

I build releases on Windows.
If you look at [the `.gitattributes` file](https://github.com/remkop/picocli/blob/main/.gitattributes) for this project, you will see this line:

```
*.html          text eol=lf
```

My understanding is that this configuration should cause git to convert any CRLF line endings to LF before committing, so that the `.html` files in the repository only have LF line endings.
Apparently this is conversion not happening...

Looking at [this StackOverflow answer](https://stackoverflow.com/questions/9976986/force-lf-eol-in-git-repo-and-working-copy), and the [linked article](https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/), it is also possible that your git configuration for `core.eol` is incorrect. My understanding is that `native` is the default, and likely the value you want.

You can run `git config --global core.eol` to see what this value is set to on your system. If nothing comes back that means you are on the using the default which is [native](url).

Next, the article has some details on what happens when git writes to the object database and to the working directory. 
Looking at the `.gitattributes` file for the picocli project, everything looks correct.
The only thing I could think of was that perhaps you are using what the article calls The Old System, which depends on the `core.autocrlf` configuration setting.

You can run `git config --global core.autocrlf` to see what this value is set to on your system. If nothing comes back that means you are on the using the default which is `false`. If you work on Windows you may want to set this to `true`.

Can you take a look at this, try the above and let me know your findings?

remkop:
Hi @fridrich have you had a chance to look at my previous comment? I'm curious to your feedback.

fridrich:
> Hi @fridrich have you had a chance to look at my previous comment? I'm curious to your feedback.

I mean that configuration converts the files every time git is invoked. Just those files under version control have CRLF line endings. It is needed to commit them then with the new LF line ending. If they are not committed, they will always be uncommitted changes in the git repository. Even if you do git reset --hard

remkop (approved):

remkop:
Hi @fridrich I merged your PR, thank you!

I also just did a release (4.7.5), so many HTML files were generated from the Asciidoc files again.
I hope that this did not mess up the line endings again, can you check?

deining:
> I also just did a release (4.7.5), so many HTML files were generated from the Asciidoc files again. I hope that this did not mess up the line endings again, can you check?

Just gave it a try, and everything is fine! Seems like this issue is resolved eventually! Thanks to all for their efforts!

remkop:
@deining Thank you for checking!

Issue #2102, opened 2023-09-01
Feature request: PropertiesDefaultProvider at first try to load properties from classpath


rimuln:
Would be possible enhance this class to load properties from resources too?

Actual state:
1. try to load from user.home
2. try to load from environment

Wanted state:
1. try to load from classpath via getResourceAsStream
2. try to load from user.home
3. try to load from environment

or allow configure IDefaultValueProviders as ordered list?

remkop:
That sounds like a good enhancement!
One caveat is that I am very hesitant to modify existing default behavior. 
So my preference would be to load from classpath after trying the current methods, not before. 

My time to work on picocli is extremely limited. 
Will you be able to provide a pull request for this?

rimuln:
I understood your opinion. But the oposite order is for us problematic. 
In general. 
You as developer distribute application with some default settings that could be bundled in jar. Then you expect that user is able override this default settings from outside.

If you pass the loading defaults from classpath as last in order then you always override user defined settings.
And if jar doesn't contains the settings then I would say the default logic is unchanged.

But if I look on it from implementation. You are right first is current code and after this is load from classpath. So theoreticaly this way:
```
        private static Properties loadProperties(CommandSpec commandSpec) {
            if (commandSpec == null) { return null; }
            Properties p = System.getProperties();
            for (String name : commandSpec.names()) {
                String propertiesFileName = "." + name + ".properties";
                URL resource = CommandLine.class.getClassLoader().getResource(propertiesFileName);
                String path = p.getProperty("picocli.defaults." + name + ".path");
                File defaultPath = new File(p.getProperty("user.home"), propertiesFileName);
                File file = path == null ? defaultPath : new File(path);
                if (file.canRead()) {
                    return createProperties(file, commandSpec);
                } else if (resource != null) {
                    file = new File(resource.getFile());
                    if (file.canRead()) {
                        return createProperties(file, commandSpec);
                    }
                }
            }
            return loadProperties(commandSpec.parent());
        }
```

rimuln:
ad 1) ok no problem

ad 2) yes this come to my mind to. In our case it is all in one fat jar made by maven, but I see in docu that picocli.jar could be separate jar so wasn't sure how it works. The command class is 100% better, just I didn't look yet how get it from CommandSpec object.

ad 3) not sure what you mean. The original logic try to load from -Dpicocli.defaults.<name>.path, and if not present then from user.home This is still present. 

[...]

PR #2107, opened 2023-09-05
#2102 PropertiesDefaultProvider at first try to load properties from …

rimuln:
…classpath

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -18972,11 +18972,25 @@ public class CommandLine {
             if (commandSpec == null) { return null; }
             Properties p = System.getProperties();
             for (String name : commandSpec.names()) {
+                String propertiesFileName = "." + name + ".properties";
                 String path = p.getProperty("picocli.defaults." + name + ".path");
-                File defaultPath = new File(p.getProperty("user.home"), "." + name + ".properties");
+                File defaultPath = new File(p.getProperty("user.home"), propertiesFileName);
                 File file = path == null ? defaultPath : new File(path);
                 if (file.canRead()) {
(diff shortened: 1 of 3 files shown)


remkop (changes requested):
This looks good overall, many thanks!

Will you be able to add a unit test please?

One minor point is that the user object of the CommandSpec may be `null` in some circumstances (for applications that use the programmatic API to create a CommandSpec, instead of reflection from annotated application classes).
So the logic needs a minor change to accommodate this.

src/main/java/picocli/CommandLine.java
```diff
@@ -18972,11 +18972,25 @@ private static Properties loadProperties(CommandSpec commandSpec) {
...
             for (String name : commandSpec.names()) {
+                String propertiesFileName = "." + name + ".properties";
                 String path = p.getProperty("picocli.defaults." + name + ".path");
-                File defaultPath = new File(p.getProperty("user.home"), "." + name + ".properties");
+                File defaultPath = new File(p.getProperty("user.home"), propertiesFileName);
                 File file = path == null ? defaultPath : new File(path);
                 if (file.canRead()) {
                     return createProperties(file, commandSpec);
+                } else {
+                    URL resource = commandSpec.userObject().getClass().getClassLoader().getResource(propertiesFileName);
```
remkop, line 18982: Minor point here is that the commandSpec's `userObject` may be `null` in some circumstances.
In that case we have little choice but to fall back to the class loader of the `picocli.CommandLine` class.
Can I ask you to make that change?

rimuln:
> Will you be able to add a unit test please?
> 
something tried if not enough not sure if I get quickly deeper insights in tests

PR #2148, opened 2023-11-11
Fix NPE in jline3 Example as ConfigurationPath cannot be null anymore

llzen44:
ConfigurationPath cannot be null anymore since commit 9706ead in jline3. An Objects.requireNonNull check now requires a valid ConfigurationPath.


remkop (changes requested):
That looks good.
Can you also update the `picocli-shell-jline3/README.md` file, which shows the same code?
Thank you!

llzen44:
Sure thing. Done
Thanks for the review

remkop (approved):

remkop:

Issue #2054, opened 2023-06-26
Add Error Prone


vorburger:
How about adding https://errorprone.info/docs/installation to PicoCLI? Using (https://github.com/tbroyer/gradle-errorprone-plugin.)

It would have found the #2053 problem, and possibly others, and will prevent regressions of them, and future new bugs.

@remkop would a PR for this be a welcome contribution?

remkop:
Hi @vorburger, yes that seems like a good idea!

One note of caution may be with the different Java versions.
When building picocli and creating releases, I use JDK 8.
Additionally, the GitHub Actions-based CI compiles and runs tests using various other JDK versions.
Both use cases would need to be supported.

If that is possible, then let's go for it!

clebertsuconic:
I have a love / hate relationship with errorprone.

While it gets some issues eventually, it generates a lot of noise that unless you are willing to fix every single warning coming from errorprone those will get ignored.

Picocli is not that huge, so it won't add much time to the build, but one issue I have with it in artemis is that it takes a good amount to time to process.


IMO though, picocli is very small. and I appreciate the fact it is small. I don't see much benefit from using errorprone.

vorburger:
> IMO though, picocli is very small. and I appreciate the fact it is small. 

@clebertsuconic adding Error Prone to the PicoCLI build (only!) will keep the PicoCLI JAR and API as small as it is... 😺 

> When building picocli and creating releases, I use JDK 8. Additionally, the GitHub Actions-based CI compiles and runs tests using various other JDK versions. Both use cases would need to be supported.

@remkop I see! Because [you use a Matrix Build](https://github.com/remkop/picocli/blob/main/.github/workflows/ci.yml) it wouldn't "just work". Given https://errorprone.info/docs/installation stating _"Please note that Error Prone must be run on JDK 11 or newer.",_  the best way maybe to simply have the [`gradle-errorprone-plugin`](https://github.com/tbroyer/gradle-errorprone-plugin) run only on Java 11+ ... I've never done this with Gradle, but given how imperative Gradle is, and that it can bake Pizzas in addition to building software, presumably there is some way to [make it so](https://www.youtube.com/watch?v=FaLyasJPyUU)... does anyone reading this happen to know how to do that? Otherwise I'll look a bit more into it when I have some time.

vorburger:
> presumably there is some way to [make it so](https://www.youtube.com/watch?v=FaLyasJPyUU)

[...]

PR #2175, opened 2023-12-15
Remove unused extra format arguments

### LINKED ISSUE #2053: FormatString bug in AutoComplete found by Error Prone
I'm running https://errorprone.info/ on this project, and it claims to have found a bug:

```
picocli/src/main/java/picocli/AutoComplete.java:788 : error: [FormatString] extra format arguments: used 2, provided 3
                buff.append(format("%s      positionals=$( compReplyArray \"${%s_pos_param_args[@]}\" )\n", indent, paramName, currWord));

picocli/src/main/java/picocli/AutoComplete.java:832 : error: [FormatString] extra format arguments: used 2, provided 3
                buff.append(format("%s      COMPREPLY=( $( compReplyArray \"${%s_option_args[@]}\" ) )\n", indent, bashify(option.paramLabel()), currWord));
```

See https://errorprone.info/bugpattern/FormatString ... @remkop  from a quick glance, I think this is a real bug, not a false positive? As far as I understand it in both of those cases there are indeed only x2 `%s` but 3 arguments. 

I would by happy to send a CL, but it's not immediately clear to me which of the 3 is wrong and should be removed.

vorburger:
fixes #2053


vorburger:
@jsotuyod this is re. [this](https://github.com/remkop/picocli/pull/1759#discussion_r1245058532) and [this](https://github.com/remkop/picocli/pull/1759#discussion_r1244849171) - I hope this looks right? Tx!

PS: The upcoming PR for #2054 will autodetect this during the build.

remkop (approved):

PR #2185, opened 2023-12-19
Mark GROOVY_CLOSURE_CLASS as final and private

vorburger:


remkop (approved):
I had omitted this because it is a nested class inside the `private static class DefaultFactory` class. However I have no objection to this change.

vorburger:
> I had omitted this because it is a nested class inside the `private static class DefaultFactory` class. However I have no objection to this change.

FYI Some Checkstyle-like (internal) tool pointed out the missing final, just because of the all uppercase name.

The private may indeed be unnecessary technically, I just added that for consistency while I was at it (the tool did not).

PR #2347, opened 2024-10-31
I hope the last of the saga of line-endings

fridrich:
The change in the .gitattributes is crucial. If we leave it only "text",  the EOL will be platform dependent. So, I changed all those where the comments say that they should be changed to eol=lf. Then I had to manually add those files that were having some CRLF before and now the git is not marking them as modified any more.
What introduced the CRLF it was the release commit. I hope with this change in .gitattributes, that will not happen any more even if the docs are generated on a windows machine.


remkop (approved):

remkop:
Apologies for the very long wait. 
This has been merged.
Thank you for your contribution!

Issue #1804, opened 2022-09-08
Add support for pluggable ITypeConverterFactory 


remkop:
(See discussion in https://github.com/remkop/picocli/issues/1707)

There are cases where applications want programmatic control of the type converter used.
The use case mentioned in #1707 is a single type converter for many Enum classes.
Another potential use case is a single type converter for all implementations of an interface.

The proposed interface is as follows:

```java
/**
 * Interface that allows customization of what type converters to use for specific types
 * and options or positional parameters.
 * If a type converter factory is set for a command, then this factory is first queried to produce a type converter.
 * If no type converter factory is set, or the factory returns a null type converter, 
 * then the default picocli logic is used to attempt to find a type converter.
 * @since 4.x
 */
interface ITypeConverterFactory {
    /**
     * Returns a type converter for the specified type for the specified option or positional parameter,
     * or returns {@code null} if the default converter should be used.
     * 
     * @param type the type (or supertype) for which to return a type converter
     * @param argSpec the option or positional parameter for which to return a type converter
     * @return a type converter or {@code null}
     * @since 4.x
     */
    ITypeConverter<? extends T> createTypeConverter(Class<T> type, ArgSpec argSpec);
}
```

As is the convention with other getters/setters, applications will be able to install a custom type converter for the full command hierarchy via a getter and setter in `CommandLine`:

```
/** Sets the specified type converter factory for this command and all subcommands. */
CommandLine::setTypeConverterFactory(ITypeConverterFactory) : void

/** Returns the custom type converter factory for this command, or {@code null} if none was set. */
CommandLine::getTypeConverterFactory() : ITypeConverterFactory
```

As is the convention with other getters/setters, applications will be able to install a custom type converter on a per-command basis via a getter and setter in `CommandSpec`:

```
/** Sets the specified type converter factory for this command only. */
CommandSpec::typeConverterFactory(ITypeConverterFactory) : void

/** Returns the custom type converter factory for this command, or {@code null} if none was set. */
CommandSpec::typeConverterFactory() : ITypeConverterFactory
```

There is no default implementation, that is, by default, `CommandLine::getTypeConverterFactory()` returns `null`.
If a type converter factory is set, and it returns a non-null type converter, then this converter is used, otherwise the default picocli logic is used to attempt to find a type converter.

@garretwilson Thoughts/feedback?

garretwilson:
@remkop I think that provides the functionality needed to hook into type conversion for all enums types.

[...]

PR #2449, opened 2025-07-12
Pluggable enum conversion facility for picocli.

garretwilson:
Pluggable enum conversion facility for remkop/picocli#1804 .

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -3422,7 +3422,7 @@ public class CommandLine {
      * @see #registerConverter(Class, ITypeConverter)
      */
     public CommandLine registerEnumConverter(IEnumConverter enumConverter) {
-        interpreter.enumConverter = Assert.notNull(enumConverter, "enumCconverter");
+        interpreter.enumConverter = Assert.notNull(enumConverter, "enumConverter");
         for (CommandLine command : getCommandSpec().commands.values()) {
             command.registerEnumConverter(enumConverter);
         }


remkop (changes requested):
Thank you for the pull request. I will look in more detail later (so I may have more feedback later) but overall it looks good. 

Please add documentation and tests and I think this can be merged.

Please add a section "Custom Enum Type Converters" to the user manual 
(located here: `docs/index.adoc` - HTML is auto-generated from this).
I suggest after section `=== Handling Invalid Input` and
before section `=== Option-specific Type Converters`.

Also, please provide JUnit tests that verify at least

- that registering a custom enum type converter will result in that enum type converter being used for options and positional parameters with that enum type
- that the custom enum type converter will NOT be used if a custom `ITypeConverter` is registered for that enum type
- a custom enum type converter falling back to the default enum type converter
- setting `CommandLine::setCaseInsensitiveEnumValuesAllowed` to `true` prior to parsing will result in the custom enum type converter being called with value `true`
- not calling `CommandLine::setCaseInsensitiveEnumValuesAllowed` prior to parsing will result in the custom enum type converter being called with value `false`
- throwing a TypeConversionException from the registered custom enum type converter will result in a `ParameterException` with a user-friendly message from `CommandLine::parseArgs`
- registering a `null` custom enum type converter is rejected immediately with a NullPointerException (not silently accepted, causing problems during parsing)
- calling `CommandLine::registerEnumConverter` will register the custom enum type converter for that command and all its subcommands and sub-subcommands

(most tests are here: `src/test/java/picocli`)

src/main/java/picocli/CommandLine.java
```diff
@@ -5017,6 +5048,67 @@ public interface ITypeConverter<K> {
...
+        @Override
+        public <E extends Enum<E>> E convert(String value, Class<E> enumType, boolean caseInsensitiveEnumValuesAllowed) throws Exception {
+            try { return Enum.valueOf(enumType, value); }
+            catch (IllegalArgumentException ex) {
+                for (E enumConstant : enumType.getEnumConstants()) {
+                    String str = enumConstant.toString();
+                    String name = enumConstant.name();
+                    if (value.equals(str) || value.equals(name) || caseInsensitiveEnumValuesAllowed && (value.equalsIgnoreCase(str) || value.equalsIgnoreCase(name))) {
+                        return enumConstant;
+                    }                            }
```
garretwilson, line 5094: I just noticed this formatting oddity and thought I had made a typo, but then I saw it was in the original code, so I'll leave it as-is for the moment.

remkop: I noticed it too in the original code. Looks like a newline was accidentally deleted. It was not intentional, please fix in the PR.

remkop:
@garretwilson The project now builds correctly on Java 24 and Gradle 8.14. The Github CI build now also includes Java 24. 
Thank you for raising this. Hopefully this addresses your concerns.

garretwilson:
Unfortunately, even adding a JUnit 4 `@Ignore` to `Issue1125_1538_OptionNameOrSubcommandAsOptionValue`, Gradle doesn't complete tests:

```
…
> Task :picocli-codegen-tests-java9plus:test

Issue1380Test > testingWithExclusiveTrue FAILED
    org.junit.ComparisonFailure at Issue1380Test.java:68

Issue1380Test > testingWithExclusiveFalse FAILED
    org.junit.ComparisonFailure at Issue1380Test.java:87
…
```

It's at this point I realize again that going in circles is going to waste the whole morning, so I'm stopping this for today.

Could you make sure that `gradle test` works from the command line on Java 24 so that I can have working suite to start with before adding to the tests? I don't want to be fighting the failing tests at every turn. Thanks.

remkop:
I am guessing that the failing tests were because the usage help message had ANSI escape codes in the output in your environment. 

I fixed the tests to ensure that the comparison is without the escape codes regardless of the environment.

garretwilson:
Good news — I am now able to run `gradle test` on this branch and it succeeds. Thank you for fixing those tests.

garretwilson:
FYI, using `gradle test` with Gradle 8.14.3 still works with new LTS Java 25! But there are a significant number of API deprecation warnings, so I recommend you start building with a higher Java version and addressing these. Here are a couple of examples:

```
…\picocli-examples\src\main\java\picocli\examples\i18n\I18NDemo.java:22: warning: [deprecation] URL(String) in URL has been deprecated
        try (java.util.Scanner s = new java.util.Scanner(new java.net.URL(URL).openStream())) {
                                                         ^
…\picocli-examples\src\main\java\picocli\examples\i18n\localecontrol\LocaleControl.java:21: warning: [deprecation] Locale(String) in Locale has been deprecated
        Locale.setDefault(new Locale(locale));
```

garretwilson:
@remkop I've added the unit tests you've requested. As I mentioned in issue #1804, I started with tests an LLM generated, and I haven't looked at the tests super closely. However I did glance over them just now and tweaked them a little. 

It's amazing, after I got the LLM aligned with what I was looking for (e.g. specifying a Java 8 build, JUnit 4, consistency with existing picocli tests, etc) how little tweaking I needed to do. Here are some example:

* I changed `received.set(Boolean.valueOf(caseInsensitiveEnumValuesAllowed))` to `received.set(caseInsensitiveEnumValuesAllowed)` to use autoboxing, because I didn't see the point of spelling it out the conversion.
* In one JUnit test that expected a `NullPointerException`, for some reason it had used the ancient-style of catching and ignoring an exception while using `fail()` if the exception did _not_ happen. I simplified this using `@Test(expected = NullPointerException.class)`.
* In the last test it was wrapping the `parseArgs()` calls in a `try … catch (Exception ignored) { }`, and I removed these because they aren't useful or even appropriate here (and ignoring exceptions in general are 99.999% of the time not what should be done anyway). In further conversations with the LLM it confirmed these `try … catch` blocks weren't appropriate. (Why it put them there in the first place illustrates some interesting aspects of LLMs that we can discuss over a drink some day.) Frankly the fact that an LLM could put that reasonably complex test together based upon such a vague request, understanding the code and not only testing inheritance but also using a technique to distinguish among subcommands, is amazing in itself.

I haven't updated any documentation; I'll have to do that on another day when I can find another few minutes free. In the meantime please take a look at the tests and tell me if they're OK. Please also respond to the earlier [question I had](https://github.com/remkop/picocli/issues/1804#issuecomment-3368697449) in the main ticket about how you want to test a "user-friendly error message". Thanks and have a great weekend.

garretwilson:
> Please add a section "Custom Enum Type Converters" to the user manual  
> (located here: `docs/index.adoc` - HTML is auto-generated from this).  
> I suggest after section `=== Handling Invalid Input` and  
> before section `=== Option-specific Type Converters`.

I took a look at the docs, and section "4.2. Custom Type Converters" is the one about converters, so I'm thinking right _before_ section `=== Handling Invalid Input` (i.e. at the end of the "Custom Type Converters" section) might be better. (Technically before "Multi Parameter Type Converters" might have even been better, but I was afraid of changing the numbering.)

Take a look and let me know what you think. I think these docs are the last thing to finalize for this PR.

PR #2463, opened 2025-08-25
feature/fish completion

maxandersen:
rebased #1925 as I realized fish now works with jbang scripts so I got a good reason to make this work :)

I've added basic support to the GenerateCompletion command for choosing which variant to generate bash/zsh or fish.

But I didn't grok what the meaning is for the AutoComplete which generates code does does not run without user modification so i did not add/change that. 

I've tested it in jbang at https://github.com/jbangdev/jbang/pull/2195 and would love to see a picocli release with this included so I can release jbang with fish completion :)


remkop:
Hi @maxandersen thank you for helping push this forward.
Fish completion, exciting!

I see this one test that now seems to be failing for some reason:

```
> Task :picocli-codegen-tests-java9plus:test

NativeImageConfigGeneratorProcessorTest > testGenerateReflectForNestedInnerEnumWithoutPackage FAILED
    com.google.common.truth.ComparisonFailureWithFacts at NativeImageConfigGeneratorProcessorTest.java:235

62 tests completed, 1 failed, 6 skipped
```

Does this need to be added to `picocli-codegen-tests-java9plus/src/test/resources/expected-filelist-reflect.json`?

```json
  {
    "name" : "picocli.AutoComplete$GenerateCompletion$Shell",
    "allDeclaredConstructors" : true,
    "allPublicConstructors" : true,
    "allDeclaredMethods" : true,
    "allPublicMethods" : true
  },
```

maxandersen:
sounds about right. pushed.

maxandersen:
the failure on java 9 is that it does not seem available on github actions?
also for java 6 is failure  - but i can't reproduce locally; can one see the test results so can see actual comparison failure?

PR #2497, opened 2025-11-26
Better performance

claraacv:
For a better perfomance, string concatenations in a loop were change to StringBuilder objects.


remkop (commented):
To be honest, I’m not a fan of making the code more verbose for no reason. 

Can you quantify the performance gains, and provide context to when (which scenarios) performance would be better? It would have to be significantly faster in common scenarios for me to merge this.

PR #2499, opened 2026-01-12
Fix #2498 by dropping down hyphen to underscore in autocomplete bash script

### LINKED ISSUE #2498: AutoComplete.bash removes hyphen in aliases, causing duplicate function names
**How to reproduce**
Run `AutoComplete.bash(...)` on a command that has an alias or name using a hyphen (-). The function name for the auto-complete bash script function then removes the hyphen.

**More Context**
This can seemingly cause issues if the alias is only differentiated with a hyphen, i.e. when having a command with `aliases = {"foobar", "foo-bar"}`. The generated bash script then has two functions with the same name. Changing this behavior so that it simply "drops down" the hyphen to and underscore should prevent this, although removing the duplicate function might be the cleaner solution.

btw04:
As discussed in #2498, this should probably fix the duplicate-name issue. I still think that the alias function is a duplicate which could be completely removed, but this should be an important patch nonetheless.

Please let me know if this patch no longer conforms with the intention of `bashify`.

### DIFF
--- a/src/main/java/picocli/AutoComplete.java
+++ b/src/main/java/picocli/AutoComplete.java
@@ -26,6 +26,7 @@ import java.net.InetAddress;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


btw04:
A note on this: I tested it for the `aliases = {"foobar", "foo-bar"}` case where it worked as intended.

remkop:
@btw04 Thank you for the PR!
It covers the specific error case, but if an option has an alias with underscores, we will have the same problem again. 
It's probably better to eliminate duplicate function names after bashify is called. 


(Away from pc now, maybe we can store the names in a set, that may be one way to implement this...)

btw04:
You are right, I did not quite consider the bigger picture here. Maybe I can also think up a cleaner solution in the next couple of days.

btw04:
I think this should be a fairly local patch for the issue. There possibly are other issues related to this, if alias duplication is not checked in another place (i.e. if two different commands have the same subcommand), but I would argue such cases to be a different issue needing a different approach. 

Specifically, I think that the initial problem where two aliases are only differentiated by a hyphen might be the most common. This and similar cases should be fully resolved now.

remkop:
Thank you for the PR!

I noticed that your proposal uses fairly recent API and language features, and I want to point out that picocli is still compiled on Java 5 (https://github.com/remkop/picocli?tab=contributing-ov-file). 
It will be slightly more verbose and less elegant with the old API and syntax but can I ask you to take a look? 
Many thanks! 🙏

btw04:
Thank you for your time. I must have missed that part of the contributing information, sorry! The new commit refactors it for Java 5.

remkop (approved):

PR #2503, opened 2026-02-06
Support jline.jansi alongside fusesource.jansi

### LINKED ISSUE #2437: Support JLine Jansi.
The picocli code currently has some code to see if [Jansi](https://github.com/fusesource/jansi) is installed, e.g.

```java
// loading this class will load the native library org.fusesource.jansi.internal.CLibrary
Class<?> ansiConsole = Class.forName("org.fusesource.jansi.AnsiConsole");
Field out = ansiConsole.getField("out");
return out.get(null) == System.out;
```

In the [notes to the recent Jansi 2.4.2 release](https://github.com/fusesource/jansi/releases/tag/jansi-2.4.2), I noticed they say:

> ⚠️ WARNING
> This library is no longer maintained. It has been merged into `org.jline:jansi`.
> Please update your dependencies to use `org.jline:jansi` instead.

There wasn't much other information, but I suppose they are referring to [JLine](https://jline.github.io/). In its [GitHub repository](https://github.com/jline/jline3/),I see that there are indeed similar classes, for example [`org.jline. jansi.AnsiConsole`](https://github.com/jline/jline3/blob/master/jansi-core/src/main/java/org/jline/jansi/AnsiConsole.java). I'm guessing that's what the Jansi release notes were referring to.

Admittedly there is little information about which we should be using, but it's not unreasonable to interpret this as saying we should switch to [`org.jline. jansi.AnsiConsole`](https://github.com/jline/jline3/blob/master/jansi-core/src/main/java/org/jline/jansi/AnsiConsole.java) going forward. I haven't made that switch in my application yet, but it looks like when I do, it would only work if picocli adds support for it.

Oondanomala:
Jansi has been merged into JLine, which changes the package name (and the system property name).
This pr updates the code to use the updated package name.
Support for the old name is kept to avoid breaking changes. (though I can remove it if it's not wanted)

I chose to prefer the older one if both versions are available to avoid potential breakages, but I can change that if choosing a more up to date version is preferred.

~~Note: the `out` field is `private` in the new versions, ~~so I switched to the method with the same name (which has always been there so there is no compatibility concern)~~ so I switched to calling `isInstalled` instead (which should be more correct anyway). This means Jansi versions older than 2.1.0 will not work. Hopefully that's not a problem.~~
To support both old and new versions of Jansi, it will first try to call `isInstalled` (which should be more correct anyway) and only fall back to the previous method if `isInstalled` is not present.

Fixes #2437

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -17988,8 +17988,7 @@ public class CommandLine {
                     } catch (ClassNotFoundException e) {
                         ansiConsole = Class.forName("org.jline.jansi.AnsiConsole");
                     }
-                    Object out = ansiConsole.getDeclaredMethod("out").invoke(null);
-                    return out == System.out;
+                    return (Boolean) ansiConsole.getDeclaredMethod("isInstalled").invoke(null);
                 } catch (Exception reflectionFailed) {
                     return false;
                 }
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -17988,7 +17988,13 @@ public class CommandLine {
                     } catch (ClassNotFoundException e) {
                         ansiConsole = Class.forName("org.jline.jansi.AnsiConsole");
                     }
-                    return (Boolean) ansiConsole.getDeclaredMethod("isInstalled").invoke(null);
+                    try {
+                        return (Boolean) ansiConsole.getDeclaredMethod("isInstalled").invoke(null);
+                    } catch (ReflectiveOperationException e) {
+                        // isInstalled was "only" added in 2.1.0, try to support older jansi
+                        Field out = ansiConsole.getField("out");
+                        return out.get(null) == System.out;
+                    }
                 } catch (Exception reflectionFailed) {
                     return false;
                 }
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -17954,29 +17954,42 @@ public class CommandLine {
                 if (jansiInstalled == null) { jansiInstalled = calcIsJansiConsoleInstalled(); }
                 return jansiInstalled;
             }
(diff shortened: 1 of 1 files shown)


remkop:
Thank you!
Let me take a look later, hopefully this weekend.

Oondanomala:
Turns out calling the `out` method initializes the output stream, so I've replaced it with a call to `isInstalled`, which should be more correct anyway.

`isInstalled` was "only" added in Jansi 2.1.0 (released 6 years ago), hopefully it's fine to drop support for anything older.

garretwilson:
Good news; JLine has now been modularized, so any native allow overrides needed to be passed to the JVM can now be more granular. But it looks like one may not need the native portions of JLine even.

Please read this [excellent summary](https://github.com/jline/jline3/issues/1646#issuecomment-3951949914) covering how JLine can be used and whether native is even required (it probably isn't). @Oondanomala , you might want to check that summary to make sure the solution you're putting in is compatible with its recommendations.

Oondanomala:
picocli only uses (reflectively, so there is no dependency) Jansi to check whether it has been installed so it can enable ansi on Windows, so it specifically "needs" (it will work fine without it, as it does not attempt to install it itself) the native module and _only_ that one. It has its own code for making ansi codes.
So as far as I can tell that should not affect anything :D

And while modern Windows terminals support ansi without needing to install Jansi, considering picocli still supports Java 5 I would say it's best to keep supporting old terminals :)
There should probably be a check to enable ansi codes on modern Windows terminals that support ansi without needing Jansi, but that is for another PR.

That said, I noticed the test suite actually depends on an even older version of Jansi before `2.1.0` for Java 5 compatibility,
so I added back support for reading the `out` field directly if `isInstalled` isn't available.
This means the PR now does not drop support for any Jansi version!

remkop (approved):

jdimeo:
Very excited for this - thank you. Recently upgraded to JLine 4 to get richer CLI features and all the ANSI detection/code broke.

Issue #2491, opened 2025-11-21
Spring Boot 4 support


ydolzhenko:
Spring Boot 4 is out and it would be great if there is a picocli starter based on it.

remkop:
Thank you for the suggestion!
Will you be able to provide a pull request for this?

ydolzhenko:
Yep, I will take a look

davinkevin:
Highly looking for this one after I discovered `picocli` wasn't compatible 😅. 
It's the last step of this migration 😇… can't wait for it!

cafec0c0:
I'm also running into this issue with a project I'm working on. It looks to me as though it is sufficient to just add the Spring Boot 4 dependencies as I have been able to successfully use the library without any code changes. 

As the 3.x and 4.x versions are incompatible I would like to suggest creating a new `picocli-spring-boot-4-starter` artifact to differentiate.  

Unfortunately, another issue is that the spring boot 4.x bytecode is compiled against Java 17 which would mean that building the artifact would require Java 17 to be used in the picocli pipelines.

I see that there hasn't been any update on this issue in the last few months so I'm happy to provide an initial PR that we can iterate on 🙂

remkop:
> I'm also running into this issue with a project I'm working on. It looks to me as though it is sufficient to just add the Spring Boot 4 dependencies as I have been able to successfully use the library without any code changes.
> 
> As the 3.x and 4.x versions are incompatible I would like to suggest creating a new `picocli-spring-boot-4-starter` artifact to differentiate.
> 
> Unfortunately, another issue is that the spring boot 4.x bytecode is compiled against Java 17 which would mean that building the artifact would require Java 17 to be used in the picocli pipelines.
> 
> I see that there hasn't been any update on this issue in the last few months so I'm happy to provide an initial PR that we can iterate on 🙂

Yes, a new artifact is a good idea. 
A PR would be very welcome, I have not been able to find time or energy to work on picocli, any help is appreciated!

Albertoimpl:
We are very interested on this support. We have not been able to upgrade to boot 4 our project. I see the PR has been there for a while now. Is there anything any outsider can do to help?

Albertoimpl:
Hey @remkop sorry to ping again but it's almost one year since Spring Boot 4 was released. 
There has been this community PR but never made it through.
Are you planning on officially discontinuing the project? 
Is there anything anyone can do to help?
Thanks!

remkop:
[...]

PR #2507, opened 2026-03-03
[#2491] Add spring-boot-4-starter module

cafec0c0:
This PR addresses the issue raised in #2491 which requests support for Spring Boot 4.

To accommodate this, a new `spring-boot-4-starter` module has been added.

The new module reuses the existing spring boot code, as it is still code-compatible with the code for spring boot 2 and 3, but does require using newer spring boot dependencies (4.x.x).

This unfortunately requires a Java 17+ compiler, and Java 9+ drops support for targeting Java 1.5.

I've thought of two solutions to this, but any others are welcomed:
1) 2 Compiler passes: the first with a Java 8 compiler and the second with a Java 17 compiler. I think this would need some CI configuration to only use Java 17 for modules that require it (i.e. running `./gradlew :picocli-spring-boot-4-starter:build`), so we don't inadvertently overwrite the modules compiled previously.
2) Use Java 17+ for all modules: this would simplify CI, but would break Java 1.5 and 1.6 compatibility for modules.


martamedio:
@remkop Is there any chance this could be included in future releases? 🙏🏼

paulcwarren:
+1 to add support for boot 4 please

PR #2509, opened 2026-04-03
Allow @Spec on method parameters for method-based subcommands

### LINKED ISSUE #1670: allow a @Spec CommandSpec parameter to methods that define a subcommand
### Discussed in https://github.com/remkop/picocli/discussions/1668

<div type='discussions-op-text'>

<sup>Originally posted by **bartweber** April 28, 2022</sup>
A subcommand defined by a method needs to throw a `ParameterException`. The `ParameterException` expects `CommandLine` as the first argument. This object `CommandLine` can be obtained from `CommandSpec`. `CommandSpec` can be injected by PicoCLI by annotating a field for this with `@Spec`

```
@Spec
private CommandSpec spec;
```

Using this injected `CommandSpec` to obtain the `CommandLine` will not be the `CommandLine` of the subcommand but of the parent.

**How to obtain the `CommandLine` of the subcommand defined by a method?**

One solution is like:
```
spec.subcommands().get("my-subcommand")
```
But this could cause runtime errors if the command name was changed by a developer which forgot to update the key as well for looking up the `CommandLine` object for the subcommand as is done in this code block.

(...)

A better solution would be to allow '@CommandLine.Spec' as a parameter of methods defining a subcommand.
</div>

EmperorSARATH:
This PR addresses issue #1670 by allowing `@Spec` to be used on method parameters in method-based subcommands.

### Changes:
- Added `ElementType.PARAMETER` to `@Spec` to enable usage on method parameters.
- Inject the correct `CommandSpec` via `commandMethodParamValues()` for method subcommands.
- Prevent `@Spec` parameters from being treated as CLI arguments.
- Scoped the change to method parameters to avoid affecting existing behavior.
- Verified that all existing tests pass.

### Motivation:
Previously, method-based subcommands could not directly receive a `CommandSpec` via a parameter, forcing developers to use fragile workarounds like looking up the subcommand from the parent spec. This PR provides a clean and safe way to inject the subcommand's `CommandSpec`.

### Example usage:

```java
import picocli.CommandLine;
import picocli.CommandLine.Command;
import picocli.CommandLine.Spec;
import picocli.CommandLine.Model.CommandSpec;

@Command(name = "app", subcommands = {TestApp.class})
public class TestApp {

    public static void main(String[] args) {
        CommandLine cmd = new CommandLine(new TestApp());
        cmd.execute(args);
    }

    @Command(name = "sub")
    void sub(@Spec CommandSpec spec) {
        System.out.println("Subcommand name: " + spec.name());
    }
}


```

Example run :

```
$ java -cp .:build/libs/picocli-4.7.8-SNAPSHOT.jar TestApp sub
Subcommand name: sub
```

Fixes: #1670

### DIFF
--- a/src/main/java/picocli/CommandLine.java
+++ b/src/main/java/picocli/CommandLine.java
@@ -46,6 +46,7 @@ import picocli.CommandLine.Help.Ansi.Style;
 import picocli.CommandLine.Help.Ansi.Text;
 import picocli.CommandLine.Model.*;
 import picocli.CommandLine.ParseResult.GroupMatchContainer;
+import picocli.CommandLine.Spec;
 
 import static java.util.Locale.ENGLISH;
 import static picocli.CommandLine.Help.Column.Overflow.SPAN;
@@ -4486,7 +4487,7 @@ public class CommandLine {
      * @since 3.2
      */
     @Retention(RetentionPolicy.RUNTIME)
-    @Target({ElementType.FIELD, ElementType.METHOD})
+    @Target({ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER})
     public @interface Spec {
         /** Identifies what kind of {@code CommandSpec} should be injected.
          * @since 4.3.0 */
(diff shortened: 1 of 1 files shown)


remkop:
This looks good, thank you for the contribution!
I plan to include this in the 4.8 release.

